PR #11, opened 2023-02-02
build: Add make recipes for deploying local builds

jimmidyson:
Running `make dev.run-on-kind` will start a KinD cluster, use `clusterctl` to deploy
relevant providers, build Docker images, copy to the KinD cluster, and deploy
the extensions via Helm.


dkoshkin (approved):
README and new makefile targets are working great on my Mac.

Tested by correctly creating a cluster with `spec.topology`

PR #16, opened 2023-02-04
fix: Calico deployment to work with CAPD template

dkoshkin:
Disable [Pod Security Standards](https://kubernetes.io/docs/concepts/security/pod-security-standards/) and will PR upstream to allow configuring the exemptions namespaces.

I considered committing the template yaml file with a modified spec, but this CAPD dev cluster is just for testing.


jimmidyson (approved):

jimmidyson:
I updated this to actually use a modified template to get used to working with pod security standards, hope you don't mind :relaxed:

dkoshkin:
Oh nice that's better, thanks.

PR #27, opened 2023-02-14
feat: Enable controller manager

jimmidyson:
Currently run in same process as runtime hooks server.


dkoshkin (approved):
Changes look good, tested locally too.

PR #29, opened 2023-02-14
feat: delete Services type LoadBalancer on BeforeClusterDelete

dkoshkin:
Fixes https://github.com/d2iq-labs/capi-runtime-extensions/issues/4

Automatically delete workload cluster Services with type `LoadBalancers` before deleting the cluster.

It checks for certain conditions on the Cluster object if Services should be deleted.
It's also possible to disable this behavior by adding an annotation `capi-runtime-extensions.d2iq-labs.com/loadbalancer-gc` to the Cluster object.

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -4,5 +4,5 @@
 package constants
 
 const (
-	LoadBalancerGCAnnotation = "capi-runtime-extensions.d2iq-labs.com/loadbalancer-gc"
+	LoadBalancerGCAnnotation = "labs.d2iq.io/loadbalancer-gc"
 )
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -1,0 +1,8 @@
+// Copyright 2023 D2iQ, Inc. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package constants
+
+const (
+	LoadBalancerGCAnnotation = "capi-runtime-extensions.d2iq-labs.com/loadbalancer-gc"
+)
--- a/pkg/k8s/annotations/annotations.go
+++ b/pkg/k8s/annotations/annotations.go
@@ -6,13 +6,8 @@ package annotations
 import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 // Get will get the value of the supplied annotation.
-func Get(obj metav1.Object, name string) (value string, found bool) {
+func Get(obj metav1.Object, name string) (string, bool) {
 	annotations := obj.GetAnnotations()
-	if len(annotations) == 0 {
-		return "", false
-	}
-
-	value, found = annotations[name]
-
-	return
+	val, found := annotations[name]
+	return val, found
 }
--- a/pkg/k8s/annotations/annotations.go
+++ b/pkg/k8s/annotations/annotations.go
@@ -1,0 +1,18 @@
+// Copyright 2023 D2iQ, Inc. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package annotations
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+// Get will get the value of the supplied annotation.
+func Get(obj metav1.Object, name string) (value string, found bool) {
+	annotations := obj.GetAnnotations()
+	if len(annotations) == 0 {
+		return "", false
+	}
+
+	value, found = annotations[name]
+
+	return
+}
--- a/pkg/k8s/deleter/deleter.go
+++ b/pkg/k8s/deleter/deleter.go
@@ -5,13 +5,15 @@ package deleter
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"sigs.k8s.io/cluster-api/api/v1beta1"
@@ -22,6 +24,8 @@ import (
 	"github.com/d2iq-labs/capi-runtime-extensions/pkg/k8s/annotations"
 )
 
+var ErrFailedToDeleteService = errors.New("kubernetes Services deletion failed")
+
 type Deleter struct {
 	cluster *v1beta1.Cluster
 	client  client.Client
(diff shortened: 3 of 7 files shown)


pkg/constants/constants.go
```diff
@@ -0,0 +1,8 @@
+// Copyright 2023 D2iQ, Inc. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package constants
+
+const (
+	LoadBalancerGCAnnotation = "capi-runtime-extensions.d2iq-labs.com/loadbalancer-gc"
```
jimmidyson, line 7: Let's use same domain as in API (https://github.com/d2iq-labs/capi-runtime-extensions/blob/main/api/v1alpha1/groupversion_info.go#L16):

```suggestion
	LoadBalancerGCAnnotation = "labs.d2iq.io/loadbalancer-gc"
```

pkg/k8s/annotations/annotations.go
```diff
@@ -0,0 +1,18 @@
...
+func Get(obj metav1.Object, name string) (value string, found bool) {
+	annotations := obj.GetAnnotations()
+	if len(annotations) == 0 {
+		return "", false
+	}
+
+	value, found = annotations[name]
+
+	return
+}
```
jimmidyson, line 18: nit: only use named returns if really necessary, and prefer explicit return values

Also don't need length check, even if annotations is nil.

```suggestion
func Get(obj metav1.Object, name string) (string, bool) {
	annotations := obj.GetAnnotations()
	val, found := annotations[name]
	return val, found
}
```

pkg/k8s/deleter/deleter.go
```diff
@@ -0,0 +1,188 @@
...
+func (ol objectMetaList) asCommaSeparatedString() string {
+	out := ""
+	separator := ""
+	for n := range ol {
+		object := ol[n]
+		out += fmt.Sprintf("%s%s/%s", separator, object.Namespace, object.Name)
+		separator = ", "
+	}
+	return out
+}
```
jimmidyson, line 42: nit: simplify with `strings.Join`

```suggestion
func (ol objectMetaList) asCommaSeparatedString() string {
	names := make([]string, 0, len(ol))
	for _, obj := range ol {
		names = append(names, fmt.Sprintf("%s/%s", obj.Namespace, obj.Name))
	}
	return strings.Join(names, ",")
}
```

pkg/k8s/deleter/deleter.go
```diff
@@ -0,0 +1,188 @@
...
+		log:     log,
+	}
+}
+
+func (d *Deleter) DeleteServicesWithLoadBalancer(ctx context.Context) error {
+	err := deleteServicesWithLoadBalancer(ctx, d.client, d.log)
+	if err != nil {
+		return err
+	}
+	return nil
```
jimmidyson, line 57:
```suggestion
	return deleteServicesWithLoadBalancer(ctx, d.client, d.log)
```

pkg/k8s/deleter/deleter.go
```diff
@@ -0,0 +1,188 @@
...
+	log logr.Logger,
+) error {
+	log.Info("Listing Services with type LoadBalancer")
+	services := &corev1.ServiceList{}
+	err := c.List(ctx, services)
+	if err != nil {
+		return fmt.Errorf("error listing Services: %w", err)
+	}
+
+	svcsFailedToBeDeleted := make(objectMetaList, 0)
```
jimmidyson, line 72: nit: Better to just nil assign this rather than create an empty list to avoid potentially unnecessary allocation.
```suggestion
	var svcsFailedToBeDeleted objectMetaList
```

pkg/k8s/deleter/deleter.go
```diff
@@ -0,0 +1,188 @@
...
+	for i := range services.Items {
+		svc := &services.Items[i]
+		if needsDelete(svc) {
+			log.Info(fmt.Sprintf("Deleting Service %s/%s", svc.Namespace, svc.Name))
+			if err = c.Delete(ctx, svc); err != nil {
```
jimmidyson, line 77: Should we handle ignore found error here?

pkg/k8s/deleter/deleter.go
```diff
@@ -0,0 +1,188 @@
...
+		if err != nil {
+			if errors.IsNotFound(err) {
+				err = nil
+				return true, nil
+			}
+		}
+		return false, nil
+	})
+
+	return
```
jimmidyson, line 146: Be simple: remove named return (prefer explicit return), capture error as normal, returning nil from func if `errors.IsNotFound` is true.

```suggestion
func waitForServiceDeletion(
	ctx context.Context,
	c client.Client,
	service *corev1.Service,
) (error) {
	backoff := wait.Backoff{
		Duration: 1 * time.Second,
		Factor:   1.5,
		Jitter:   0,
		Steps:    13,
	}
	return wait.ExponentialBackoff(backoff, func() (bool, error) {
		key := client.ObjectKey{
			Namespace: service.Namespace,
			Name:      service.Name,
		}
		err := c.Get(ctx, key, service)
		if err != nil {
			if errors.IsNotFound(err) {
				return true, nil
			}
			return false, err
		}
		return false, nil
	})
```

pkg/k8s/deleter/deleter.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+func failedToDeleteServicesError(svcsFailedToBeDeleted objectMetaList) error {
+	//nolint:goerr113 // This error is specific to this function
+	return fmt.Errorf("the following Services could not be deleted "+
+		"and must cleaned up manually before deleting the cluster: %s", svcsFailedToBeDeleted.asCommaSeparatedString())
+}
```
jimmidyson, line 153: Can still use a sentinel error, makes error checking easier in tests:

var ErrFailedToDeleteService = errors.New("service deletion failed")

```suggestion
func failedToDeleteServicesError(svcsFailedToBeDeleted objectMetaList) error {
	return fmt.Errorf("%w: the following Services could not be deleted "+
		"and must cleaned up manually before deleting the cluster: %s", ErrFailedToDeleteService, svcsFailedToBeDeleted.asCommaSeparatedString())
}
```

pkg/k8s/deleter/deleter.go
```diff
@@ -0,0 +1,188 @@
...
+	if err != nil {
+		return false, fmt.Errorf(
+			"converting value %s of annotation %s to bool: %w",
+			val,
+			constants.LoadBalancerGCAnnotation,
+			err,
+		)
+	}
+
+	// use the Cluster phase to determine if its safe to skip deleting
```
jimmidyson, line 171: nit
```suggestion
	// use the Cluster phase to determine if it's safe to skip deleting
```

pkg/k8s/deleter/deleter.go
```diff
@@ -0,0 +1,188 @@
...
+	//
+	// when ClusterPhasePending or ClusterPhaseProvisioning Kubernetes API has not been created
+	// and the user would not have been able to create any Kubernetes resources that would prevent cleanup
+	//nolint:lll // long URL cannot be split up
+	// https://github.com/kubernetes-sigs/cluster-api/blob/7f879be68d15737e335b6cb39d380d1d163e06e6/controllers/cluster_controller_phases.go#L44-L50
+	//
+	// when ClusterPhaseDeleting its too late to try to cleanup
```
jimmidyson, line 178: nit
```suggestion
	// when ClusterPhaseDeleting it's too late to try to cleanup
```

jimmidyson (approved):
Loving this! Most of my comments are stylistic or nits rather than functional. Thank you for adding this.

pkg/k8s/deleter/deleter.go
dkoshkin: 😅

jimmidyson: Punctuation is important :joy: i feel like a bit of a **** doing this now :facepalm:

dkoshkin: Haha I'm glad you did! Now its out there though so I'll never get into that writing program at Harvard 😆

pkg/k8s/deleter/deleter.go
dkoshkin: Good idea! There is a convenient `client.IgnoreNotFound(err)` function.

PR #98, opened 2023-08-07
feat: Use interface to register handlers

jimmidyson:
This also implements and registers a handler based on some old code from this repo for deleting load-balancer services before cluster delete. The code has been refactored to not block return but instead block by sending retry-after response to CAPI.


mhrabovcin (approved):

internal/runtimehooks/webhooks/server.go
```diff
@@ -78,68 +78,87 @@ func (s *Server) Start(ctx context.Context) error {
...
-		return err
-	}
-	if err := webhookServer.AddExtensionHandler(server.ExtensionHandler{
-		Hook:        runtimehooksv1.ValidateTopology,
-		Name:        "validate-topology",
-		HandlerFunc: topologyMutationHandlers.DoValidateTopology,
-	}); err != nil {
-		setupLog.Error(err, "error adding handler")
-		return err
+	allHandlers := []handlers.NamedHandler{servicelbgc.New(client)}
```
dkoshkin, line 81: A future improvement would be for the Handler to register itself, similar to the SchemeBuilder pattern [and now this](https://github.com/kubernetes-sigs/cluster-api/blob/2061cabe56774afa2627cb15f155ed12cf1fc9ea/api/v1beta1/machine_types.go#L309-L311)

pkg/handlers/servicelbgc/handler.go
```diff
@@ -0,0 +1,90 @@
...
+
+	err = deleteServicesWithLoadBalancer(ctx, remoteClient, log)
+	switch {
+	case errors.Is(err, ErrFailedToDeleteService):
+		resp.Status = runtimehooksv1.ResponseStatusFailure
+		resp.Message = err.Error()
+	case errors.Is(err, ErrServicesStillExist):
+		resp.Status = runtimehooksv1.ResponseStatusSuccess
+		resp.Message = err.Error()
+		resp.RetryAfterSeconds = 5
```
dkoshkin, line 86: Nice! We need to be careful to treat this like an actual controller and be cautious that it can restart at any point and it gracefully handle retrying.

dkoshkin (approved):

internal/runtimehooks/webhooks/server.go
jimmidyson: Great idea, and one I was going to implement but wasn't sure where the handlers would actually be registered.

pkg/handlers/servicelbgc/handler.go
jimmidyson: Yep, idempotence ftw. I think this handles that for this handler, but obviously need to add tests.

PR #107, opened 2023-08-17
feat: Deploy calico CNI via CRS

jimmidyson:


pkg/handlers/cni/calico/manifests/docker/installation.yaml
```diff
@@ -0,0 +1,23 @@
...
+  name: default
+spec:
+  cni:
+    type: Calico
+  # Configures Calico networking.
+  calicoNetwork:
+    # Note: The ipPools section cannot be modified post-install.
+    ipPools:
+    - blockSize: 26
+      cidr: ${POD_SUBNET}
```
dkoshkin, line 18: Should `cidr` be set to an empty value or a default value instead? If I understand this not a used as a template right?

pkg/handlers/cni/calico/manifests/docker/installation.yaml
```diff
@@ -0,0 +1,23 @@
+# Copyright 2023 D2iQ, Inc. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# This section includes base Calico installation configuration.
+# For more information, see: https://docs.projectcalico.org/reference/installation/api
+apiVersion: operator.tigera.io/v1
+kind: Installation
```
dkoshkin, line 7: Having this in the codebase is a great start, but I do think we'll need add support for user overrides here sooner than later. Its a good strategy though to only allow overrides to a limited config, and always provide defaults: Calico `Installation`, CSI `StorageClass`, CPI `ConfigMap/Secret` that may be it really.

jimmidyson: I think I'd like to move this from being embedded to reference default configmap from the API, and then copying it that way to where it's needed. A default would be created when installing the helm chart, but could be disabled via some value. This would allow users to create their own default configmaps/secrets.

Then allowing configuration per cluster via annotations to use other specific configmaps.

pkg/handlers/cni/calico/manifests/docker/installation.yaml
jimmidyson: Yes, this should just contain the default... I got caught up in trying to use unstructured helpers to set a field in an element in a nested slice 😅. I'll add the default directly in here and then figure out how we can allow overrides. I'm starting to think about discover variables mutation hook to be able to specify it rather than annotations, but that will be more work so will start with annotations.

pkg/handlers/cni/calico/handler.go
```diff
@@ -0,0 +1,310 @@
...
+		log.Error(err, "failed to apply Tigera ClusterResourceSet")
+		resp.SetStatus(runtimehooksv1.ResponseStatusFailure)
+		resp.SetMessage(fmt.Sprintf("failed to apply Tigera ClusterResourceSet: %v", err))
+	}
+
+	if err := applyProviderCNICRS(ctx, manifestsFS, s.client, &req.Cluster, log); err != nil {
+		log.Error(err, "failed to apply CNI installation ClusterResourceSet")
+		resp.SetStatus(runtimehooksv1.ResponseStatusFailure)
+		resp.SetMessage(fmt.Sprintf("failed to apply CNI installation ClusterResourceSet: %v", err))
+	}
```
mhrabovcin, line 112: Should this hook block until the installation is complete? Could we verify that CNI is working correctly?

jimmidyson: This is one of the problems of using CRS, there is nothing to wait on really other than whether the manifests are applied in remote cluster. I don't think there's a need to block until cni is running, rest of cluster deployment can continue while the CNI pods are started imo. We should be very careful about when we block.

pkg/handlers/cni/calico/manifests/docker/installation.yaml
jimmidyson: I've added the ability to override pod subnet via annotation in latest commit to show how that is possible.

pkg/handlers/cni/calico/handler.go
```diff
@@ -0,0 +1,310 @@
...
+		"cluster",
+		clusterKey,
+	)
+
+	// Safe to set this to success response before we actually do the apply as the error handling
+	// will update for failure response properly.
+	resp.SetStatus(runtimehooksv1.ResponseStatusSuccess)
+
+	if v, ok := req.Cluster.GetLabels()[cni.CNILabelKey]; !ok || v != CNILabelValue {
+		log.V(4).Info("Skipping Calico CNI handler, CNI provider is not calico")
```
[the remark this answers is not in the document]
jimmidyson: Done, added more explicit reason.

pkg/handlers/cni/calico/manifests/docker/installation.yaml
jimmidyson: Added default into file, and allowed override via annotation in latest commit.

pkg/handlers/cni/calico/handler.go
dkoshkin: At most we could wait for the `ClusterResourceSetBinding`, but I think that would only to surface errors and set Conditions on the Cluster object.

pkg/handlers/cni/calico/manifests/docker/installation.yaml
dkoshkin: >I've added the ability to override pod subnet via annotation in latest commit to show how that is possible.

I think this value should come from the Cluster object directly, idk the side-effects of changing it only 1 place but it would be confusing if those values were different.
https://github.com/kubernetes-sigs/cluster-api/blob/82eff49867008365d3b26f82b55ff29a67880aa7/api/v1beta1/cluster_types.go#L362C15-L362C15
WDYT?

jimmidyson: Ooo yes I'd missed that. But what would multiple cidrs specified here mean?

pkg/handlers/cni/calico/handler.go
jimmidyson: I think a condition to say CNI is deployed and working is a good idea, but what should we check for? I feel the same with CSI and CPI btw.

pkg/handlers/cni/calico/manifests/docker/installation.yaml
jimmidyson: We could use the equivalent of `calico node status` against a cluster using the Calico API via a controller on a regular poll perhaps.

dkoshkin (approved):
Thanks!

PR #115, opened 2023-08-24
feat: add httpproxy external patch

mhrabovcin:


docs/content/service-loadbalancer-gc.md
```diff
@@ -16,4 +16,4 @@ This hook is enabled by default, and can be explicitly disabled by omitting the
 If deploying via Helm, then this can be disabled by setting `handlers.ServiceLoadBalancerGC.enabled=false`.
 
 By default, all clusters will be cleaned up when deleting, but this can be opted out from by setting the annotation
-`capiext.labs.d2iq.io//loadbalancer-gc=false`.
+`capiext.labs.d2iq.io/loadbalancer-gc=false`.
```
jimmidyson, line 19: Thank you!

internal/runtimehooks/webhooks/server.go
```diff
@@ -105,9 +107,12 @@ func (s *Server) Start(ctx context.Context) error {
 		return err
 	}
 
+	httProxyConfig := httpproxyconfig.New()
 	allHandlers := []handlers.NamedHandler{
 		servicelbgc.New(client),
 		calico.New(client, *s.calicoCNIConfig),
+		name.NewDiscoveryVariables(name.Suffix("vars"), httProxyConfig),
+		name.NewGeneratePatches(name.Suffix("patch"), httProxyConfig),
```
jimmidyson, line 115: Do you really dislike the idea of doing this in the handler registration blocks rather than here?

pkg/handlers/httpproxyconfig/variables.go
```diff
@@ -0,0 +1,62 @@
...
+func GetVariable[T any](variables map[string]apiextensionsv1.JSON, name string) (T, bool, error) {
+	var v T
+	variable, found, err := topologymutation.GetVariable(variables, name)
+	if err != nil || !found {
+		return v, found, err
+	}
+
+	err = json.Unmarshal(variable.Raw, &v)
+	return v, err == nil, err
+}
```
jimmidyson, line 24: This is generic enough to move out to it's own package, perhaps `pkg/capi/variables.go` or something?

pkg/handlers/httpproxyconfig/variables.go
```diff
@@ -0,0 +1,62 @@
...
+
+// VariableSchema provides Cluster Class variable schema definition.
+func (HTTPProxyVariables) VariableSchema() clusterv1.VariableSchema {
+	return clusterv1.VariableSchema{
+		OpenAPIV3Schema: clusterv1.JSONSchemaProps{
+			Type: "object",
+			Properties: map[string]clusterv1.JSONSchemaProps{
+				"http": {
+					Description: "HTTP proxy for CAPI controllers.",
```
jimmidyson, line 45: Are these variables really just for CAPI controllers?

internal/runtimehooks/webhooks/server.go
mhrabovcin: I've removed that part and split the handler into 2 separate handler:

```
http-proxy-vars
http-proxy-inject
```

pkg/handlers/httpproxyconfig/variables.go
mhrabovcin: I've changed name to generic.

internal/runtimehooks/webhooks/server.go
jimmidyson: If we want to enforce a separate handler per hook type, we could set the function names in the interfaces to be the same but with different parameter types.

mhrabovcin: I agree, but I think in this case it makes sense to split this handler into 2 separate handlers, as we could have multiple handlers that would somehow act on http proxy variable.

charts/capi-runtime-extensions/README.md
```diff
@@ -38,6 +38,8 @@ A Helm chart for capi-runtime-extensions
 | handlers.CalicoCNI.defaultTigeraOperatorConfigMap.name | string | `"tigera-operator"` |  |
 | handlers.CalicoCNI.enabled | bool | `true` |  |
 | handlers.ServiceLoadBalancerGC.enabled | bool | `true` |  |
+| handlers.http-proxy-inject.enabled | bool | `true` |  |
+| handlers.http-proxy-vars.enabled | bool | `true` |  |
```
dkoshkin, line 42: Maybe something like this for a consistent case
```suggestion
| handlers.HTTPProxyPatch.enabled | bool | `true` |  |
| handlers.HTTPProxyVars.enabled | bool | `true` |  |
```

mhrabovcin: I've changed variable names as suggested.

jimmidyson (approved):
Great work!

PR #126, opened 2023-08-30
feat: Add audit policy patch

jimmidyson:


pkg/handlers/httpproxy/inject.go
```diff
@@ -111,19 +102,8 @@ func (h *httpProxyPatchHandler) GeneratePatches(
...
-						Names: []string{
-							"default-worker",
-						},
-					},
-				},
-			}
-			if err := generatePatch(
-				obj, vars, &holderRef, defaultWorkerSelector, log,
+			if err := patches.Generate(
+				obj, vars, &holderRef, selectors.AllWorkersSelector(), log,
```
jimmidyson, line 106: @mhrabovcin @dkoshkin I changed the http proxy patch selector to target all workers rather than just `default-worker` - do you agree this is correct?

pkg/handlers/auditpolicy/inject_test.go
```diff
@@ -0,0 +1,136 @@
...
+				}
+				return operations, nil
+			},
+			ConsistOf(
+				MatchAllFields(Fields{
+					"Operation": Equal("add"),
+					"Path":      Equal("/spec/template/spec/kubeadmConfigSpec/files"),
+					"Value":     HaveLen(1),
+				}),
+				jsonpatch.NewOperation(
```
mhrabovcin, line 73: 👍

mhrabovcin (approved):

PR #128, opened 2023-08-31
feat: calculate default no_proxy values based on cluster

mhrabovcin:
Implements https://github.com/d2iq-labs/capi-runtime-extensions/issues/30

Example of generated values:

```
          [Service]
          Environment="HTTP_PROXY=http://example.com"
          Environment="http_proxy=http://example.com"
          Environment="NO_PROXY=localhost,127.0.0.1,192.168.0.0/16,10.128.0.0/12,kubernetes,kubernetes.default,.svc,.svc,.svc.cluster,.svc.cluster.
local,no-1.example.com"
          Environment="no_proxy=localhost,127.0.0.1,192.168.0.0/16,10.128.0.0/12,kubernetes,kubernetes.default,.svc,.svc,.svc.cluster,.svc.cluster.
local,no-1.example.com"
```

### DIFF
--- a/pkg/handlers/httpproxy/inject.go
+++ b/pkg/handlers/httpproxy/inject.go
@@ -1,20 +1,27 @@
 // Copyright 2023 D2iQ, Inc. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+// +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=clusters,verbs=watch;list;get
+
 package httpproxy
 
 import (
 	"context"
+	"errors"
+	"fmt"
+	"strings"
 
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/apimachinery/pkg/types"
+	capiv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	bootstrapv1 "sigs.k8s.io/cluster-api/bootstrap/kubeadm/api/v1beta1"
 	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
 	runtimehooksv1 "sigs.k8s.io/cluster-api/exp/runtime/hooks/api/v1alpha1"
 	"sigs.k8s.io/cluster-api/exp/runtime/topologymutation"
 	ctrl "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/client"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/d2iq-labs/capi-runtime-extensions/common/pkg/handlers"
 	"github.com/d2iq-labs/capi-runtime-extensions/common/pkg/handlers/mutation"
@@ -30,6 +37,7 @@ const (
 
 type httpProxyPatchHandler struct {
 	decoder runtime.Decoder
+	client  ctrlclient.Reader
 }
 
 var (
@@ -37,7 +45,7 @@ var (
 	_ mutation.GeneratePatches = &httpProxyPatchHandler{}
 )
 
-func NewPatch() *httpProxyPatchHandler {
+func NewPatch(cl ctrlclient.Reader) *httpProxyPatchHandler {
 	scheme := runtime.NewScheme()
 	_ = bootstrapv1.AddToScheme(scheme)
 	_ = controlplanev1.AddToScheme(scheme)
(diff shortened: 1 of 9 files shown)


jimmidyson:
No proxy shouldn't have values with scheme should it?

mhrabovcin:
> No proxy shouldn't have values with scheme should it?

@jimmidyson yes, we should add validation to `additionalNo` value. The values at the end are coming from the user input config.

dlipovetsky:
Looks good!

Clusters in some infrastructures will need additional values to NO_PROXY.

For example, in AWS, the metadata service IP (`169.254.169.254`), and the load-balancer subdomain (`.elb.amazonaws.com`).

I think we determine the infrastructure by 
(a)  checking the `Kind` of the infrastructure reference in the Cluster spec to determine the infrastructure, OR
(b) reading the `konvoy.d2iq.io/provider` label

Would be a good follow-up. WDYT?

pkg/handlers/httpproxy/inject.go
```diff
@@ -122,3 +137,78 @@ func (h *httpProxyPatchHandler) GeneratePatches(
...
+	return noProxy
+}
+
+// mapServiceDomain generates NO_PROXY values for service domain to allow use
+// of short service name in cluster.
+// Example: my.custom => [.svc, .svc.my, .svc.my.custom].
+func mapServiceDomain(serviceDomain, svcName string) []string {
+	mapped := []string{svcName}
+	progress := svcName
+	for _, part := range strings.Split(serviceDomain, ".") {
```
dkoshkin, line 209: Will this always be in `*.*` format?

jimmidyson: I think it could be an arbitrary number of segments but would need to check. I also think we could probably get away just with adding the whole cluster domain rather than individual parts

jimmidyson: Validation in kubeadm at https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/apis/kubeadm/validation/validation.go#L531 defined in regexp https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation.go#L205

jimmidyson: We should add pattern validation to the open API schema, I need to do the same for cert sans

mhrabovcin: Thanks for confirming this @jimmidyson. The parser implemented here will map any number of parts so I think it works correctly.

pkg/handlers/httpproxy/inject.go
```diff
@@ -1,20 +1,27 @@
...
 	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/apimachinery/pkg/types"
+	capiv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	bootstrapv1 "sigs.k8s.io/cluster-api/bootstrap/kubeadm/api/v1beta1"
 	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
 	runtimehooksv1 "sigs.k8s.io/cluster-api/exp/runtime/hooks/api/v1alpha1"
 	"sigs.k8s.io/cluster-api/exp/runtime/topologymutation"
 	ctrl "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/client"
+	ctrclient "sigs.k8s.io/controller-runtime/pkg/client"
```
jimmidyson, line 24: nit: use `ctrlclient` for consistency with import in other files
```suggestion
	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
```

pkg/handlers/httpproxy/inject.go
jimmidyson: I'm not sure we need to add every part of the cluster domain though. I would expect only to include the full domain, otherwise this could have some  (highly unlikely but possible) invalid behaviour if there is an valid external domain that is a prefix of the cluster domain.

`my.custom => [.svc, svc.my.custom]`

pkg/handlers/httpproxy/variables.go
```diff
@@ -76,9 +77,10 @@ func (HTTPProxyVariables) VariableSchema() clusterv1.VariableSchema {
 					Description: "HTTPS proxy value.",
 					Type:        "string",
 				},
-				"no": {
-					Description: "No Proxy list.",
-					Type:        "array",
+				"additionalNo": {
+					Description: "Additional No Proxy list that will be added to the automatically calculated " +
+						"values that will apply no_proxy configuration for cluster internal network.",
```
jimmidyson, line 82: Could you please add the default no proxy list that this will be appended to?

mhrabovcin: Added.

pkg/handlers/httpproxy/inject.go
mhrabovcin: Changed the behavior to only include values as described.

jimmidyson (approved):

PR #129, opened 2023-08-31
feat: Add API server cert SANs patch

jimmidyson:

PR #130, opened 2023-09-01
feat: Add DNS patch

jimmidyson:
Depends on #129.

### DIFF
--- a/docs/content/dns.md
+++ b/docs/content/dns.md
@@ -1,0 +1,42 @@
+---
+title: "DNS"
+---
+
+To enable configuration of DNS components enable the `dnsvars` and `dnspatch` external patches on `ClusterClass`.
+
+```yaml
+apiVersion: cluster.x-k8s.io/v1beta1
+kind: ClusterClass
+metadata:
+  name: <NAME>
+spec:
+  patches:
+    - name: dns
+      external:
+        generateExtension: "dnspatch.<external-config-name>"
+        discoverVariablesExtension: "dnsvars.<external-config-name>"
+```
+
+On the cluster resource then specify desired DNS image repo:
+
+```yaml
+apiVersion: cluster.x-k8s.io/v1beta1
+kind: Cluster
+metadata:
+  name: <NAME>
+spec:
+  topology:
+    variables:
+      - name: dns
+        value:
+          imageRepository: a.b.c.example.com
+```
+
+Applying this configuration will result in the DNS image repository being correctly set in the
+`KubeadmControlPlaneTemplate`.
+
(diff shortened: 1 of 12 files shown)


docs/content/dns.md
```diff
@@ -0,0 +1,46 @@
+---
+title: "DNS"
+---
+
+If the API server can be accessed by alternative DNS addresses then setting additional SANs on the API server
+certificate is necessary in order for clients to successfully validate the API server certificate.
+
+To enable the API server certificate SANs enable the `dnsvars` and `dnspatch`
+external patches on `ClusterClass`.
```
dkoshkin, line 9: Sorry I'm not understanding the relationship between additional SANs and this DNS image field https://github.com/kubernetes-sigs/cluster-api/blob/c29c5f4d467422784d4f29b704972b4527fa6f94/bootstrap/kubeadm/types/upstreamv1beta3/types.go#L176-L184, can you please clarify it.

jimmidyson: Copy-paste mistake :sweat:

jimmidyson:
@dkoshkin do we really need to be able to override DNS image repo here? Is it not enough to just be able to override the registry images come from at kubeadmconfig top level that applies to all images?

dkoshkin:
> @dkoshkin do we really need to be able to override DNS image repo here? Is it not enough to just be able to override the registry images come from at kubeadmconfig top level that applies to all images?

I had the same thought for this repo, DKP has a very specific usecase because of how FIPS support works, but for a more generic solution I think having a variable for the registry is much more valuable.

jimmidyson:
For the fips case do we really need to override the DNS registry via `"/spec/template/spec/kubeadmConfigSpec/clusterConfiguration/dns/imageRepository"` or is it enough to be able just to override `"/spec/template/spec/kubeadmConfigSpec/clusterConfiguration/imageRepository"` (https://github.com/kubernetes-sigs/cluster-api/blob/82eff49867008365d3b26f82b55ff29a67880aa7/bootstrap/kubeadm/api/v1beta1/kubeadm_types.go#L139)? Why do we need to override the individual component image repositories (which really should be image registry iiuc...)?

jimmidyson:
We could also just solve that with inline patches on fips cluster classes perhaps in addition to generic external patches 🤔

jimmidyson:
Ah I see the note that coredns should always be pulled from default k8s registry

jimmidyson:
Closing - as discussed above this does not appear to be a common enough use case to provide a patch for. This can be handled via inline patches, or a different patch webhook deployment.

PR #134, opened 2023-09-05
refactor: Use controller manager to start runtime hooks server

jimmidyson:
This allows us to more easily use a caching client (less overhead on
API server) and simplifies startup/shutdown.

This commit also removes the ability to configure what hooks are enabled
as they are all opt-in in various ways (e.g. adding patches to cluster,
labels for CNI, etc) so this felt redundant and again simplifies things,


mhrabovcin (approved):
Nice!

PR #140, opened 2023-09-07
feat: add variable grouping

mhrabovcin:
Adds new API that allows to expose group of variables under single handler grouped under single name:

```
group := variables.Group{
  "http": HTTPProxyVariables{},
  "extra_sans": ExtraAPIServerCertSANsVariables{}
}
variables.NewAsSingleVariableHandler("dkp", group, false)
```


mhrabovcin:
We've talked about this and we'll be going different direction.

PR #142, opened 2023-09-11
feat: Add ClusterConfig variable and patch handler

jimmidyson:


docs/content/http-proxy.md
```diff
@@ -34,7 +34,7 @@ spec:
       - name: proxy
         values:
           http: http://example.com
-          https: http://example.com
+          https: https://example.com
```
mhrabovcin, line 37: Nit: you can have http endpoint for https proxy

pkg/handlers/httpproxy/inject.go
```diff
@@ -40,14 +41,21 @@ const (
...
 var (
 	_ handlers.Named           = &httpProxyPatchHandler{}
 	_ mutation.GeneratePatches = &httpProxyPatchHandler{}
 )
 
-func NewPatch(cl ctrlclient.Reader) *httpProxyPatchHandler {
+func NewPatch(
+	cl ctrlclient.Reader,
+	variableName string,
+	variableFieldPath ...string,
```
mhrabovcin, line 57: This is repeated between handlers. Can't the meta generate patch figure out how to map variables?

mhrabovcin (approved):
Amazing!

docs/content/http-proxy.md
jimmidyson: Good point, reverted.

pkg/handlers/httpproxy/inject.go
jimmidyson: Possibly - any ideas how?

mhrabovcin: Some code needs to be aware about the relationship between our the `ClusterConfig` variable and partial variables placement and the handler that is interested in the variable. I can only thing about creating additional code where we'd capture this mapping and I am not sure if it would be any better compare to handler knowing where to reach for given variable.

common/pkg/capi/clustertopology/handlers/mutation/meta.go
```diff
@@ -0,0 +1,48 @@
...
+) {
+	for _, h := range mgp.wrappedHandlers {
+		wrappedResp := &runtimehooksv1.GeneratePatchesResponse{}
+		h.GeneratePatches(ctx, req, wrappedResp)
+		resp.Items = append(resp.Items, wrappedResp.Items...)
+		if wrappedResp.Message != "" {
+			resp.Message = strings.TrimPrefix(resp.Message+"\n"+wrappedResp.Message, "\n")
+		}
+		if wrappedResp.Status != runtimehooksv1.ResponseStatusSuccess {
+			wrappedResp.Status = runtimehooksv1.ResponseStatusSuccess
```
mhrabovcin, line 44: What is this code doing?

jimmidyson: Something that it shouldn't, let me fix it. It was supposed to propagate the first error received.

jimmidyson: Fixed now, I hope.

mhrabovcin: Can you add unit test for this fix? 🙏

jimmidyson: Yeah, I was being lazy :sweat:

jimmidyson: Added unit tests now.

PR #148, opened 2023-09-12
build: add tooling to generate examples files

dkoshkin:
Added a new target `make examples.sync` and `gojq` changes to enable this hook. The idea is that as we add more patches that can replace the default ones in the capi-quickstart we can remove the patches in the ClusterClass. Using `clusterctl` also makes it simpler to keep up with upstream sample.


mhrabovcin (approved):

jimmidyson:
I was wondering what the kustomize patches approach would look like? I might take a stab at that before we merge this. Personally I feel that kustomize patches feels more k8s idiomatic and more understandable than using gojq.

jimmidyson:
@dkoshkin @mhrabovcin I pushed a follow up commit to this PR to use kustomize instead of gojq - let me know what you think, and we can revert if we prefer the gojq approach.

hack/examples/bases/capd/kustomization.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+configurations:
+- kustomizeconfig.yaml
+
+resources:
+- https://github.com/kubernetes-sigs/cluster-api/releases/download/v1.5.1/clusterclass-quick-start.yaml
+- https://github.com/kubernetes-sigs/cluster-api/releases/download/v1.5.1/cluster-template-development.yaml
```
dkoshkin, line 12: nit: it would be cool to get the version from the installed `clusterctl`

hack/examples/bases/capd/kustomization.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+namePrefix: capd-
```
dkoshkin, line 14: 👍

examples/capi-quickstart/capd-cluster.yaml
```diff
@@ -18,26 +17,14 @@ spec:
...
-      - name: etcdImageTag
-        value: ""
-      - name: coreDNSImageTag
-        value: ""
-      - name: podSecurityStandard
-        value:
-          audit: restricted
-          enabled: true
-          enforce: baseline
-          warn: restricted
```
dkoshkin: The other variables are optional, but this one is still needed for Calico to come up right?

make/examples.mk
```diff
@@ -1,7 +1,6 @@
 # Copyright 2023 D2iQ, Inc. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-
 KUBERNETES_VERSION ?= v1.27.5
```
dkoshkin, line 4: Lets either remove this `KUBERNETES_VERSION ` var or use it in the kustomize

examples/capi-quickstart/capd-cluster.yaml
jimmidyson: :thinking: It didn't seem to be. I tested this locally and it worked. Would have to think about why though to properly understand.

hack/examples/bases/capd/kustomization.yaml
jimmidyson: Yes! But like I said, I don't know how to use variables in kustomizations :facepalm:

make/examples.mk
jimmidyson: I want to use it! But how?

examples/capi-quickstart/capd-cluster.yaml
dkoshkin: Same it worked for me too 🤔

make/examples.mk
```diff
@@ -0,0 +1,6 @@
+# Copyright 2023 D2iQ, Inc. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+.PHONY: examples.sync
+examples.sync: ## Syncs the examples by fetching upstream examples and applying kustomize patches
+	hack/examples/sync.sh $(KUBERNETES_VERSION)
```
jimmidyson, line 6: I don't see where the `$(KUBERNETES_VERSION)` argument is used here?

dkoshkin: Good catch! GH Copilot added it and I forgot to remove it 😄 . But at this point it doesn't provide much value since there is only a single example. I would prefer to wait until there is more.

PR #149, opened 2023-09-12
feat: new Kubernetes image registry patch

dkoshkin:
My first external patch and a example PR of the changes to migrate an inline patch from the quickstart to an external patch.

Depends on https://github.com/d2iq-labs/capi-runtime-extensions/pull/148


api/v1alpha1/clusterconfig_types.go
```diff
@@ -25,6 +25,9 @@ type ClusterConfigSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
 
+	// +optional
+	ImageRepository *ImageRepository `json:"imageRepository,omitempty"`
```
mhrabovcin, line 29:
```suggestion
	// +optional
	KubernetesImageRepository *ImageRepository `json:"kubernetesImageRepository,omitempty"`
```

Is the image repository some known term in CAPI context? If this value will affect only some images the name could be more specific.

docs/content/image-repository.md
```diff
@@ -0,0 +1,40 @@
...
+external patches on `ClusterClass`.
+
+```yaml
+apiVersion: cluster.x-k8s.io/v1beta1
+kind: ClusterClass
+metadata:
+  name: <NAME>
+spec:
+  patches:
+    - name: apiserver-cert-sans
```
mhrabovcin, line 17:
```suggestion
    - name: image-repository
```

docs/content/image-repository.md
```diff
@@ -0,0 +1,40 @@
+---
+title: "Image Repository"
+---
+
+Override the container image repository used when pulling Kubernetes images.
+
+To enable the API server certificate SANs enable the `extraapiservercertsansvars` and `extraapiservercertsanspatch`
+external patches on `ClusterClass`.
```
mhrabovcin, line 8:
```suggestion
```

docs/content/image-repository.md
```diff
@@ -0,0 +1,40 @@
...
+  name: <NAME>
+spec:
+  patches:
+    - name: apiserver-cert-sans
+      external:
+        generateExtension: "imagerepositorypatch.capi-runtime-extensions"
+        discoverVariablesExtension: "imagerepositoryvars.capi-runtime-extensions"
+```
+
+On the cluster resource then specify desired certificate SANs values:
```
mhrabovcin, line 23:
```suggestion
On the cluster resource then specify desired Kubernetes image repository value:
```

make/examples.mk
```diff
@@ -20,6 +21,8 @@ examples.sync: kind.create clusterctl.init
       --kubernetes-version $(KUBERNETES_VERSION) \
       --control-plane-machine-count=1 \
       --worker-machine-count=1 | \
+    gojq --yaml-input --yaml-output \
+      '. | (select(.kind=="ClusterClass").spec.patches|= del(.[] | select(.name == "imageRepository")))' | \
```
mhrabovcin, line 25: I like this, in the end we should be able to delete all patches and add our single external patch.

mhrabovcin (approved):

api/v1alpha1/clusterconfig_types.go
jimmidyson: Agreed, I would prefer to make this name more explicit naming, including changing it from `Repository` to `Registry`.

dkoshkin: 👍 I used `imageRepository` because thats what the kubeadm config has but it the benefit of the context being part of `kubeadmConfigSpec/clusterConfiguration` struct. I'll make it more explicit here.

docs/content/image-repository.md
dkoshkin: You're seeing how the sausage is made by most copy and pasting for new handlers, we would really benefit from a kubebuilder style generator.

api/v1alpha1/clusterconfig_types.go
dkoshkin: @jimmidyson the `Repository` vs `Registry` doesn't bother me too much, I can see why in kubeadm it is called `repository`, but I think "registry" is more commonly used in Kubernetes docs, I'll change it too.

jimmidyson (approved):

jimmidyson:
Nice work!

PR #151, opened 2023-09-14
refactor: Use controller manager options for pprof handler

jimmidyson:


mhrabovcin (approved):
🚀 https://eng.d2iq.com/blog/profiling-kubernetes-controllers-with-pprof/

PR #152, opened 2023-09-14
feat: CNI provider deployment via variables instead of labels

jimmidyson:
As discussed, let's try to use variables to drive addon deployments rather than labels. Ultimately this still uses labels for the ClusterResourceSet selectors but we can rely on the `cluster.x-k8s.io/cluster-name` label for the selector.

Missing tests, sorry, just wanted to push this and will finish this off on Monday.


dkoshkin (approved):
I like it!

PR #154, opened 2023-09-14
refactor: how handlers are added to server

dkoshkin:
When adding new handlers I kept forgetting to add it to the meta hander. Although this commit doesn't really solve it does make it easier to visually see a difference. It also simplifies adding handlers to the server by not asking for a package variable when calling `<patchpackage>.NewPatch()`.

Ideally we should be able to automatically build the metahandler from `patchHandlers` but I didn't see how to do that with the 2 interfaces we have now `Named` and `GeneratePatches`.


jimmidyson (approved):
Love it!

PR #159, opened 2023-09-19
fix: Handle multiple meta mutators cleanly

jimmidyson:
This applies each of the mutators before generating patches.


common/pkg/capi/clustertopology/handlers/mutation/meta.go
```diff
@@ -5,22 +5,44 @@ package mutation
...
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	bootstrapv1 "sigs.k8s.io/cluster-api/bootstrap/kubeadm/api/v1beta1"
+	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
 	runtimehooksv1 "sigs.k8s.io/cluster-api/exp/runtime/hooks/api/v1alpha1"
+	"sigs.k8s.io/cluster-api/exp/runtime/topologymutation"
 
 	"github.com/d2iq-labs/capi-runtime-extensions/common/pkg/capi/clustertopology/handlers"
 )
 
+type MetaMutater interface {
```
jimmidyson, line 20: @mhrabovcin Need to figure out how to massage http proxy handler to implement this interface. Problem is the cluster retrieval needs access to the request. Any ideas?

jimmidyson: OK figured it out, set cluster key in context

common/pkg/capi/clustertopology/handlers/cluster_context.go
```diff
@@ -0,0 +1,48 @@
...
+	"context"
+	"errors"
+
+	capiv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	runtimehooksv1 "sigs.k8s.io/cluster-api/exp/runtime/hooks/api/v1alpha1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// clusterKeyContextKey is how we find a cluster key from a context.Context.
+type clusterKeyContextKey struct{}
```
dkoshkin: Just curious why a struct instead of a string constant?

common/pkg/capi/clustertopology/handlers/cluster_context.go
```diff
@@ -0,0 +1,48 @@
...
+	if !ok || clusterKey.Name == "" {
+		return client.ObjectKey{}, errors.New(
+			"failed to detect cluster name from GeneratePatch request",
+		)
+	}
+
+	return clusterKey, nil
+}
+
+func ClusterKeyInto(
```
dkoshkin: I was a little confused by the function name and the signature that it takes in a context and then returns a context, is there a more descriptive function name?

```suggestion
func SetClusterKeyInto(
```

mhrabovcin (approved):
Awesome!

dkoshkin (approved):
Great changes ❤️

common/pkg/capi/clustertopology/handlers/cluster_context.go
jimmidyson: To prevent conflicts. As this is just an arbitrary string, other code could inject stuff into this key. Using an unexported struct like this makes that safer.

dkoshkin:
Verified manually too and seeing HTTPProxy on the control-plane

PR #164, opened 2023-09-20
feat: adds nfd 

faiq:


api/v1alpha1/clusterconfig_types.go
```diff
@@ -213,6 +216,17 @@ func (CNI) VariableSchema() clusterv1.VariableSchema {
 	}
 }
 
+// NFD tells us to enable or disable the node feature discovery addon.
+type NFD struct{}
+
+func (NFD) VariableSchema() clusterv1.VariableSchema {
+	return clusterv1.VariableSchema{
+		OpenAPIV3Schema: clusterv1.JSONSchemaProps{
+			Type: "Object",
```
mhrabovcin, line 225:
```suggestion
			Type: "object",
```

jimmidyson: Perhaps we could add some builders for this structure?

jimmidyson (approved):

pkg/handlers/nfd/handler.go
```diff
@@ -143,11 +146,7 @@ func (n *DefaultNFD) ensureNFDConfigmapForCluster(
 	}
 	err = client.ServerSideApply(ctx, n.client, cmForCluster)
 	if err != nil {
-		return nil, fmt.Errorf("failed to server side apply %w", err)
-	}
-	err = n.client.Get(ctx, ctrlclient.ObjectKeyFromObject(cmForCluster), cmForCluster)
-	if err != nil {
-		return nil, fmt.Errorf("failed to fetch server side applied configmap %w", err)
+		return nil, fmt.Errorf("failed to apply NFD ConfigMap for cluster: %w", err)
```
faiq, line 149: Interesting, we don't need to get a fresh copy after the server side apply?

jimmidyson: Oh wait sorry no you don't need to refetch it, it's updated by the `ServerSideApply` call. Ultimately this calls `client.Patch` which is documented as `Patch patches the given obj in the Kubernetes cluster. obj must be a struct pointer so that obj can be updated with the content returned by the Server` so following a patch the passed in object is updated.

jimmidyson: This is the same semantics as `Create` or `Update`, where the passed in object is used in the request and updated with the response before returning to the caller.

faiq: Awesome! Good change then. thank you

docs/content/nfd.md
```diff
@@ -0,0 +1,41 @@
...
+  name: <NAME>
+spec:
+  patches:
+    - name: cluster-config
+      external:
+        generateExtension: "clusterconfigpatch.capi-runtime-extensions"
+        discoverVariablesExtension: "clusterconfigvars.capi-runtime-extensions"
+```
+
+On the cluster resource then specify this `cni` value:
```
dkoshkin, line 27:
```suggestion
On the cluster resource then specify this `nfd` value:
```

dkoshkin (approved):
The code looks good but I didn't manually test. @faiq did you create a cluster and saw the NFD pod getting deployed?

jimmidyson:
@dkoshkin I tested it and it worked, shared proof in slack 😅

PR #172, opened 2023-09-21
feat: AWS cluster config patch

jimmidyson:
AWS cluster config is marked as required and region is defaulted via OpenAPI
definition.


api/v1alpha1/aws_clusterconfig_types.go
```diff
@@ -0,0 +1,119 @@
...
+				"region": Region("").VariableSchema().OpenAPIV3Schema,
+			},
+			Required: []string{"region"},
+		},
+	}
+}
+
+type Region string
+
+const (
```
faiq, line 44: We should drop this. Neither upstream projects nor the aws-sdk-go have this sort of validation as far as i can tell.

dkoshkin: ➕

examples/capi-quick-start/aws-cluster-class.yaml
```diff
@@ -25,20 +25,10 @@ spec:
...
-              valueFrom:
-                template: |
-                  {{ if .region }}{{ .region }}{{ else }}us-west-2{{ end }}
-          selector:
-            apiVersion: infrastructure.cluster.x-k8s.io/v1beta2
-            kind: AWSClusterTemplate
-            matchResources:
-              infrastructureCluster: true
-      description: aws region to target for cluster creation
-      name: region
```
dkoshkin, line 41: 👋  we barely knew ya

api/v1alpha1/aws_clusterconfig_types.go
jimmidyson: Aww i liked this :joy: Without this, if a dodgy region is specified then the cluster won't get rejected and we'd need to figure out how to surface the issue to the user.

api/v1alpha1/aws_clusterconfig_types.go
```diff
@@ -0,0 +1,119 @@
...
+	Spec AWSClusterConfigSpec `json:"spec,omitempty"`
+}
+
+// AWSClusterConfigSpec defines the desired state of AWSClusterConfig.
+type AWSClusterConfigSpec struct {
+	// +optional
+	Region *Region `json:"region,omitempty"`
+}
+
+func (AWSClusterConfigSpec) VariableSchema() clusterv1.VariableSchema {
```
dkoshkin, line 29: Is this missing public docs?

examples/capi-quick-start/aws-cluster.yaml
```diff
@@ -24,6 +24,8 @@ spec:
             cni:
               provider: calico
             nfd: {}
+      - name: awsClusterConfig
```
dkoshkin, line 27: 👍 I like having it a separate variable, it enables us to only enable the provider specific handler, no need to have docker cod in the codepath for an AWS CluterClass.

hack/examples/bases/aws/kustomization.yaml.tmpl
```diff
@@ -78,15 +69,14 @@ patches:
                   infrastructureCluster: true
           description: AWSClusterStaticIdentity identityRef to use when creating the cluster
     - op: "add"
-      path: "/spec/variables"
+      path: "/spec/patches/0"
```
dkoshkin, line 72: Lol I thought I knew how these patches worked but obviously I don't 😆 . Doesn't `/spec/patches/0` mean to put it in the [0] element in the slice, how does the example ClusterClass file end up with this as the second variable?

dkoshkin (approved):

hack/examples/bases/aws/kustomization.yaml.tmpl
jimmidyson: It's an insert before the specified index rather than an overwrite of the element at specified index.

examples/capi-quick-start/aws-cluster.yaml
jimmidyson: I'm thinking we can combine the two and have single variable for each provider that combines generic config and provider specific config. The variable will be the same name regardless of provider but have a different schema.

api/v1alpha1/aws_clusterconfig_types.go
jimmidyson: We could allow it to be configurable via the helm chart, and default to anything. For now though I think I'll just remove it and open an issue for future improvements.

jimmidyson: Removed. We can set this at the runtime hook setting in the clusterclass if need be. I'll work on that later, more as an example of how we can use settings than a necessary feature though.

api/v1alpha1/aws_clusterconfig_types.go
jimmidyson: Yes - is it ok if i add them after I combine the generic and provider patch hooks though?

jimmidyson: I'm gonna merge and I will follow up with docs in follow up PR.

PR #173, opened 2023-09-22
feat: Combine generic variables with provider specific variables

jimmidyson:
1. Combine generic variables with provider specific variables, leading to a
   single variable per provider. This is provided by a separate discover
   variables hook per provider that will have to be referenced by name in the
   clusterclass.
2. The name of this single variable will always be `clusterConfig` regardless of
   provider, but 1 will ensure that this variable has only the relevant fields
   for the cluster provider. Having the same variable name gives a consistent
   feel across providers.
3. Remove registration of individual patches and only keep meta patches. This
   simplifies things. We keep the implementations separate for better
   testability and also to allow for inclusion in other libraries.

Depends on #172 .


dkoshkin (dismissed):

docs/content/audit-policy.md
```diff
@@ -4,10 +4,10 @@ title: "Audit policy"
 
 Kubernetes auditing provides a security-relevant, chronological set of records documenting the sequence of actions in a
 cluster. The cluster audits the activities generated by users, by applications that use the Kubernetes API, and by the
-control plane itself. The `auditpolicypatch` external patch will generate appropriate configuration for the Kubernetes
+control plane itself. The `clusterconfigpatch` external patch will generate appropriate configuration for the Kubernetes
 control plane.
 
-To enable the audit policy enable the `auditpolicypatch` external patch on `ClusterClass`.
+To enable the meta handler enable the `clusterconfigvars` and `clusterconfigpatch` external patches on `ClusterClass`.
```
jimmidyson, line 10: Actually this would be provider specific eg awsclusterconfig which includes the necessary variables. I think we should drop the individual patch docs and only have the provider specific docs as this project only exposes them with this pr

dkoshkin: Going to delete this commit, we can update the docs in a separate PR.

jimmidyson: Thanks for doing it anyway!

dkoshkin (approved):

PR #174, opened 2023-09-22
feat: add imageRegistryCredentials handler

dkoshkin:
This is big boy 😬
Adds a handler to set image registry credentials on the Cluster using [Kubelet image credential provider](https://kubernetes.io/docs/tasks/administer-cluster/kubelet-credential-provider/)
and [dynamic credential provider](https://github.com/mesosphere/dynamic-credential-provider).
```
      - name: clusterConfig
        value:
          imageRegistryCredentials:
            url: https://my-registry.io
            secret: my-registry-credentials
``` 

This handler also expects a Secret with keys `username` and `password` to already exist. Alternatively we can simplify this and expect an existing Secret as [CAPI SecretFileSource ](https://github.com/kubernetes-sigs/cluster-api/blob/main/bootstrap/kubeadm/api/v1beta1/kubeadmconfig_types.go#L497) that the handler doesn't need to read/write the Secret. But that didn't seem right as the format of the Secret would need to be very specific to the [dynamic credential provider](https://github.com/mesosphere/dynamic-credential-provider) project.

---
Also verified by create a docker cluster and checking the `KubeadmConfig`


api/v1alpha1/clusterconfig_types.go
```diff
@@ -186,6 +191,40 @@ func (ExtraAPIServerCertSANs) VariableSchema() clusterv1.VariableSchema {
...
+// ImageRegistryCredentials required for providing credentials for an image registry URL.
+type ImageRegistryCredentials struct {
+	// Registry URL.
+	URL string `json:"url"`
+
+	// Name of the Secret containing the registry credentials.
+	// The Secret should have keys 'username' and 'password'.
+	// This credentials Secret is not required for some registries, e.g. ECR.
+	// +optional
+	Secret string `json:"secret,omitempty"`
```
dkoshkin, line 203: Looking for suggestions on what this should look like 🙏

common/pkg/testutils/capitest/request/items.go
```diff
@@ -41,6 +49,10 @@ func NewKubeadmConfigTemplateRequestItem(uid types.UID) runtimehooksv1.GenerateP
 				Kind:       "KubeadmConfigTemplate",
 				APIVersion: bootstrapv1.GroupVersion.String(),
 			},
+			ObjectMeta: metav1.ObjectMeta{
```
dkoshkin, line 52: The handler uses this name and namespace to create a new Secret with the credentials.

api/v1alpha1/clusterconfig_types.go
```diff
@@ -186,6 +191,40 @@ func (ExtraAPIServerCertSANs) VariableSchema() clusterv1.VariableSchema {
 	}
 }
 
+// ImageRegistryCredentials required for providing credentials for an image registry URL.
+type ImageRegistryCredentials struct {
```
dkoshkin, line 195: I've kept this struct independent from the related Containerd mirrors configuration.

jimmidyson: Great idea!

api/v1alpha1/clusterconfig_types.go
jimmidyson: I think making this a `secretRef` (like an `ObjectReference`) just with `name` field for now would make more sense. If we make that it's own struct then we'll be able to reuse that as well.

api/v1alpha1/clusterconfig_types.go
```diff
@@ -186,6 +191,40 @@ func (ExtraAPIServerCertSANs) VariableSchema() clusterv1.VariableSchema {
...
+	// Registry URL.
+	URL string `json:"url"`
```
jimmidyson, line 197: I'm not sure URL is quite right here as technically this could be any `matchImages` glob couldn't it?

api/v1alpha1/clusterconfig_types.go
```diff
@@ -39,6 +39,10 @@ type ClusterConfigSpec struct {
 	// +optional
 	ExtraAPIServerCertSANs ExtraAPIServerCertSANs `json:"extraAPIServerCertSANs,omitempty"`
 
+	// TODO: Add support for multiple registries.
+	// +optional
+	ImageRegistryCredentials ImageRegistryCredentials `json:"imageRegistryCredentials,omitempty"`
```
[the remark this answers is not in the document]
jimmidyson: Yeah that looks good to me.

api/v1alpha1/clusterconfig_types.go
dkoshkin: Now I don't know anymore lol 🤔 

This PR adds the following:
```
  matchImages:
  - {{ printf "%q" .MirrorHost }}
  - "*"
  - "*.*"
  - "*.*.*"
  - "*.*.*.*"
  - "*.*.*.*.*"
  - "*.*.*.*.*.*"
```
https://github.com/d2iq-labs/capi-runtime-extensions/pull/174/files#diff-3996a5b712640bdb338abbda2b49ab7519154b412bdb3072be1c65e5d8917dccR13-R18

But that only makes sense if the registry is also a mirror.

With https://github.com/d2iq-labs/capi-runtime-extensions/pull/174#discussion_r1335902058 its also getting pretty difficult to reason what all these configurations are for and how to use globs as URLs or try to match up a glob to a registry mirror.

Maybe we want something like this instead?:
```
imageRegistries:
  credentials:
  - url: [https://]myregistry.com # intentionally a URL with a potential schema and not an image glob
    username: <>
    password: <>
    mirror: false
```

jimmidyson: These credentials should only be in the dynamic-credential-provider config file though right? And the URL/glob could map to a single username/password so I think this is OK. I don't think mirror makes any difference here.

dkoshkin: Discussed this offline, to keep it simple at first, we will ask for a URL with the credentials.
Then when image registry mirrors are implemented, we will use the URL to also configure Containerd and the DynamicCredentialProvider configuration as a mirror.

pkg/handlers/generic/mutation/imageregistries/credentials/templates/kubelet-image-credential-provider-config.yaml.gotmpl
```diff
@@ -9,7 +9,6 @@ providers:
   {{- end }}
   {{- end }}
   matchImages:
-  - {{ printf "%q" .MirrorHost }}
```
dkoshkin: Send all images to the DynamicCredentialProvider.

pkg/handlers/generic/mutation/imageregistries/credentials/templates/dynamic-credential-provider-config.yaml.gotmpl
```diff
@@ -1,13 +1,6 @@
 apiVersion: credentialprovider.d2iq.com/v1alpha1
 kind: DynamicCredentialProviderConfig
 credentialProviderPluginBinDir: /etc/kubernetes/image-credential-provider/
-{{- with .MirrorHost }}
-{{- if and (ne . "registry-1.docker.io") (ne . "docker.io") }}
-mirror:
-  endpoint: {{ printf "%q" . }}
-  credentialsStrategy: MirrorCredentialsOnly
-{{- end }}
-{{- end }}
```
dkoshkin: Only configure it a a mirror based on a different config (TODO)

dkoshkin:
```
    - content: |
        apiVersion: kubelet.config.k8s.io/v1beta1
        kind: CredentialProviderConfig
        providers:
        - name: dynamic-credential-provider
          args:
          - get-config
          - -c
          - /etc/kubernetes/dynamic-credential-provider-config.yaml
          matchImages:
          - "*"
          - "*.*"
          - "*.*.*"
          - "*.*.*.*"
          - "*.*.*.*.*"
          - "*.*.*.*.*.*"
          defaultCacheDuration: "0s"
          apiVersion: credentialprovider.kubelet.k8s.io/v1beta1
```

dkoshkin:
Tested with a private image in a Docker cluster:
```
Events:
  Type     Reason                  Age                From               Message
  ----     ------                  ----               ----               -------
  Normal   Scheduled               42s                default-scheduler  Successfully assigned default/nginx to docker-quick-start-md-0-krcqf-25k4v-r5k84
  Warning  FailedCreatePodSandBox  41s                kubelet            Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "9b29e435643db3670bc4185f354708dc719157d6f1e2de121f86a992ebac1ba2": plugin type="calico" failed (add): stat /var/lib/calico/nodename: no such file or directory: check that the calico/node container is running and has mounted /var/lib/calico/
  Normal   SandboxChanged          27s (x2 over 41s)  kubelet            Pod sandbox changed, it will be killed and re-created.
  Normal   Pulling                 26s                kubelet            Pulling image "dkoshkin/nginx:stable"
  Normal   Pulled                  9s                 kubelet            Successfully pulled image "dkoshkin/nginx:stable" in 9.280283022s (16.86162273s including waiting)
  Normal   Created                 9s                 kubelet            Created container nginx
  Normal   Started                 9s                 kubelet            Started container nginx
```

api/v1alpha1/clusterconfig_types.go
```diff
@@ -28,6 +29,9 @@ type GenericClusterConfig struct {
 	// +optional
 	ExtraAPIServerCertSANs ExtraAPIServerCertSANs `json:"extraAPIServerCertSANs,omitempty"`
 
+	// +optional
+	ImageRegistries ImageRegistries `json:"imageRegistries,omitempty"`
```
jimmidyson, line 33:
```suggestion
	ImageRegistries *ImageRegistries `json:"imageRegistries,omitempty"`
```

api/v1alpha1/clusterconfig_types.go
```diff
@@ -175,6 +180,88 @@ func (ExtraAPIServerCertSANs) VariableSchema() clusterv1.VariableSchema {
...
+	ImageRegistryCredentials ImageRegistryCredentials `json:"credentials,omitempty"`
+}
+
+func (ImageRegistries) VariableSchema() clusterv1.VariableSchema {
+	return clusterv1.VariableSchema{
+		OpenAPIV3Schema: clusterv1.JSONSchemaProps{
+			Description: "Configuration for image registries.",
+			Type:        "object",
+			Properties: map[string]clusterv1.JSONSchemaProps{
+				"credentials": imageRegistryCredentialsSchema,
```
jimmidyson, line 194:
```suggestion
				"credentials": ImageRegistryCredentials{}.VariableSchema().OpenAPIV3Schema,
```

api/v1alpha1/clusterconfig_types.go
```diff
@@ -175,6 +180,88 @@ func (ExtraAPIServerCertSANs) VariableSchema() clusterv1.VariableSchema {
...
+		},
+	}
+}
+
+var (
+	imageRegistryCredentialsSchema = clusterv1.JSONSchemaProps{
+		Type:        "array",
+		UniqueItems: true,
+		Items:       &imageRegistryCredentialsResourceSchema,
+	}
```
jimmidyson, line 205: Embed this in `func (ImageRegistryCredentials) VariableSchema()`

api/v1alpha1/clusterconfig_types.go
```diff
@@ -175,6 +180,88 @@ func (ExtraAPIServerCertSANs) VariableSchema() clusterv1.VariableSchema {
...
+					"name": {
+						Description: "The name of the Secret containing the registry credentials.",
+						Type:        "string",
+					},
+					"namespace": {
+						Description: "The namespace of the Secret containing the registry credentials. " +
+							"Defaults to the namespace of the KubeadmControlPlaneTemplate and KubeadmConfigTemplate" +
+							" that reference this variable.",
+						Type: "string",
+					},
```
jimmidyson, line 231: Let's make this a `ObjectReference` schema so we can reuse it in other places when required.

jimmidyson (approved):

dkoshkin:
Thank you @jimmidyson for the thorough review and the fixups!

jimmidyson:
@dkoshkin Can we remove the `do not -merge` label or are you still testing stuff out?

dkoshkin:
🚢

PR #187, opened 2023-09-25
fix: bring back missing docker handlers

dkoshkin:
Structured this in the same way as the AWS handlers.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -7,6 +7,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"slices"
 
 	"github.com/spf13/pflag"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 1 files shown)


cmd/main.go
```diff
@@ -150,7 +150,12 @@ func main() {
 
 	// dockerMetaPatchHandlers combines all Docker patch and variable handlers under a single handler.
 	// It allows to specify configuration under a single variable.
-	dockerMetaPatchHandlers := []mutation.MetaMutater{}
+	dockerMetaPatchHandlers := append(
+		[]mutation.MetaMutater{
+			region.NewMetaPatch(),
```
jimmidyson, line 155: We don't need this, aws specific.

jimmidyson (changes requested):
Thanks for fixing this 🙏

cmd/main.go
dkoshkin: Oops copy and paste without really looking, fixing 🤦‍♂️

jimmidyson (approved):

PR #188, opened 2023-09-26
feat: Use external patch for Docker provider custom image

jimmidyson:
This required introducing some more infrastructure provider selectors which will be useful going forward.

Also fixes a bug in generating patches for providers using unstructured decoder, previously patches wouldn't be applied and silently ignored.


api/v1alpha1/docker_clusterconfig_types.go
```diff
@@ -21,11 +25,21 @@ type DockerClusterConfig struct {
 // DockerClusterConfigSpec defines the desired state of DockerClusterConfig.
 type DockerClusterConfigSpec struct {
 	GenericClusterConfig `json:",inline"`
+
+	//+optional
+	CustomImage *string `json:"customImage,omitempty"`
```
dkoshkin, line 30: Do we want this as `OCIImage`?
```suggestion
	CustomImage *OCIImage `json:"customImage,omitempty"`
```

common/pkg/capi/clustertopology/patches/selectors/selectors.go
```diff
@@ -20,46 +20,70 @@ func ControlPlane() clusterv1.PatchSelector {
...
 				},
 			},
 		},
 	}
 }
 
-func AllWorkersSelector() clusterv1.PatchSelector {
+// InfrastructureCluster selector matches against infrastructure clusters.
+// Passing in the API version (not the API group) is required because different providers could support different API
+// versions. This also allows for a oatch to select multiple infrastructure versions for the same provider.
```
dkoshkin, line 39:
```suggestion
// versions. This also allows for a patch to select multiple infrastructure versions for the same provider.
```

common/pkg/capi/clustertopology/patches/selectors/selectors.go
```diff
@@ -20,46 +20,70 @@ func ControlPlane() clusterv1.PatchSelector {
...
+		Kind: kind,
+		MatchResources: clusterv1.PatchSelectorMatch{
+			InfrastructureCluster: true,
+		},
+	}
+}
+
+// InfrastructureWorkerMachineTemplates selector matches against infrastructure machines.
+// Passing in the API version (not the API group) is required because different providers could support different API
+// versions. This also allows for a oatch to select multiple infrastructure versions for the same provider.
```
dkoshkin, line 55:
```suggestion
// versions. This also allows for a patch to select multiple infrastructure versions for the same provider.
```

common/pkg/capi/clustertopology/patches/selectors/selectors.go
```diff
@@ -20,46 +20,70 @@ func ControlPlane() clusterv1.PatchSelector {
...
-					"*",
-				},
+				Names: []string{"*"},
 			},
 		},
 	}
 }
 
-// InfrastructureCluster selector matches against infrastructure clusters.
+// InfrastructureWorkers selector matches against infrastructure machines.
```
dkoshkin, line 73:
```suggestion
// InfrastructureControlPlaneMachines selector matches against infrastructure control-plane machines.
```

pkg/handlers/docker/mutation/customimage/inject.go
```diff
@@ -0,0 +1,215 @@
...
+				kubernetesVersion, found, err := variables.Get[string](
+					vars,
+					fieldPath[0],
+					fieldPath[1:]...)
+				if err != nil {
+					return err
+				}
+				if !found {
+					return fmt.Errorf(
+						"missing required builtin variable: %s",
```
mhrabovcin, line 120: Why `builtin` ?

pkg/handlers/docker/mutation/customimage/inject.go
```diff
@@ -0,0 +1,215 @@
...
+		customImageVar,
+	)
+
+	err = patches.Generate(
+		obj,
+		vars,
+		&holderRef,
+		selectors.InfrastructureWorkerMachineTemplates("v1beta1", "DockerMachineTemplate"),
+		log,
+		func(obj *unstructured.Unstructured) error {
```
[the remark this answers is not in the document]
jimmidyson: By virtue of being applied on docker cluster class, yes I think so, but I agree there should be a check here just in case it gets applied to a wrong cluster provider.

dkoshkin (approved):
Thanks!

pkg/handlers/docker/mutation/customimage/inject.go
```diff
@@ -0,0 +1,215 @@
...
+				}
+
+				customImageVar = v1alpha1.OCIImage(
+					defaultKinDImageRepository + ":" + kubernetesVersion,
+				)
+			}
+
+			log.WithValues(
+				"patchedObjectKind", obj.GetObjectKind().GroupVersionKind().String(),
+				"patchedObjectName", client.ObjectKeyFromObject(obj),
```
mhrabovcin, line 132: Can we log the resolved custom image var?

api/v1alpha1/docker_clusterconfig_types.go
jimmidyson: Oops! Thank you!

pkg/handlers/docker/mutation/customimage/inject.go
jimmidyson: Copy-paste error... thanks!

pkg/handlers/docker/mutation/customimage/inject.go
jimmidyson: Done.

mhrabovcin (approved):

pkg/handlers/docker/mutation/customimage/inject.go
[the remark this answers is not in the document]
jimmidyson: Actually, yes the selector can only select objects of kind `DockerMachineTemplate` in the `infrastructure.cluster.x-k8s.io/v1beta1` API version so this is safe

api/v1alpha1/docker_clusterconfig_types.go
jimmidyson: Thanks!

common/pkg/capi/clustertopology/patches/selectors/selectors.go
jimmidyson: Thanks!

PR #190, opened 2023-09-26
feat: vendor infrastructure provider APIs

dkoshkin:
@jimmidyson and I were briefly talking about this idea. Here is an attempt of using infra provider types that are synced from upstream repos. It is easier to work with Go types instead of `unstructured`.
Switched the AWS region handler to use these new types to show that it works.

An important part of this PR is the automated syncing. I added it as makefile targets to keep it simple, but can be extended if no longer enough as we add additional providers.


make/apis.mk
```diff
@@ -0,0 +1,31 @@
+# Copyright 2023 D2iQ, Inc. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+export CAPA_REPO := https://github.com/kubernetes-sigs/cluster-api-provider-aws.git
+export CAPA_VERSION := v2.2.2
+export CAPA_API_PACKAGE_NAME := sigs.k8s.io/cluster-api-provider-aws/v2/api/
```
supershal, line 6: should we limit syncing to api versions that we use? for ex. `v1beta2` here?

dkoshkin: Good idea, done https://github.com/d2iq-labs/capi-runtime-extensions/pull/190/commits/957e6666082c442da169f49870605c5063ba06da

dkoshkin:
https://github.com/d2iq-labs/capi-runtime-extensions/pull/190/commits/aec53639e10adaa6537f937980ad856a23c52a0b here is a fun idea @faiq and I saw in some CAPI provider (don't remember which one), using a separate go.mod for each provider. This has the benefit of using dependabot to keep the version updated.

If we don't want this, we should still bring in the directory change from that commit, I think placing it in the more explicit dir is nicer `common/pkg/external/sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta2/`.

jimmidyson:
This is amazing!

I think I'd prefer to have a separate go module just for provider APIs rather than common. That would make it useful across other projects too, keeping dependencies to a minimum for API module. WDYT?

jimmidyson:
I love this! I've taken the liberty of making a few changes (all can be reverted if you disagree with what I've done btw):

- Use go mod cache paths instead of vendoring

- Moving stuff around a bit (not the final API destination btw): moved placeholder modues into hack/third-party for what feels like a "cleaner", never-import-this, kinda place.

- Added CAPD to show how we can extend this for other providers nice and easily

- Fixed up build stuff, golangci-lint ignoring, etc and added pre-commit hook to ensure it is kept up to date

jimmidyson (approved):

mhrabovcin (approved):

make/examples.mk
```diff
@@ -4,7 +4,7 @@
 export KUBERNETES_VERSION := v1.27.5
 
 export CLUSTERCTL_VERSION := $(shell clusterctl version -o short 2>/dev/null)
-export CAPA_VERSION := $(shell grep -E -e "sigs.k8s.io/cluster-api-provider-aws/v2" external/cluster-api-provider-aws/v2/go.mod | cut -d' ' -f3)
+export CAPA_VERSION := $(shell cd hack/third-party/capa && go list -m -f '{{ .Version }}' sigs.k8s.io/cluster-api-provider-aws/v2)
```
dkoshkin, line 7: Nice

make/go.mk
```diff
@@ -126,9 +126,9 @@ endif
 .PHONY: lint.%
 lint.%: ## Runs golangci-lint for a specific module
 lint.%: ; $(info $(M) linting $* module)
-	$(if $(filter-out root,$*),cd $* && )golines -w $$(go list ./... | sed "s|^$$(go list -m)|.|")
+	$(if $(filter-out root,$*),cd $* && )golines -w $$(go list ./... | grep -v pkg/external | sed "s|^$$(go list -m)|.|")
 	$(if $(filter-out root,$*),cd $* && )golangci-lint run --fix --config=$(GOLANGCI_CONFIG_FILE)
-	$(if $(filter-out root,$*),cd $* && )golines -w $$(go list ./... | sed "s|^$$(go list -m)|.|")
+	$(if $(filter-out root,$*),cd $* && )golines -w $$(go list ./... | grep -v pkg/external | sed "s|^$$(go list -m)|.|")
```
dkoshkin, line 131: Of course! Im stealing this for personal projects :)

.pre-commit-config.yaml
```diff
@@ -22,6 +22,12 @@ repos:
     language: system
     files: "^hack/examples/"
     pass_filenames: false
+  - id: apis-sync
```
dkoshkin, line 25: Good idea!

common/pkg/capi/clustertopology/handlers/mutation/meta.go
```diff
@@ -39,10 +41,14 @@ func NewMetaGeneratePatchesHandler(name string, m ...MetaMutater) handlers.Named
 	scheme := runtime.NewScheme()
 	_ = bootstrapv1.AddToScheme(scheme)
 	_ = controlplanev1.AddToScheme(scheme)
+	_ = capav1.AddToScheme(scheme)
+	_ = capdv1.AddToScheme(scheme)
```
dkoshkin, line 45: Good to know

.golangci.yml
```diff
@@ -1,6 +1,10 @@
 # Copyright 2021-2023 D2iQ, Inc. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+run:
+  skip-dirs:
+    - pkg/external
```
dkoshkin, line 6: Does this need to be?
```suggestion
    - common/pkg/external/
```

jimmidyson: No, because the linter works on a per module basis, in the root module this package isn't linted, only in the common module where this is the correct path to ignore, but it shares linter config...

common/pkg/capi/clustertopology/handlers/mutation/meta.go
jimmidyson: Yeah, the meta handler has to be able to decode all kinds. We do still have the fallback to unstructured decoder for any meta handlers we don't have APIs vendored for.

PR #192, opened 2023-09-27
feat: adds aws ebs config

faiq:
Implements the CSI driver for AWS EBS as well as a "defaultStorageClass" option. 

A user can test this functionality with CAPA clusters with the following under the main cluster config variable

```
   addons: 
            csi:
              providers:
                - name: aws-ebs
              defaultClassName: aws-ebs
```


pkg/handlers/generic/lifecycle/csi/aws-ebs/handler.go
```diff
@@ -0,0 +1,132 @@
...
+	c ctrlclient.Client,
+	cfg *AWSEBSConfig,
+) *AWSEBS {
+	return &AWSEBS{
+		client: c,
+		config: cfg,
+	}
+}
+
+func (a *AWSEBS) EnsureCSIConfigMapForCluster(ctx context.Context, cluster *clusterv1.Cluster) (*corev1.ConfigMap, error) {
```
dkoshkin, line 52: How about we move this into the main CSI handler or some other shared package? And here we can have functions that return driver details.

Things like:
```
func GetName()
func GetConfigMapName()
```

faiq: Maybe `GetName` and `EnsureConfigMap` would be good. I suspect that the code for creating the CRS would be very similar for all of them, so that code can live on the handler level

faiq: I don't even think we need `GetName` now that I think about it more. The user will provide a map to the CSIHandler which maps names to CSI providers

hack/addons/update-aws-ebs-csi.sh
```diff
@@ -0,0 +1,103 @@
+#!/usr/bin/env bash
+set -euo pipefail
```
dkoshkin, line 2: Please update this to match https://github.com/d2iq-labs/capi-runtime-extensions/pull/207 🙏

pkg/handlers/generic/lifecycle/csi/aws-ebs/handler.go
```diff
@@ -0,0 +1,103 @@
...
+			Name:      a.config.defaultAWSEBSConfigMapName,
+		},
+	}
+	defaultAWSEBSCSIConfigMapObjName := ctrlclient.ObjectKeyFromObject(
+		awsEBSCSIConfigMap,
+	)
+	err := a.client.Get(ctx, defaultAWSEBSCSIConfigMapObjName, awsEBSCSIConfigMap)
+	if err != nil {
+		return nil, fmt.Errorf(
+			"failed to retrieve default Tigera Operator manifests ConfigMap %q: %w",
```
dkoshkin, line 69:
```suggestion
			"failed to retrieve default EBS CSI manifests ConfigMap %q: %w",
```

pkg/handlers/generic/lifecycle/csi/aws-ebs/handler.go
```diff
@@ -0,0 +1,103 @@
...
+			defaultAWSEBSCSIConfigMapObjName,
+			err,
+		)
+	}
+
+	awsEBSConfigMap := generateAWSEBSCSIConfigMap(awsEBSCSIConfigMap, cluster)
+	if err := client.ServerSideApply(ctx, a.client, awsEBSConfigMap); err != nil {
+		return nil, fmt.Errorf(
+			"failed to apply Tigera Operator manifests ConfigMap: %w",
```
dkoshkin, line 78:
```suggestion
			"failed to apply EBS CSI manifests ConfigMap: %w",
```

pkg/handlers/generic/lifecycle/csi/aws-ebs/handler.go
```diff
@@ -0,0 +1,103 @@
...
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: cluster.Namespace,
+			Name:      defaultAWSEBSCSIConfigMap.Name,
+		},
+		Data:       defaultAWSEBSCSIConfigMap.Data,
+		BinaryData: defaultAWSEBSCSIConfigMap.BinaryData,
+	}
+
+	return namespacedTigeraConfigMap
```
dkoshkin, line 102:
```suggestion
	namespacedEBSCSIConfigMap := &corev1.ConfigMap{
		TypeMeta: metav1.TypeMeta{
			APIVersion: corev1.SchemeGroupVersion.String(),
			Kind:       "ConfigMap",
		},
		ObjectMeta: metav1.ObjectMeta{
			Namespace: cluster.Namespace,
			Name:      defaultAWSEBSCSIConfigMap.Name,
		},
		Data:       defaultAWSEBSCSIConfigMap.Data,
		BinaryData: defaultAWSEBSCSIConfigMap.BinaryData,
	}

	return namespacedEBSCSIConfigMap
```

pkg/handlers/generic/lifecycle/csi/handler.go
```diff
@@ -0,0 +1,172 @@
...
+	return &CSIHandler{
+		client:          c,
+		variableName:    clusterconfig.MetaVariableName,
+		variablePath:    []string{"addons", variableRootName},
+		ProviderHandler: handlers,
+	}
+}
+
+func (c *CSIHandler) Name() string {
+	return "CSIHANDLER"
```
dkoshkin, line 59: nit: the other handlers are camelCase
```suggestion
	return "CSIHandler"
```

api/v1alpha1/clusterconfig_types.go
```diff
@@ -312,3 +317,42 @@ func (NFD) VariableSchema() clusterv1.VariableSchema {
 		},
 	}
 }
+
+type CSIProviders struct {
+	// +optional
+	Providers []*CSIProvider `json:"providers,omitempty"`
```
dkoshkin, line 323: Let me know if Im missing the reason for a pointer please.
```suggestion
	Providers []CSIProvider `json:"providers,omitempty"`
```

pkg/handlers/generic/lifecycle/csi/handler.go
```diff
@@ -0,0 +1,172 @@
...
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+const (
+	variableRootName = "csiProviders"
+)
+
+type CSIProvider interface {
+	EnsureCSIConfigMapForCluster(context.Context, *clusterv1.Cluster) (*corev1.ConfigMap, error)
```
dkoshkin, line 31: Nice use of an interface here

api/v1alpha1/clusterconfig_types.go
```diff
@@ -15,7 +15,9 @@ import (
 )
 
 const (
-	CNIProviderCalico = "calico"
+	CNIProviderCalico                 = "calico"
+	CSIProviderAWSEBS                 = "aws-ebs-sc"
+	CSIProviderLocalVolumeProvisioner = "localvolumeprovisioner"
```
dkoshkin, line 20: Is `localvolumeprovisioner` supported? I don't see the specs for it?

api/v1alpha1/clusterconfig_types.go
```diff
@@ -321,6 +323,9 @@ type Addons struct {
 
 	// +optional
 	NFD *NFD `json:"nfd,omitempty"`
+
+	// +optional
+	CSIProviders CSIProviders `json:"csi,omitempty"`
```
dkoshkin, line 328: I see removed the pointer, but if its optional the convention is for it to be a pointer.
```suggestion
	CSIProviders *CSIProviders `json:"csi,omitempty"`
```

examples/capi-quick-start/docker-cluster.yaml
```diff
@@ -24,6 +24,7 @@ spec:
           addons:
             cni:
               provider: calico
+              value: {}
```
dkoshkin, line 27: Added by mistake? I don't see `value` defined in https://github.com/d2iq-labs/capi-runtime-extensions/blob/af46151552bcdb8a8c35166b819d0ddc1aee99c2/api/v1alpha1/clusterconfig_types.go#L340-L360

hack/examples/bases/docker/kustomization.yaml.tmpl
```diff
@@ -32,6 +32,13 @@ patches:
         - name: "clusterConfig"
           value:
             docker: {}
+    - op: "add"
+      path: "/spec/topology/variables/0/value/addons"
+      value:
+        nfd: {}
+        cni:
+          provider: calico
+          value: {}
```
dkoshkin, line 41:
```suggestion
        cni:
          provider: calico
```

api/v1alpha1/clusterconfig_types.go
```diff
@@ -15,7 +15,9 @@ import (
 )
 
 const (
-	CNIProviderCalico = "calico"
+	CNIProviderCalico                 = "calico"
+	CSIProviderAWSEBS                 = "aws-ebs-sc"
```
dkoshkin, line 19: What does "sc" mean here, maybe just this?
```suggestion
	CSIProviderAWSEBS                 = "aws-ebs"
```

api/v1alpha1/clusterconfig_types.go
```diff
@@ -370,6 +376,45 @@ func (NFD) VariableSchema() clusterv1.VariableSchema {
 	}
 }
 
+type CSIProviders struct {
+	// +optional
+	Providers []*CSIProvider `json:"providers,omitempty"`
+	// +optional
+	DefaultClassName string `json:"defualtClassName,omitempty"`
```
dkoshkin, line 383: I don't see how this get used, should we remove it for now and keep it simple and not set a default SC?

hack/addons/update-aws-ebs-csi.sh
```diff
@@ -0,0 +1,103 @@
...
+  sed 's|imagePullPolicy: Always|imagePullPolicy: IfNotPresent|g' \
+    >"${ASSETS_DIR}/${FILENAME}"
+
+# Add default storage class here.
+cat <<EOF >"${ASSETS_DIR}/aws-ebs-csi-temp.yaml"
+---
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: ebs-sc
```
dkoshkin, line 32: Should this match the csi provider name?
```suggestion
  name: aws-ebs
```

hack/examples/bases/aws/kustomization.yaml.tmpl
```diff
@@ -32,6 +32,15 @@ patches:
...
             aws: {}
+    - op: "add"
+      path: "/spec/topology/variables/0/value/addons"
+      value:
+        cni:
+          provider: calico
+        nfd: {}
+        csi:
+          providers:
+            - name: aws-ebs-sc
```
dkoshkin, line 43: If "sc" is for StorageClass lets drop it. Its deploying more than just the StorageClass.
```suggestion
            - name: aws-ebs
```

make/addons.mk
```diff
@@ -3,9 +3,11 @@
 
 export CALICO_VERSION := v3.26.1
 export NODE_FEATURE_DISCOVERY_VERSION := 0.14.1
+export AWS_CSI_SNAPSHOT_CONTROLLER_VERSION := v6.3.0
+export AWS_EBS_CSI_VERSION := release-1.23
 
 .PHONY: addons.sync
-addons.sync: $(addprefix update-addon.,calico nfd)
+addons.sync: $(addprefix update-addon.,calico nfd aws-csi)
```
dkoshkin, line 10: This needs to match `update-addon.aws-ebs-csi` right?
```suggestion
addons.sync: $(addprefix update-addon.,calico nfd aws-ebs-csi)
```

pkg/handlers/generic/lifecycle/csi/aws-ebs/handler.go
```diff
@@ -0,0 +1,103 @@
...
+	flags.StringVar(
+		&a.defaultsNamespace,
+		prefix+".defaultsNamespace",
+		corev1.NamespaceDefault,
+		"namespace of the ConfigMap used to deploy AWS EBS CSI driver",
+	)
+	flags.StringVar(
+		&a.defaultAWSEBSConfigMapName,
+		prefix+".aws-ebs-provider-configmap-name",
+		"aws-ebs-csi-crs-cm",
```
dkoshkin, line 33: More consistent with NFD and Calico where we don't include "cm" or "crs"
```suggestion
		"aws-ebs-csi",
```

hack/addons/update-aws-ebs-csi.sh
```diff
@@ -0,0 +1,103 @@
...
+EOF
+
+readonly KUSTOMIZE_BASE_DIR="${SCRIPT_DIR}/kustomize/aws-ebs/"
+mv "${ASSETS_DIR}/${FILENAME}" "${KUSTOMIZE_BASE_DIR}"/"${FILENAME}"
+kustomize --load-restrictor=LoadRestrictionsNone build --reorder=none "${KUSTOMIZE_BASE_DIR}" >>"${ASSETS_DIR}/${FILENAME}"
+rm "${KUSTOMIZE_BASE_DIR}"/"${FILENAME}"
+
+readonly OUTPUT_FILE=aws-ebs-configmap
+
+kubectl create configmap aws-ebs-csi-crs-cm --dry-run=client --output yaml \
```
dkoshkin, line 91:
```suggestion
kubectl create configmap aws-ebs-csi --dry-run=client --output yaml \
```

dkoshkin (commented):
Good job! Just minor comments

I'll take a look at the update script since it may be causing CI to fail.

make/addons.mk
dkoshkin: Oh I see either line 20 or 21 needs to change to match. (I would prefer `aws-ebs-csi` that matches the rest of the code)

api/v1alpha1/clusterconfig_types.go
faiq: storage class

api/v1alpha1/clusterconfig_types.go
faiq: We should leave it. I plan on adding this functionality as a part of this PR.

faiq: implemented.

pkg/handlers/generic/lifecycle/csi/handler.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+	return nil
+}
+
+func setDefaultStorageClass(
+	ctx context.Context,
+	cm *corev1.ConfigMap,
+) error {
+	for k, contents := range cm.Data {
+		if strings.Contains(k, ".yaml") {
```
dkoshkin, line 203: I wonder if we should iterate through all the keys until we find a `kindStorageClass`, relying on just the `.yaml` may not be enpugh.
And then return an error if we don't find one to change.

faiq: thats not a bad idea. it would be a lot safer to do it that way and allow us some flexibility as well.

faiq: implemented

dkoshkin (approved):
Looks good!

PR #198, opened 2023-10-02
feat: Deploy default clusterclasses via helm

jimmidyson:
Thought this was nicer when deploying via helm and can be disabled via values if users have their own clusterclasses to deploy.


charts/capi-runtime-extensions/templates/clusterclass_aws.yaml
```diff
@@ -0,0 +1,6 @@
+# Copyright 2023 D2iQ, Inc. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+{{- if and .Values.deployDefaultClusterClasses (.Capabilities.APIVersions.Has "infrastructure.cluster.x-k8s.io/v1beta2/AWSClusterTemplate") }}
```
dkoshkin, line 4: Good thinking using `Capabilities`

dkoshkin (approved):

mhrabovcin (approved):

PR #204, opened 2023-10-02
fix: move provider fields under aws and docker

dkoshkin:
Moving the AWS fields under `aws: {}`  and Docker fields under `docker: {}`


dkoshkin:
How is it passing in my IDE but not in CI 🤔 
image

jimmidyson:
@dkoshkin Are you running the AWS meta mutator tests? This fails locally for me btw.

jimmidyson:
Pushed a fix to meta handler tests.

jimmidyson:
I was thinking what this would look like if we had a single `ClusterConfig` type with all providers present, but we could validate only one is provided. We would then need to construct the variable schema as necessary for the meta mutator. WDYT? Instead of having separate provider specific top level cluster config types.

jimmidyson:
@dkoshkin I implemented my last suggestion in a new commit - let me know what you think. We an always revert if you don't think it's worth doing it like that.

api/v1alpha1/docker_clusterconfig_types.go
```diff
@@ -4,56 +4,15 @@
...
 	GenericClusterConfig `json:",inline"`
 }
 
-// DockerClusterConfigSpec defines the desired state of DockerClusterConfig.
-type DockerClusterConfigSpec struct {
-	// +optional
-	Docker *DockerSpec `json:"docker,omitempty"`
-
-	//+optional
-	CustomImage *OCIImage `json:"customImage,omitempty"`
```
dkoshkin, line 30: Wait, I messed up here with the `CustomImage`.

jimmidyson: I tested this before I pushed so it's really interesting that this didn't break anything re usage via patches, etc :sweat_smile:

PR #210, opened 2023-10-05
fix: Correctly set external cloud provider for AWS

jimmidyson:
Extracting the external cloud provider fix from #192. I noticed some other issues (e.g. CNI ingress rules) and will push some follow up PRs too, but wanted to get this in separately so we can actually create AWS clusters (totally broken right now).


dkoshkin (approved):
Thanks @faiq for finding and fixing the bug and thanks @jimmidyson for extracting it here.

PR #211, opened 2023-10-05
feat: CAPA v2.2.4

jimmidyson:
This updates vendored APIs.


dkoshkin (commented):
Should we regenerate the examples too and pull in the cloud-provider change?

dkoshkin:
👍 Ahh yep I checked https://github.com/kubernetes-sigs/cluster-api-provider-aws/releases/download/v2.2.4/cluster-template.yaml which did have the change, but the ClusterClass file doesn't, thanks.

dkoshkin (approved):

PR #213, opened 2023-10-05
fix: Ensure CNI ingress rules are added to AWSCluster

jimmidyson:
This fixes 2 problems:

- handler was not registered in AWS meta patch handler
- handler was not updated to use `clusterConfig.addons.cni` when CNI config was moved under addons property


jimmidyson:
Found this isn't idempotent so fixing now and adding tests - please do not merge.

jimmidyson:
@dkoshkin This is ready to go now - idempotency ftw :sweat_smile:

PR #217, opened 2023-10-06
feat: add AWS instance type patch

dkoshkin:
Depends on https://github.com/d2iq-labs/capi-runtime-extensions/pull/216

Tested manually:
```
- apiVersion: infrastructure.cluster.x-k8s.io/v1beta2
  kind: AWSMachineTemplate
  metadata:
    name: aws-quick-start-control-plane-dlj45
    namespace: default
    template:
      spec:
        iamInstanceProfile: control-plane.cluster-api-provider-aws.sigs.k8s.io
        instanceType: m5.2xlarge
- apiVersion: infrastructure.cluster.x-k8s.io/v1beta2
  kind: AWSMachineTemplate
metadata
    name: aws-quick-start-md-0-infra-kpc8m
    namespace: default
  spec:
    template:
      spec:
        iamInstanceProfile: nodes.cluster-api-provider-aws.sigs.k8s.io
        instanceType: m5.4xlarge
```


pkg/handlers/aws/mutation/instancetype/inject_control_plane.go
```diff
@@ -0,0 +1,133 @@
...
+	return newAWSInstanceTypeControlPlanePatchHandler(
+		clusterconfig.MetaVariableName,
+		clusterconfig.MetaControlPlaneConfigName,
+		awsclusterconfig.AWSVariableName,
+		VariableName,
+	)
+}
+
+func newAWSInstanceTypeControlPlanePatchHandler(
+	variableName string,
```
supershal, line 59:
```suggestion
	metaVariableName string,
```
explicitly using `meta` in the name helps understanding the code better.

pkg/handlers/aws/mutation/instancetype/inject_control_plane.go
```diff
@@ -0,0 +1,133 @@
...
+			).Info("setting instance type in control plane AWSMachineTemplate spec")
+
+			obj.Spec.Template.Spec.InstanceType = string(instanceTypeVar)
+
+			return nil
+		},
+	)
+}
+
+func (h *awsInstanceTypeControlPlanePatchHandler) GeneratePatches(
```
supershal, line 127: this method is only useful helper in unit tests since the meta handler will directly call the `Mutate` func of its registered handlers. It does not hurt to keep it here though.

supershal (approved):
other than some nits, the changes looks good.

pkg/handlers/aws/mutation/instancetype/inject_control_plane.go
dkoshkin: We use the "meta" prefix to indicate that its like `clusterConfig` or `workerConfig`. Lets instead refactor all of it and drop this concept of "meta" handlers from all handlers.

pkg/handlers/aws/mutation/instancetype/inject_control_plane.go
dkoshkin: It also needed to satisfy this interface https://github.com/d2iq-labs/capi-runtime-extensions/blob/8fad4c739699d216f4ef6e62d0ebf014ccb39516/common/pkg/capi/clustertopology/handlers/mutation/interfaces.go#L19-L25

pkg/handlers/aws/mutation/instancetype/inject_control_plane.go
supershal: thats sounds good as well. Lets pair on it. I also want to propose if we should consider replacing `meta`, `mutate`, `inject` keywords

pkg/handlers/aws/mutation/instancetype/inject_control_plane.go
supershal: I think `GeneratePatches` interface does not need to satisfy at this level. it just has to be satisfied at the meta handler level.

PR #224, opened 2023-10-11
refactor: Use common handler code

dkoshkin:
Refactor common handler code. This is not meant to completely support all handlers (i.e. `auditpolicy` with no vars, `httpproxy` that fetches additional information from the Cluster)

Its a little tricky because we rely on generics, so I would like to get some feedback before making change to the rest of the handlers.
Honestly I don' think this really solves much, but would love some feedback anyways.
Im going to explore additional refactoring to make the existing handler code less verbose.


dkoshkin:
Unnecessary with something like https://github.com/d2iq-labs/capi-runtime-extensions/pull/226

PR #226, opened 2023-10-12
refactor: Remove usage of non-meta handlers

dkoshkin:
Remove code for mutation handlers that was used to support individual handlers. As we are only registering the meta handlers, this makes it simpler to add new handlers and only a single place to add variable unit tests.

I also removed this assertion since the meta registration code will already fail if `Mutate` is not implemented and  `mutation.GeneratePatches` and `commonhandlers.Named` interfaces are no needed if registering in the meta handler.
```
	_ commonhandlers.Named     = &...PatchHandler{}
	_ mutation.GeneratePatches = &...PatchHandler{}
	_ mutation.MetaMutator     = &...PatchHandler{}
```


pkg/handlers/generic/clusterconfig/variables_test.go
```diff
@@ -19,8 +20,31 @@ func TestVariableValidation(t *testing.T) {
 		ptr.To(v1alpha1.GenericClusterConfig{}.VariableSchema()),
 		false,
 		NewVariable,
+		// CNI
 		capitest.VariableTestDef{
-			Name: "valid proxy config only",
+			Name: "CNI: set with valid provider",
```
jimmidyson, line 25: Can we split these variables tests into separate test files per variable property? Otherwise I think this is going to get difficult to maintain and ensure we have coverage of variable definitions.

pkg/handlers/docker/mutation/customimage/inject_worker_test.go
```diff
@@ -1,21 +0,0 @@
...
-package customimage
-
-import (
-	"testing"
-
-	"github.com/d2iq-labs/capi-runtime-extensions/common/pkg/capi/clustertopology/handlers/mutation"
-	"github.com/d2iq-labs/capi-runtime-extensions/pkg/handlers/docker/mutation/customimage/tests"
-)
-
-func TestWorkerGeneratePatches(t *testing.T) {
```
supershal, line 13: how are the patches can be unit tested after this refactor?

pkg/handlers/aws/mutation/cni/calico/inject.go
```diff
@@ -38,21 +33,11 @@ type calicoPatchHandler struct {
 	variableFieldPath []string
 }
 
-var (
-	_ commonhandlers.Named     = &calicoPatchHandler{}
-	_ mutation.GeneratePatches = &calicoPatchHandler{}
-	_ mutation.MetaMutator     = &calicoPatchHandler{}
```
supershal, line 44: we still want to keep validating `MetaMutator` interface.  `_ mutation.MetaMutator     = &calicoPatchHandler{}`

pkg/handlers/docker/mutation/customimage/inject_worker_test.go
jimmidyson: Only via the meta mutator tests - but keeping the test files in the `tests` package under this handler.

pkg/handlers/aws/mutation/cni/calico/inject.go
jimmidyson: I think should be covered by the `New...` func shouldn't it?

dkoshkin: Right this is not really needed, the code here fails to compile if `Mutate` is not implemented: https://github.com/d2iq-labs/capi-runtime-extensions/blob/8d4207bc721c275e233e914b4e21f84f35fc1d38/pkg/handlers/aws/mutation/metapatch_handler.go#L20-L36

supershal (approved):
Thank you for the clean up

PR #229, opened 2023-10-16
feat: deploy AWS CPI

### LINKED ISSUE #215: Deploy CPI for AWS
- build process
- lifecycle handler
- api

thinking something like this 
```
addons:
  cpi:
   providers:
   - name: aws-cpi
``` 

controller should check if provider is correct for infra type

faiq:
resolves #215 fo kubernetes 1.27

i would recommend creating a cluster using the aws cluster in the `examples` directory and checking the existence of the aws-cloud-controller-manager daemonset in the kube-system namespace


hack/addons/update-aws-cpi.sh
```diff
@@ -0,0 +1,50 @@
...
+trap_add "rm -rf ${ASSETS_DIR}" EXIT
+helm repo add aws-cloud-controller-manager https://kubernetes.github.io/cloud-provider-aws
+
+CHART_VERSION=""
+if [ "${AWS_CPI_VERSION}" = "1.27.1" ]; then
+  CHART_VERSION="0.0.8"
+fi
+
+export AWS_CPI_FROM_HELM_FILE="${ASSETS_DIR}/aws-cpi-from-helm-${AWS_CPI_VERSION}.yaml"
+helm template aws-cloud-controller-manager aws-cloud-controller-manager/aws-cloud-controller-manager --version="${CHART_VERSION}" >> "${AWS_CPI_FROM_HELM_FILE}"
```
dkoshkin, line 29: Can we do this with `kustomize` directly like we [already do for NFD](https://github.com/d2iq-labs/capi-runtime-extensions/blob/349bf97cf3aba491f782db694cfbdbd1a4802048/hack/addons/kustomize/nfd/kustomization.yaml.tmpl#L13-L22)?

make/addons.mk
```diff
@@ -5,6 +5,8 @@ export CALICO_VERSION := v3.26.3
 export NODE_FEATURE_DISCOVERY_VERSION := 0.14.1
 export AWS_CSI_SNAPSHOT_CONTROLLER_VERSION := v6.3.0
 export AWS_EBS_CSI_VERSION := release-1.23
+export AWS_CPI_VERSION_127 := v1.27.1
+export AWS_CPI_VERSION_126 := v1.26.1
```
dkoshkin, line 9: Maybe we can have a list of versions here and the make target updates it all?

pkg/handlers/generic/lifecycle/cpi/aws/handler.go
```diff
@@ -0,0 +1,180 @@
...
+					cpiDaemonSet,
+				)
+				if err != nil {
+					log.Error(err, "failed to convert unstructured to DaemonSet")
+					return nil, fmt.Errorf("failed to convert unstructured to DaemonSet %w", err)
+				}
+				cpiDaemonSet.Spec.Template.Spec.Containers[0].Args = append(
+					cpiDaemonSet.Spec.Template.Spec.Containers[0].Args,
+					fmt.Sprintf("--cluster-name=%s", cluster.Name),
+				)
```
dkoshkin, line 148: I'm not seeing `--cluster-name` in the example [specs](https://github.com/kubernetes/cloud-provider-aws/blob/a76e0ad4e9e0ff6bc8b09991ff97bf9e2cc10a77/examples/existing-cluster/base/aws-cloud-controller-manager-daemonset.yaml#L32-L38) or the helm chart.

Can you please try without it, not needing this would really help removing all this code here.

pkg/handlers/generic/lifecycle/utils/utils.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/d2iq-labs/capi-runtime-extensions/common/pkg/k8s/client"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	crsv1 "sigs.k8s.io/cluster-api/exp/addons/api/v1beta1"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+func EnsureCSICRSForClusterFromConfigMap(
```
dkoshkin, line 19: Since this is more generic now.
```suggestion
func EnsureCRSForClusterFromConfigMap(
```

make/addons.mk
dkoshkin, line 9: This var is not being used.

dkoshkin (commented):
Looks great! Didn't test it myself but please test with and without `--cluster-name` in the DS and paste your test files and what you are seeing in the cluster.

pkg/handlers/generic/lifecycle/cpi/aws/handler.go
faiq: This was taken from our konvoy2 code https://github.com/mesosphere/konvoy2/blob/main/pkg/capi/generator/templates/cpi/aws-cpi-manifests.yaml#L159

make/addons.mk
faiq: yeah, i can't seem to find the right helm release for it :(

pkg/handlers/generic/lifecycle/cpi/aws/handler.go
faiq: removed and it seems to work.

dkoshkin (approved):

PR #281, opened 2024-01-09
build: Use ko for building OCI image

jimmidyson:
Simplifies a lot of the Docker build config to use ko.


hack/flakes/flake.nix
```diff
@@ -2,7 +2,7 @@
   description = "Useful flakes for golang and Kubernetes projects";
 
   inputs = {
-    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+    nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
```
faiq, line 5: What is this?

.pre-commit-config.yaml
```diff
@@ -73,10 +73,6 @@ repos:
   hooks:
   - id: actionlint-system
     stages: [commit]
-- repo: https://github.com/hadolint/hadolint
-  rev: v2.12.1-beta
-  hooks:
-  - id: hadolint
```
faiq, line 79: why are we removing this?

jimmidyson: Because we no longer have a Dockerfile.

hack/flakes/flake.nix
jimmidyson: This is just changing to a different nix channel, using nixpkgs rather than nixos (which is more used for the OS). It doesn't make any difference, just feels a bit more correct.

.pre-commit-config.yaml
faiq: love to hear it!

faiq (approved):

PR #283, opened 2024-01-09
feat: Use CAAPH to deploy Calico on workload clusters

jimmidyson:
Tested locally with Docker provider and it works!


make/clusterctl.mk
```diff
@@ -11,6 +11,7 @@ clusterctl.init:
 	    clusterctl init \
 	      --kubeconfig=$(KIND_KUBECONFIG) \
 	      --infrastructure docker,aws \
+				--addon helm \
```
dkoshkin, line 14:
```suggestion
	      --addon helm \
```

charts/capi-runtime-extensions/README.md
```diff
@@ -31,14 +31,10 @@ A Helm chart for capi-runtime-extensions
 | deployDefaultClusterClasses | bool | `true` |  |
 | deployment.replicas | int | `1` |  |
 | env | object | `{}` |  |
-| hooks.CalicoCNI.defaultInstallationConfigMaps.AWSCluster.configMap.content | string | `""` |  |
```
dkoshkin, line 34: What are your thoughts on supporting both methods? I'm thinking how we could support this in DKP without being forced to do CRS to CAAPH migration for existing providers all at once.

jimmidyson: It's opt in to use the CNI add-on management via this project. Clusters that want to use a different means could just not opt in via the variable and manage their own crs at they are doing now perhaps? I'd rather not support multiple ways to deploy CNI in this project if we can avoid it.

faiq: I think we should support both ways to deploy addons. It would make this project a lot more extensible. Ideally we would be able to support clusters that use CRS and Helm Releases

jimmidyson: I'm not sure it makes sense for a user to be able to select how they want to deploy CNI, via ClusterResourceSet or via CAAPH. The only way I can see that being possible is if the variable definition changes to be able to specify an addon deployment strategy and that feels like it complicates the simple API we're creating via the existing variable definitions.

jimmidyson: Also the way we had to manipulate the calico installation resource when using CRS was definitely suboptimal at best, fragile at worst.

dkoshkin: The reason I brought this up is because I _think_ we already have a usecase for both in DKP, but would still like to leverage this project for addon installation.

Currently the plan is to keep existing infra providers using CRS but for any new providers to use CAAPH.

jimmidyson: Ok I'll have think how we can support that nicely from a variable perspective. This would be the same across all addons I presume and so having a consistent UX for specifying the addon strategy would be good.

dkoshkin: Right and it can't be a global controller setting since different clusters will want to use different strategies.
But maybe having a cluster wide option is enough?
Something like:

```
spec:
  topology:
    variables:
      - name: clusterConfig
        value:
          addons:
            strategy: <CAAPH/CRS> (and default to CAAPH)
```

jimmidyson: Yeah could do that. That just means that every addon has to support both strategies though.

jimmidyson: Also I didn't know whether to default or not, I didn't for now.

dkoshkin: >Yeah could do that. That just means that every addon has to support both strategies though.
Oh good point, we probably don't want that.

faiq (approved):

PR #288, opened 2024-01-11
feat: Add ClusterResourceSet strategy for CNI installation

jimmidyson:
Allows for further strategies to be added later, e.g. CAAPH and selected per-addon.


api/v1alpha1/addon_types.go
```diff
@@ -0,0 +1,139 @@
...
+
+package v1alpha1
+
+import (
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+
+	"github.com/d2iq-labs/capi-runtime-extensions/common/pkg/capi/clustertopology/variables"
+)
+
+type Addons struct {
```
dkoshkin, line 12: Thanks for moving these out

api/v1alpha1/addon_types.go
```diff
@@ -0,0 +1,139 @@
...
+const (
+	AddonStrategyClusterResourceSet AddonStrategy = "ClusterResourceSet"
+)
+
+// CNI required for providing CNI configuration.
+type CNI struct {
+	// +optional
+	Provider string `json:"provider,omitempty"`
+	// +optional
+	Strategy AddonStrategy `json:"strategy,omitempty"`
```
dkoshkin, line 52: Remind me again please how does `optional` and `omitempty` work here with `Required` on line 75. Should these not be marked optional here?

charts/capi-runtime-extensions/templates/cni/calico/manifests/tigera-operator-configmap.yaml
```diff
@@ -12,4 +12,5 @@ data:
 kind: ConfigMap
 metadata:
   creationTimestamp: null
-  name: '{{ .Values.hooks.CalicoCNI.defaultTigeraOperatorConfigMap.name }}'
+  name: '{{ .Values.hooks.CalicoCNI.crsStrategy.defaultTigeraOperatorConfigMap.name
+    }}'
```
dkoshkin, line 16: I think changed by accident?
```suggestion
  name: '{{ .Values.hooks.CalicoCNI.crsStrategy.defaultTigeraOperatorConfigMap.name}}'
```

charts/capi-runtime-extensions/templates/deployment.yaml
```diff
@@ -29,7 +29,7 @@ spec:
         imagePullPolicy: "{{ .Values.image.pullPolicy }}"
         args:
         - --webhook-cert-dir=/runtimehooks-certs/
-        - --calicocni.defaultsNamespace={{ .Release.Namespace }}
+        - --calicocni.crs.defaultsNamespace={{ .Release.Namespace }}
```
dkoshkin, line 32: So in the future we could have something like this right?
```
- --calicocni.crs.defaultsNamespace={{ .Release.Namespace }}
- --calicocni.caaph.defaultsNamespace={{ .Release.Namespace }}
```

dkoshkin (approved):
Thanks! Just had a few questions.

charts/capi-runtime-extensions/templates/deployment.yaml
jimmidyson: Exactly that. Sound OK?

api/v1alpha1/addon_types.go
jimmidyson: Optional and omitempty in these API definitions are standard when a default can be added via mutating webhook - standard k8s API behaviour. While that might seem odd that the OpenAPI definition says it is required, that is OK because the OpenAPI has a default declared. I agree it is a bit confusing, I think this is actually OK. If the OpenAPI definition was not set as required then `strategy: null` would be acceptable and not defaulted (at least I think that's the expected behaviour!).

charts/capi-runtime-extensions/templates/cni/calico/manifests/tigera-operator-configmap.yaml
jimmidyson: This is due to addons sync using kustomize and kustomize wrapping the value as it is now longer. This still works thanks to YAML semantics, even if it doesn't look very nice.

dkoshkin: Ahh got it 👍

charts/capi-runtime-extensions/templates/deployment.yaml
dkoshkin: Yep, was just confirming!

PR #292, opened 2024-01-17
feat: containerd configuration for mirror registry

supershal:
Fixes Issue: https://github.com/d2iq-labs/capi-runtime-extensions/issues/203

- Create containerd mirror configuration for a registry
- Update Dynamic Credentials Provider configuration to use mirror credentials
- Unit tests to check the generated kubeadm files
- Updated documentation for registry variable
- Changed schema of the imageRegistry variable to better represent registeries, credentials and mirror information

```
variables:
      - name: clusterConfig
        value:
          imageRegistries:
            - url: https://my-registry.io
              credentials:
                secretRef:
                  name: my-registry-credentials
              mirror:
                secretRef:
                  name: my-mirror-ca-cert-secret
```
Work In Progress:
 Test mirror configuration with ECR registry

PR #293, opened 2024-01-18
feat: Use latest dynamic credential provider and v1 kubelet API

jimmidyson:
Required to fully support latest kubelet credential providers prior to merging #292.


supershal (approved):
Thank you.

PR #296, opened 2024-01-19
test: fix race condition between registry credentials and mirror tests

supershal:
Stacked on https://github.com/d2iq-labs/capi-runtime-extensions/pull/292
Fixes flakes in unit tests for image registries: https://github.com/d2iq-labs/capi-runtime-extensions/actions/runs/7564146208/job/20597802432#step:5:68
```
 Message: failed to apply mutation: failed to apply Image Registry Credentials Secret: server-side apply failed: Operation cannot be fulfilled on secrets "test-kubeadmconfigtemplate-registry-config": object was modified
```
When a static credentials are provided for registry, the patch generator creates secret that contains static credentials provider configuration with `registry-config` suffix. This fix is created with server side apply. however Server side apply does not work with the fake client, hack around it by pre-creating empty Secrets. https://github.com/kubernetes-sigs/controller-runtime/issues/2341

When we added more test cases for mirror registry, two parallel running tests were changing the same secrets that resulted in above error intermittently.

Fix:
- ensure that the test secret name for static credentials provider is unique to avoid updating the same secret in parallel.

### DIFF
--- a/common/pkg/testutils/capitest/request/items.go
+++ b/common/pkg/testutils/capitest/request/items.go
@@ -21,8 +21,8 @@ import (
 
 const (
 	ClusterName                                  = "test-cluster"
-	KubeadmConfigTemplateRequestObjectName       = "test-kubeadmconfigtemplate"
-	KubeadmControlPlaneTemplateRequestObjectName = "test-kubeadmcontrolplanetemplate"
+	kubeadmConfigTemplateRequestObjectName       = "test-kubeadmconfigtemplate"
+	kubeadmControlPlaneTemplateRequestObjectName = "test-kubeadmcontrolplanetemplate"
 	Namespace                                    = corev1.NamespaceDefault
 )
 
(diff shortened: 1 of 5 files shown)


common/pkg/testutils/capitest/request/items.go
```diff
@@ -45,15 +45,24 @@ func NewRequestItem(
...
 }
 
-func NewKubeadmConfigTemplateRequestItem(uid types.UID) runtimehooksv1.GeneratePatchesRequestItem {
+func NewKubeadmConfigTemplateRequestItem(
+	uid types.UID,
+) runtimehooksv1.GeneratePatchesRequestItem {
+	return NewKubeadmConfigTemplateRequest(uid, kubeadmConfigTemplateRequestObjectName)
+}
+
+func NewKubeadmConfigTemplateRequest(
```
dkoshkin, line 54: How about something like this? I don't think dropping the "Item" is clear what the difference is
```suggestion
func NewNamedKubeadmConfigTemplateRequestItem(
```

common/pkg/testutils/capitest/request/items.go
```diff
@@ -75,8 +84,9 @@ func NewKubeadmConfigTemplateRequestItem(uid types.UID) runtimehooksv1.GenerateP
 	)
 }
 
-func NewKubeadmControlPlaneTemplateRequestItem(
+func NewKubeadmControlPlaneTemplateRequest(
```
dkoshkin, line 87: Also here

pkg/handlers/generic/mutation/imageregistries/credentials/tests/generate_patches.go
```diff
@@ -5,6 +5,7 @@ package tests
```
dkoshkin, line 5: Maybe we should rename this file to `generate_credentials_patches.go` to match the new `generate_mirror_patches.go`?

dkoshkin (approved):
Looks good, just a couple of nits

supershal:
It was difficult to refactor #292 without this changes . cherry-picked commits from this PR to the parent PR. https://github.com/d2iq-labs/capi-runtime-extensions/pull/292
Closing this PR.

PR #333, opened 2024-02-06
feat: Add Cilium CNI addon support

jimmidyson:


charts/capi-runtime-extensions/values.yaml
```diff
@@ -2,30 +2,39 @@
...
+      defaultPodSubnet: 192.168.0.0/16
+      crsStrategy:
+        defaultTigeraOperatorConfigMap:
+          name: tigera-operator
+        defaultInstallationConfigMaps:
+          DockerCluster:
+            create: true
+            configMap:
+              name: calico-cni-crs-installation-dockercluster
+              content: ""
```
supershal, line 16: Could not find references to `content` used to create helm installation configmaps.

jimmidyson: Used in https://github.com/d2iq-labs/capi-runtime-extensions/blob/39ceadb8b958bcca0efeed7c983c675d330184cd/charts/capi-runtime-extensions/templates/cni/calico/manifests/docker/crs-installation.yaml#L12

dkoshkin (approved):
Thanks! Pushed a commit to fix the README.
Also tested both Calico and Cillium with the Docker provider and they look good.

PR #331, opened 2024-02-05
feat: introduce a Go module for /api

dkoshkin:
The intent is to make it easier to import the API in other projects. Split the PR into a few commits to make it easier to review.

The changes are just moving files around, no actual code changes.

PR #335, opened 2024-02-07
ci: Tag api module on release

jimmidyson:
This enables it to be imported directly into other Go projects.

Follow up to #331.


dkoshkin (approved):
Thank you! I saw some logic around finding go.mod and thought that would have been enough.

jimmidyson:
@dkoshkin yeah we don't use the `make tag` for this unfortunately - maybe we should but this is a bit tricky when using release-please.

PR #343, opened 2024-02-08
feat: Support multiple registry credentials if specified

jimmidyson:


pkg/handlers/generic/mutation/mirrors/mirror.go
```diff
@@ -8,7 +8,7 @@ import (
 	"context"
 	_ "embed"
 	"fmt"
-	"text/template"
+	"html/template"
```
supershal, line 11: Thanks

pkg/handlers/generic/mutation/imageregistries/credentials/tests/generate_patches.go
```diff
@@ -48,30 +43,6 @@ func TestGeneratePatches(
 		),
 	)
 
-	// Server side apply does not work with the fake client, hack around it by pre-creating empty Secrets
-	// https://github.com/kubernetes-sigs/controller-runtime/issues/2341
-	require.NoError(
```
supershal, line 53: great way to fix this.

supershal (approved):

pkg/handlers/generic/mutation/mirrors/mirror.go
jimmidyson: Oops! Thanks for spotting this!

PR #344, opened 2024-02-09
fix: use correct AWS EBS CSI images

dkoshkin:
The manifests in the upstream CSI repos may not have the correct images set.
Override the image to match the expected version.

Because of the mismatch between the image and the specs, the EBS CSI deployment failed with:
```
kubectl logs -n kube-system ebs-csi-controller-5564c47bd7-wphl2 
Defaulted container "ebs-plugin" out of: ebs-plugin, csi-provisioner, csi-attacher, csi-snapshotter, csi-resizer, liveness-probe
unknown flag: --batching
```


supershal (approved):
Good catch.

jimmidyson (approved):
I would prefer to see this switched to using the Helm chart install for EBS CSI as this solves a number of the image and patch issues, but this is good for now. Thanks for the fix @dkoshkin!

PR #346, opened 2024-02-09
test: add a Makefile target to update image

dkoshkin:
Added a new Makefile target to just build and redeploy the webhook on an existing kind cluster.
This target is useful when testing bug fixes in downstream projects.

### DIFF
--- a/make/dev.mk
+++ b/make/dev.mk
@@ -17,3 +17,13 @@ endif
 		--wait --wait-for-jobs
 	kubectl rollout restart deployment capi-runtime-extensions
 	kubectl rollout status deployment capi-runtime-extensions
+
+dev.update-webhook-image-on-kind:
+ifndef SKIP_BUILD
+	$(MAKE) release-snapshot
+endif
+	kind load docker-image --name $(KIND_CLUSTER_NAME) \
+		ko.local/capi-runtime-extensions:$$(gojq -r .version dist/metadata.json)
+	kubectl set image deployment capi-runtime-extensions webhook=ko.local/capi-runtime-extensions:$$(gojq -r .version dist/metadata.json)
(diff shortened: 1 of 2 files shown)


jimmidyson:
How about we just update run on kind target to skip clusterctl init if providers are already deployed?

dkoshkin:
> How about we just update run on kind target to skip clusterctl init if providers are already deployed?

I was using this in DKP where its not using the Helm Chart directly to deploy it, so `helm upgrade` failed because of conflicting resources without the proper Helm labels.

make/dev.mk
```diff
@@ -17,3 +17,13 @@ endif
...
 	kubectl rollout restart deployment capi-runtime-extensions
 	kubectl rollout status deployment capi-runtime-extensions
+
+dev.update-webhook-image-on-kind:
+ifndef SKIP_BUILD
+	$(MAKE) release-snapshot
+endif
+	kind load docker-image --name $(KIND_CLUSTER_NAME) \
+		ko.local/capi-runtime-extensions:$$(gojq -r .version dist/metadata.json)
+	kubectl set image deployment capi-runtime-extensions webhook=ko.local/capi-runtime-extensions:$$(gojq -r .version dist/metadata.json)
```
supershal, line 27: I cant see any difference between this and 
```
helm upgrade --install capi-runtime-extensions ./charts/capi-runtime-extensions \
		--set-string image.repository=ko.local/capi-runtime-extensions \
		--set-string image.tag=$$(gojq -r .version dist/metadata.json) \
		--wait --wait-for-jobs
``` in the `dev.run-on-kind` target

dkoshkin: helm upgrade will install all other CR from the helm chart, this new target only updates the Deployment image.

jimmidyson (approved):

PR #363, opened 2024-02-15
fix: v2/ in Containerd mirror path

dkoshkin:
Found a bug(?) in how Containerd handles mirrors with a path and it incorrectly appends `/v2` to the end of mirror, and not between the provided host and path.

See https://github.com/containerd/containerd/issues/6634#issuecomment-1944899130 for a longer explanation and example output.


jimmidyson (approved):
LGTM. Thanks for the detailed explanation.

PR #364, opened 2024-02-15
fix: set config_path in Containerd config

dkoshkin:
Set `config_path ` in the Containerd config as described in https://github.com/containerd/containerd/blob/main/docs/hosts.md#cri

~This is not enough, we also need to remove `registry.mirrors` from the config file, opening in case @jimmidyson you have some ideas on how we can do this, [TOML doesn't have a nil value](https://github.com/toml-lang/toml/issues/30)~:
```
Feb 15 05:59:13 docker-quick-start-helm-addon-cilium-rjfjv-hn9m7 containerd[7127]: time="2024-02-15T05:59:13.663662461Z" level=warning msg="failed to load plugin io.containerd.grpc.v1.cri" error="invalid plugin config: `mirrors` cannot be set when `config_path` is provided"
```

No additional changes are needed, the unnecessary mirror configuration will be removed from the OS image instead.

### DIFF
--- a/pkg/handlers/generic/mutation/mirrors/inject.go
+++ b/pkg/handlers/generic/mutation/mirrors/inject.go
@@ -99,7 +99,7 @@ func (h *globalMirrorPatchHandler) Mutate(
 			if err != nil {
 				return err
 			}
-			files, generateErr := generateFiles(
+			files, commands, generateErr := generateFilesAndCommands(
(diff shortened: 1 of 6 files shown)


jimmidyson:
I'm confused why `registry.mirrors` is in the config file in the first place?

jimmidyson (approved):

PR #367, opened 2024-02-15
build: Include CAPX APis

jimmidyson:


pkg/handlers/generic/mutation/mirrors/mirror_test.go
```diff
@@ -42,8 +42,10 @@ func Test_generateDefaultRegistryMirrorFile(t *testing.T) {
 			wantErr: nil,
 		},
 		{
-			name:   "ECR image registry with a path and no CA certificate",
-			config: &mirrorConfig{URL: "https://123456789.dkr.ecr.us-east-1.amazonaws.com/myproject"},
+			name: "ECR image registry with a path and no CA certificate",
+			config: &mirrorConfig{
+				URL: "https://123456789.dkr.ecr.us-east-1.amazonaws.com/myproject",
+			},
```
dkoshkin, line 48: I wonder why its still showing up here if https://github.com/d2iq-labs/capi-runtime-extensions/pull/368/files is merged

dkoshkin (approved):
Thanks!

pkg/handlers/generic/mutation/mirrors/mirror_test.go
jimmidyson: Because I hadn't rebased...

PR #392, opened 2024-02-22
build: Use kustomization.yaml from ESB CSI repo

jimmidyson:
No change to output but a cleaner way to pull in resources using
upstream kustomization.yamls.


dkoshkin (approved):
Thanks

PR #393, opened 2024-02-23
feat: Upgrade AWS ESB CSI and switch to using Helm chart

jimmidyson:
This will make it easier to adopt the HelmAddon strategy for deploying
the AWS ESB CSI addon.

Depends on #392


charts/capi-runtime-extensions/templates/csi/aws-ebs/manifests/aws-ebs-csi-configmap.yaml
```diff
@@ -1045,7 +1045,12 @@ data:
     kind: ServiceAccount
     metadata:
       labels:
+        app.kubernetes.io/component: csi-driver
+        app.kubernetes.io/instance: aws-ebs-csi-driver
+        app.kubernetes.io/managed-by: Helm
```
dkoshkin, line 1050: I noticed these Helm labels in other addons, I guess its not a problem for these to be here if not actually managed by helm? I also wonder if there is a helm controller if it tries to "adopt" them.

Something to test later.

dkoshkin (approved):
Thanks!

charts/capi-runtime-extensions/templates/csi/aws-ebs/manifests/aws-ebs-csi-configmap.yaml
jimmidyson: Yes that should make it possible. But yes good to test and see what happens.

jimmidyson: This is the kind of thing we should be able to do via e2e tests pretty easily I hope 🤞

jimmidyson: Opened #394

PR #410, opened 2024-02-28
build: Update AWS CPI manifest filenames

jimmidyson:
Previous names contained `-ebs` which must have been a copy-paste error from
EBS addon. No functionality change, just file rename.

And hooked update into `addons.sync` target.


dkoshkin (approved):
Thanks!

PR #411, opened 2024-02-28
fix: Disbable hubble in Cilium deployment via CRS

jimmidyson:
The Job that is created for hubble auto-generate certs cannot be
updated as it is immutable.


dkoshkin (approved):
FYI Hubble is an observability platform https://github.com/cilium/hubble/

PR #422, opened 2024-03-05
docs: simplify running examples in README

dkoshkin:
I was trying to create AWS example clusters and it was pretty cumbersome.
Changing the README in an attempt to make it easier to to use all example files.


README.md
```diff
@@ -72,8 +72,13 @@ EOF
 To create an example cluster:
 
 ```shell
-clusterctl generate cluster docker-quick-start-helm-addon-cilium \
-  --from examples/capi-quick-start/docker-cluster-cilium-helm-addon.yaml \
+export CLUSTER_NAME=docker-cluster-cilium-helm-addon
+export CLUSTER_FILE=examples/capi-quick-start/docker-cluster-cilium-helm-addon.yaml
```
jimmidyson, line 76: Not sure we need to export this here, it's only used in one place.

examples/capi-quick-start/aws-cluster-calico-crs.yaml
```diff
@@ -53,3 +53,15 @@ spec:
...
+kind: AWSClusterStaticIdentity
+metadata:
+  labels:
+    cluster.x-k8s.io/provider: aws
+  name: ${CLUSTER_NAME}
+spec:
+  allowedNamespaces:
+    list:
+      - default
+  secretRef: aws-quick-start-creds
```
jimmidyson, line 67: I don't think these are being used - https://github.com/d2iq-labs/capi-runtime-extensions/blob/main/charts/capi-runtime-extensions/defaultclusterclasses/aws-cluster-class.yaml#L34-L38

jimmidyson (changes requested):

examples/capi-quick-start/aws-cluster-calico-crs.yaml
dkoshkin: Ahh yep you're right, removed all of that

README.md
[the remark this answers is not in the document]
dkoshkin: I haven't removed the env yet because I still think its easier to set it at the top to make the command more copy and pastable. 

(My IDE even gives me a button to run the commands)
image

jimmidyson (approved):

PR #423, opened 2024-03-05
feat: add cluster-autoscaler addon

dkoshkin:
Adding the [cluster-autoscaler](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler) addon, starting with ClusterResourceSet first.

The cluster-autoscaler addon is different than the other addons because it needs access to both the CAPI objects and the workload Cluster. Because of this, it needs to always be deployed on the management Cluster and use the CAPI kubeconfig Secret to watch the workload Cluster.

Tested locally, but was not able to work on e2e tests yet, since there is no mechanism yet to make the Cluster self-managed.


dkoshkin:
Would love some foeedback on this before going much further 🙏

docs/content/addons/cluster-autoscaler.md
```diff
@@ -0,0 +1,40 @@
...
+              strategy: ClusterResourceSet
+    workers:
+      machineDeployments:
+        - class: default-worker
+          metadata:
+            annotations:
+              # Set the following annotations to configure the Cluster Autoscaler
+              # The initial MachineDeployment will have 1 Machine
+              cluster.x-k8s.io/cluster-api-autoscaler-node-group-max-size: "3"
+              cluster.x-k8s.io/cluster-api-autoscaler-node-group-min-size: "1"
```
jimmidyson, line 37: I feel like these should be set via the CRE variables which will patch with these annotations. WDYT? That would maek usage of CA easier, not having to know these specific annotations.

docs/content/addons/cluster-autoscaler.md
```diff
@@ -0,0 +1,40 @@
...
+          name: md-0
+          # Remove the replicas field, otherwise the topology controller will revert back the aytoscaler's changes
```
jimmidyson, line 39: This can be done via patch can't it?

docs/content/addons/cluster-autoscaler.md
jimmidyson: That would mean adding properties on the `workerConfig` variable I think.

hack/addons/kustomize/cluster-autoscaler/helm-values.yaml
```diff
@@ -0,0 +1,39 @@
+# Copyright 2023 D2iQ, Inc. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+---
+# This is a hack, but because a single cluster-autoscaler deployment can only monitor a single Kubernetes cluster
+# we expect 'tmpl-clustername-tmpl' and 'tmpl-clusternamespace-tmpl'
+# to be replaced with the Cluster's name and namespace.
+fullnameOverride: cluster-autoscaler-tmpl-clustername-tmpl
```
jimmidyson, line 8: What about using go template syntax? And parsing the CRS configmap through go gemplating when creating the actual resources?
```suggestion
fullnameOverride: cluster-autoscaler-{{ .clusterName }}
```

make/addons.mk
```diff
@@ -13,7 +14,7 @@ export AWS_CPI_VERSION_128 := v1.28.1
 export AWS_CPI_CHART_VERSION_128 := 0.0.8
 
 .PHONY: addons.sync
-addons.sync: $(addprefix update-addon.,calico cilium nfd aws-ebs-csi aws-cpi.127 aws-cpi.128)
+addons.sync: $(addprefix update-addon.,calico cilium nfd update-addon.cluster-autoscaler aws-ebs-csi aws-cpi.127 aws-cpi.128)
```
jimmidyson, line 17:
```suggestion
addons.sync: $(addprefix update-addon.,calico cilium nfd cluster-autoscaler aws-ebs-csi aws-cpi.127 aws-cpi.128)
```

hack/addons/kustomize/cluster-autoscaler/helm-values.yaml
jimmidyson: Oh let's stick with what you've done here - it's safe enough.

docs/content/addons/cluster-autoscaler.md
dkoshkin: What would the variables contain? Maybe if the autoscaler addon is enabled we automatically patch the Cluster by removing the replicas and adding these annotations?

Does this make sense to be in a webhook instead?

jimmidyson: Variables would contain min and max, patch would remove the replicas and set the annotations.

dkoshkin: Did some more testing today, I may be missing something but I didn't see handlers getting called for `MachineDeployments`, do we need to configure something else here?

jimmidyson: 😞 Looks as though my idea was wrong, we can't patch `MachineDeployments`, only template types. So I guess we will have to go back to the `metadata.annotations` on cluster `spec.workers.machineDeployments[*]` and we can't patch the replicas here either. This means we will have to leave it up to the user and provide a clean UX to enable/disable in the UI/CLI.

jimmidyson:
More discussion on cluster-autoscaler and ClusterClass in https://github.com/kubernetes-sigs/cluster-api/issues/8217

docs/content/addons/cluster-autoscaler.md
jimmidyson: Thinking about this UX and reading https://github.com/kubernetes-sigs/cluster-api/blob/6fede8c8cbc83e108aa6a5bc4507ed3aae096105/api/v1beta1/machinedeployment_webhook.go#L278-L306, I propose that we only set the autoscaler annotations on the machinedeployments rather than the replicas field. By default this can be set to the number of replicas and we can enable autoscaling later by simply enabling the addon and changing the min/max annotations. WDYT?

docs/content/addons/cluster-autoscaler.md
jimmidyson: No it can't 😞

docs/content/addons/cluster-autoscaler.md
jimmidyson: Great minds! That's exactly what you were doing in [`0044549` (#423)](https://github.com/d2iq-labs/capi-runtime-extensions/pull/423/commits/00445497254449eaf2e279c624f9674a21cd3069#diff-e520ef54af4a462ca097950d967c884c9775bc52571f31e56ce18d6d87241a3fL35-L42)

dkoshkin:
> More discussion on cluster-autoscaler and ClusterClass in [kubernetes-sigs/cluster-api#8217](https://github.com/kubernetes-sigs/cluster-api/issues/8217)

Interesting! We solved this in DKP by having the CLI client update both the annotations AND the MachineDeployment's replicas with an imperative command.

But I think the `--enforce-node-group-min-size` [flag](https://github.com/kubernetes/autoscaler/pull/5195) is a better solution here.

Not a huge fan that the internal annotation details are exposed to the user, but maybe thats ok for the initial implementation.

docs/content/addons/cluster-autoscaler.md
dkoshkin: 🧠 Ok so let's basically revert the last commit and have users opt-in to cluster-autoscaler and update the replicas by manipulating the annotations in the Cluster?

jimmidyson: Yeah, annotations on the MachineDeployment

pkg/handlers/generic/lifecycle/clusterautoscaler/strategy_crs.go
```diff
@@ -0,0 +1,125 @@
...
+	defaultClusterAutoscalerConfigMap string
+}
+
+func (c *crsConfig) AddFlags(prefix string, flags *pflag.FlagSet) {
+	flags.StringVar(
+		&c.defaultsNamespace,
+		prefix+".defaults-namespace",
+		corev1.NamespaceDefault,
+		"namespace of the ConfigMap used to deploy cluster-autoscaler",
+	)
```
jimmidyson, line 34: This will need wiring up in the helm chart deployment.

pkg/handlers/generic/lifecycle/clusterautoscaler/strategy_crs.go
```diff
@@ -0,0 +1,125 @@
...
+	// The cluster-autoscaler is different from other addons.
+	// It requires all resources to be created in the management cluster,
+	// which means creating the ClusterResourceSet always targeting the management cluster.
+	// In most cases, target the management cluster.
+	// But if existingManagementCluster is nil, i.e. when s.client points to a bootstrap cluster,
+	// target the cluster and assume that will become the management cluster.
+	targetCluster := existingManagementCluster
+	if targetCluster == nil {
+		targetCluster = cluster
+	}
```
jimmidyson, line 115: Nice!

jimmidyson (changes requested):

jimmidyson (approved):
Nice work!

jimmidyson:
CRS is not applied from bootstrap to workload:

```
- lastTransitionTime: "2024-03-07T14:34:49Z"
    message: '[creating object /v1, Kind=ServiceAccount quick-start-zu4cbq/cluster-autoscaler-quick-start-la8a9u:
      namespaces "quick-start-zu4cbq" not found, creating object policy/v1, Kind=PodDisruptionBudget
      quick-start-zu4cbq/cluster-autoscaler-quick-start-la8a9u: namespaces "quick-start-zu4cbq"
      not found, creating object rbac.authorization.k8s.io/v1, Kind=Role quick-start-zu4cbq/cluster-autoscaler-quick-start-la8a9u:
      namespaces "quick-start-zu4cbq" not found, creating object rbac.authorization.k8s.io/v1,
      Kind=RoleBinding quick-start-zu4cbq/cluster-autoscaler-quick-start-la8a9u: namespaces
      "quick-start-zu4cbq" not found, creating object /v1, Kind=Service quick-start-zu4cbq/cluster-autoscaler-quick-start-la8a9u:
      namespaces "quick-start-zu4cbq" not found, creating object apps/v1, Kind=Deployment
      quick-start-zu4cbq/cluster-autoscaler-quick-start-la8a9u: namespaces "quick-start-zu4cbq"
      not found]'
```

jimmidyson:
Need to specify the namespace for CA and create all resources in there.

jimmidyson:
With this approach, the CA will always be deployed on the CP nodes because workers will be set to 0, CA will be deployed and only then will worker nodes be scaled up.

dkoshkin:
> Need to specify the namespace for CA and create all resources in there.

Yep, thank you e2e tests https://github.com/d2iq-labs/capi-runtime-extensions/pull/423/commits/e9c73428b035f0b6521213e02a87ef5f923aa727
When I tested locally I created the management cluster in the `default` namespace so missed that.

dkoshkin:
> With this approach, the CA will always be deployed on the CP nodes because workers will be set to 0, CA will be deployed and only then will worker nodes be scaled up.

The CAPI webhook will set the MD replicas based on the CA annotation https://github.com/kubernetes-sigs/cluster-api/blob/main/internal/webhooks/machinedeployment.go#L289-L291 

I also see this behavior running locally when setting `--worker-machine-count=3`
```
k get md
NAME                              CLUSTER                REPLICAS   READY   UPDATED   UNAVAILABLE   PHASE     AGE   VERSION
dkoshkin-ca-test-200-md-0-zzxt7   dkoshkin-ca-test-200   3          3       3         0             Running   19m   v1.28.7
``` 

We may want to support clusters with 0 replicas, but I think we should do that as a separate effort since other addons would need similar tolerations.

dkoshkin:
Oops, I didn't see the auto-merge was enabled but the e2e tests passed 🎉

pkg/handlers/generic/lifecycle/clusterautoscaler/strategy_crs.go
```diff
@@ -0,0 +1,142 @@
...
+	// In the case when existingManagementCluster is nil, i.e. when s.client points to a bootstrap cluster,
+	// it is possible that the namespace where the cluster will be moved to, and where the cluster-autoscaler resources
+	// will be created, does not exist yet.
+	// In that case, we need to create the namespace in the target cluster.
+	clusterKey := ctrlclient.ObjectKeyFromObject(targetCluster)
+	remoteClient, err := remote.NewClusterClient(ctx, "", s.client, clusterKey)
+	if err != nil {
+		return fmt.Errorf("error creating remote cluster client: %w", err)
+	}
+	if err = utils.EnsureNamespace(ctx, remoteClient, cluster.Namespace); err != nil {
```
jimmidyson, line 127: I think we can just embed the namespace resource in the CRS configmap as we do for other addons.

dkoshkin: The reason I didn't do that here is because for the other addons this namespace is created on a fresh workload cluster where the namespace wouldn't exist.
In this case, the namespace will be created on the management cluster where this namespace can be shared across multiple clusters or exiting namespaces that have labels and annotations.

jimmidyson: Good point!

PR #426, opened 2024-03-07
feat: Single defaults namespace flag

jimmidyson:
Simplifies configuration in all handlers.


charts/capi-runtime-extensions/templates/deployment.yaml
```diff
@@ -58,6 +51,11 @@ spec:
         - containerPort: 8081
           name: probes
           protocol: TCP
+        env:
+        - name: POD_NAMESPACE
```
dkoshkin, line 55: Good idea!

dkoshkin (approved):

PR #439, opened 2024-03-14
test(e2e): Add self-hosted e2e test

jimmidyson:
Creating as draft until kubernetes-sigs/cluster-api-addon-provider-helm#189 is merged and released.

Depends on #494.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -59,8 +59,9 @@ jobs:
       matrix:
         provider:
           - Docker
-          # Uncomment below once we have the ability to run e2e tests on AWS from GHA.
+          # Uncomment below once we have the ability to run e2e tests on other providers from GHA.
           # - AWS
+          # - Nutanix
       fail-fast: false
     uses: ./.github/workflows/e2e.yml
     with:
@@ -76,8 +77,9 @@ jobs:
       matrix:
         provider:
           - Docker
-          # Uncomment below once we have the ability to run e2e tests on AWS from GHA.
+          # Uncomment below once we have the ability to run e2e tests on other providers from GHA.
           # - AWS
+          # - Nutanix
       fail-fast: false
     uses: ./.github/workflows/e2e.yml
     with:
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -54,15 +54,35 @@ jobs:
         with:
           test-results: test.json
 
-  e2e-test:
+  e2e-quick-start:
     strategy:
       matrix:
         provider:
           - Docker
+          # Uncomment below once we have the ability to run e2e tests on AWS from GHA.
+          # - AWS
       fail-fast: false
     uses: ./.github/workflows/e2e.yml
     with:
       provider: ${{ matrix.provider }}
+      focus: Quick start
+    secrets: inherit
+    permissions:
+      contents: read
+      checks: write
+
+  e2e-self-hosted:
+    strategy:
+      matrix:
+        provider:
+          - Docker
+          # Uncomment below once we have the ability to run e2e tests on AWS from GHA.
+          # - AWS
+      fail-fast: false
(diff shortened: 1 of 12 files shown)


dkoshkin (approved):
Thanks!

test/e2e/clusterautoscaler_helpers.go
```diff
@@ -0,0 +1,102 @@
...
+		client.ObjectKey{
+			Namespace: input.WorkloadCluster.Namespace,
+			Name:      input.WorkloadCluster.Name + "-kubeconfig",
+		},
+		&corev1.Secret{},
+	)
+	if err != nil && errors.IsNotFound(err) {
+		return
+	}
+	Expect(err).NotTo(HaveOccurred())
```
dkoshkin, line 61: nit: we can reuse the same function that the CA addon uses
```suggestion
	// Only check for ClusterAutoscaler if the cluster is self-managed.
	// ManagementCluster function will return a nil managementCluster if workloadClusterClient
	// is not a self-managed cluster.
	workloadClusterClient := input.ClusterProxy.GetWorkloadCluster(
		ctx, input.WorkloadCluster.Namespace, input.WorkloadCluster.Name,
	).GetClient()
	managementCluster, err := utils.ManagementCluster(ctx, workloadClusterClient)
	Expect(err).NotTo(HaveOccurred())
	if  managementCluster == nil {
		return
	}
```

.github/workflows/checks.yml
```diff
@@ -54,15 +54,35 @@ jobs:
...
           test-results: test.json
 
-  e2e-test:
+  e2e-quick-start:
     strategy:
       matrix:
         provider:
           - Docker
+          # Uncomment below once we have the ability to run e2e tests on AWS from GHA.
+          # - AWS
```
deepakm-ntnx, line 63: or Nutanix. was wondering how are we planning to pass the setup details of either aws or nutanix

dkoshkin: GHA Secrets should be enough I think.

test/e2e/clusterautoscaler_helpers.go
jimmidyson: Thanks! Applied.

.github/workflows/checks.yml
jimmidyson: Yes these will need to be environment variables specfified either via GHA secrets when running in GHA or just normal env vars (via a local envrc perhaps) for local development. These should be looked up via the e2e variables though as a consistent way to interact with variables either from e2e config file or from the environment.

jimmidyson:
To merge this, will need to update required status checks (make `e2e-test (Docker)` non-required and make the `e2e-quick-start` and `e2e-self-hosted` checks required) via Canaveral, merge it, and then all other open PRs will need to rebased to include this self-managed e2e.

test/e2e/framework/self_hosted.go
```diff
@@ -0,0 +1,484 @@
...
+			framework.CreateNamespaceAndWatchEventsInput{
+				Creator:             selfHostedClusterProxy.GetClient(),
+				ClientSet:           selfHostedClusterProxy.GetClientSet(),
+				Name:                namespace.Name,
+				LogFolder:           filepath.Join(input.ArtifactFolder, "clusters", "bootstrap"),
+				IgnoreAlreadyExists: true,
+			},
+		)
+
+		By("Loading CAPI runtime extensions image to the workload cluster")
```
faiq, line 230: Is this a mistake? Where is this happening?

jimmidyson: Oops yeah, this is actually happening on lines 201

deepakm-ntnx (approved):

dkoshkin (approved):

mhrabovcin (approved):

PR #440, opened 2024-03-14
build: controller-runtime v0.17.2

### LINKED ISSUE #402: Switch to envtest from fake-client in unit tests
controller-runtime v0.17.2 explicitly disabled fake-client SSA calls, which renders our workaround useless. The proposed solution is to use envtest for unit tests that use SSA, which is most of our lifecycle handler tests.

Once this is done, dependabot.yml should be updated to remove ignore configuration for controller-runtime upgrades.

jimmidyson:
This removes the fakessa test helper and uses testenv to provide
a proper manager backed by real API server for testing.

Fixes #402.


make/go.mk
```diff
@@ -25,7 +25,6 @@ define go_test
 		-- \
 		-covermode=atomic \
 		-coverprofile=coverage.out \
-		-race \
```
dkoshkin, line 28: Why is this being removed?

pkg/handlers/aws/mutation/metapatch_handler_test.go
```diff
@@ -6,12 +6,10 @@ package mutation
 import (
 	"testing"
 
-	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 
 	"github.com/d2iq-labs/capi-runtime-extensions/api/v1alpha1"
 	"github.com/d2iq-labs/capi-runtime-extensions/common/pkg/capi/clustertopology/handlers/mutation"
-	"github.com/d2iq-labs/capi-runtime-extensions/common/pkg/k8s/client/fakessa"
```
dkoshkin, line 14: So its ok to still use `"sigs.k8s.io/controller-runtime/pkg/client/fake"` because those did not rely on SSA

dkoshkin (approved):

pkg/handlers/aws/mutation/metapatch_handler_test.go
jimmidyson: Yes.

make/go.mk
jimmidyson: Because there's a race in controller-runtime envtest... I noticed that CAPA, which uses this same approach, also doesn't enable race here.

PR #443, opened 2024-03-15
refactor: Rename GenericClusterConfig to GenericSpec

dlipovetsky:
This reduces stuttering, and aligns better with the names of the similar types AWSSpec and DockerSpec.

This is not an API change, because the struct is inlined, and the name is not part of the API.


api/v1alpha1/clusterconfig_types.go
```diff
@@ -41,14 +41,14 @@ type ClusterConfigSpec struct {
 	// +optional
 	Docker *DockerSpec `json:"docker,omitempty"`
 
-	GenericClusterConfig `json:",inline"`
+	GenericSpec `json:",inline"`
```
jimmidyson, line 44: `GenericSpec` seems... too generic to me, but I appreciate your point. How about `ProviderAgnosticSpec`? Or `ProviderCommonSpec`? `AllProvidersSpec`? `CommonSpec`? Regardless we can rename the field to be called `Common` perhaps?

```suggestion
	Common AllProvidersSpec `json:",inline"`
```

dlipovetsky: > Regardless we can rename the field to be called Common perhaps?
Sure

dlipovetsky: What I didn't say in the description is that I'm working with the `NodeConfig`, which has a corresponding `NodeConfigSpec`. Because both types are in the same package, I'd like to find a convention that works for both.

Would `AllProvidersSpec` and `AllProvidersNodeSpec` work for you?

dlipovetsky: I like this, and I tried it, but tests failed because [validation](https://github.com/d2iq-labs/capi-runtime-extensions/blob/5a05bc4d7753dcedad33ef82c0ba99c9b02dbf4b/common/pkg/testutils/openapi/validate.go#L80) found an unknown field `common`.

I wondered why, and then I learned that the `json:",inline"` tag has no effect; it's not supported, although it was proposed in 2013: https://github.com/golang/go/issues/6213.

So, to inline the fields in `AllProviderSpec`, we have to continue to embed it.

dlipovetsky: The YAML libraries (https://pkg.go.dev/gopkg.in/yaml.v2, https://pkg.go.dev/gopkg.in/yaml.v3) used by kubernetes do support the `inline` tag.

Maybe we can/should change validation?

dlipovetsky: Ahh, I think it's because we use json throughout. For example: https://github.com/d2iq-labs/capi-runtime-extensions/blob/fc2ef20037c0dbf955b6663b0dbd4c583e357bd7/common/pkg/testutils/capitest/variables.go#L67

jimmidyson: Let's just leave it embedded then? `AllProviderSpec` sgtm.

dkoshkin:
I think we can close this now that https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/537 is merged? `GenericClusterConfig` works now that each provider is embedding it.

jimmidyson:
Agreed @dkoshkin - @dlipovetsky if you disagree and would still like to do this rename, please rebase and reopen this PR.

PR #452, opened 2024-03-18
fix: Remove vendored types for core CAPI providers (CAPD, CABPK, KCP)

dlipovetsky:
We do not need to sync the Kubeadm Control Plane or Kubeadm Bootstrap provider types,
because they are developed in the sigs.k8s.io/cluster-api module, which we define as
a dependency.

We also do not need to sync the Docker infrastructure provider types, because they in the
sigs.k8s.io/cluster-api/test module, which is developed in parallel with
sigs.k8s.io/cluster-api. For that reason, we expect no dependency conflicts.


dkoshkin (approved):
Thank you

jimmidyson:
Thank you and sorry for the confusion caused by merging your previous PR.

jimmidyson (approved):

dlipovetsky:
(Removed some trailing whitespace and force-pushed)

PR #454, opened 2024-03-19
build: Rename module to d2iq-labs/cluster-api-runtime-extensions-nutanix

jimmidyson:
As previously discussed, renaming the module to `github.com/d2iq-labs/cluster-api-runtime-extensions-nutanix`, readying for a move to nutanix-cloud-native github org (hopefully happening soon!).


dkoshkin (commented):
Will this break using it as a gomod dep?

deepakm-ntnx (approved):
should we also include licensing header change in the same? separate is fine too

deepakm-ntnx:
are we planning to rename the repo name as well?

jimmidyson:
> Will this break using it as a gomod dep?

Yes - will need to update it but I'm going to rename the repo once this is merged.

> are we planning to rename the repo name as well?

Yes.

> should we also include licensing header change in the same? separate is fine too

Good point! Let me do it in this PR.

jimmidyson:
Fixed up licenses.

PR #455, opened 2024-03-19
docs: Update docsy module

jimmidyson:
Update to latest docsy module for documentation site.

### DIFF
--- a/docs/go.mod
+++ b/docs/go.mod
@@ -6,6 +6,5 @@ module github.com/d2iq-labs/capi-runtime-extensions/docs
 go 1.20
 
 require (
-	github.com/google/docsy v0.7.1 // indirect
-	github.com/google/docsy/dependencies v0.7.1 // indirect
+	github.com/google/docsy v0.9.1 // indirect
 )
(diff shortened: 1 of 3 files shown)


docs/go.mod
```diff
@@ -6,6 +6,5 @@ module github.com/d2iq-labs/capi-runtime-extensions/docs
 go 1.20
 
 require (
-	github.com/google/docsy v0.7.1 // indirect
-	github.com/google/docsy/dependencies v0.7.1 // indirect
+	github.com/google/docsy v0.9.1 // indirect
```
dkoshkin, line 9: Should we add this in the [dependebot config](https://github.com/d2iq-labs/capi-runtime-extensions/blob/main/.github/dependabot.yml) too?

dkoshkin (approved):

docs/go.mod
jimmidyson: Can do but it's not as important and I don't know how it works with hugo - had to do a `hugo mod get -u` to update - it does something with `hugo.toml` module sections. I think this is ok to manually update when we need to.

PR #469, opened 2024-03-20
feat: Add user configuration for all providers

dlipovetsky:
- Defines a cluster-level variable for defining one or more users
- Patches bootstrap templates for control plane and worker node pools
  with user configuration


api/v1alpha1/node_types.go
```diff
@@ -54,14 +57,84 @@ func (s NodeConfigSpec) VariableSchema() clusterv1.VariableSchema {
...
+	// Sudo specifies a sudo role for the user
+	// +optional
+	Sudo *string `json:"sudo,omitempty"`
+}
+
+func (User) VariableSchema() clusterv1.VariableSchema {
+	return clusterv1.VariableSchema{
+		OpenAPIV3Schema: clusterv1.JSONSchemaProps{
+			Description: "User",
+			Type:        "TBD",
```
supershal, line 115:
```suggestion
			Type:        "object",
```

dlipovetsky: Fixed, thanks

PR #470, opened 2024-03-20
build: publish list of images for addons

dkoshkin:
Added automation to publish a yaml with a list of images needed for each of the addons.

Suggestions on the format of the file are welcome!

### DIFF
--- a/addon-images.yaml
+++ b/addon-images.yaml
@@ -1,0 +1,47 @@
+# Copyright 2023 D2iQ, Inc. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+aws-cpi-v1.27:
+  version: v1.27.1
+  images:
+    - registry.k8s.io/provider-aws/cloud-controller-manager:v1.27.1
+aws-cpi-v1.28:
+  version: v1.28.1
+  images:
+    - registry.k8s.io/provider-aws/cloud-controller-manager:v1.28.1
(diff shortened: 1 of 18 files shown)


addon-images.yaml
```diff
@@ -0,0 +1,47 @@
+# Copyright 2023 D2iQ, Inc. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+aws-cpi-v1.27:
```
supershal, line 3: How are we plan to consume this file?

dkoshkin: Great question! The idea is that any consumer (e.g. konvoy2) can run a simple `yq` query to get the list of images.

```
 yq '.*.images' addon-images.yaml
 ```
 
 or 
 
 ```
 yq '.*.images' addon-images.yaml | tr -d ' -'
 ```

supershal: nice. each client wont have to build tooling to scrape images for bundling.

supershal:
The image file format in the PR is well readable. Should we consider the image file format used by mindthegap? https://github.com/mesosphere/mindthegap/blob/main/images-example.yaml 
I am just thinking it might be easier to consume it by the `dkp` CLI directly. The only thing is we will lose some details with mindthegap format.

dkoshkin:
Will discuss offline and reopen later

PR #496, opened 2024-04-11
build: Remove unused capbk and capd hack modules

jimmidyson:
These are unused as the project already has a dependency on CAPI and CAPI test
modules which includes the APIs for these providers.


dkoshkin (approved):

dlipovetsky (approved):
Thanks for cleaning this up

PR #497, opened 2024-04-11
feat: add helm chart field

faiq:
adds a helm chart field for customers who use a private OCI registry 

this closes https://github.com/d2iq-labs/cluster-api-runtime-extensions-nutanix/pull/39


api/v1alpha1/addon_types.go
```diff
@@ -46,6 +48,10 @@ func (Addons) VariableSchema() clusterv1.VariableSchema {
 			Description: "Cluster configuration",
 			Type:        "object",
 			Properties: map[string]clusterv1.JSONSchemaProps{
+				"helmChartRepository": {
+					Pattern:     patterns.HostAndOptionalPort,
+					Description: "Optional OCI registry used to pull helm charts for adons",
```
jimmidyson, line 53: Is this just an OCI registry or could it also be https helm repository somewhere?

jimmidyson, line 53: nit

```suggestion
					Description: "Optional OCI registry used to pull helm charts for addons",
```

jimmidyson: You could expand this description and add example too via `Example` property.

faiq: I'm not sure what the difference between helm repository and OCI registry is. That entity would need to host all of the charts though. because the repository field for each addon gets set to this value.

pkg/handlers/generic/lifecycle/config/cm.go
```diff
@@ -89,7 +110,10 @@ func (h *HelmChartGetter) For(
 	if !ok {
 		return nil, fmt.Errorf("did not find key %s in %v", name, cm.Data)
 	}
-	var settings HelmChart
-	err = yaml.Unmarshal([]byte(d), &settings)
-	return &settings, err
+	var chart HelmChart
+	err = yaml.Unmarshal([]byte(d), &chart)
```
supershal, line 114: The [logic](https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/497/files#diff-cf72970549260134569fee23f53ae5f544f32b05f043a5da47c4dcd3a35de86eR94) that reads the helmChartRepo can be moved here. If the configmap retrival errors out or the component config does not exists, we can return before reading the helmChartRepo variable.

api/v1alpha1/addon_types.go
jimmidyson: Helm repositories used to be supported only accessible via https. OCi support was later introduced meaning charts could be stored and referenced from OCI registries. The only difference is the scheme in the repository URL, using oci:// instead of https://. I think we should support both here.

supershal (approved):

api/v1alpha1/addon_types.go
faiq: I don't know if all the users are going to be able to push helm charts to a single helm repository im also not sure how to validate it?

jimmidyson: This looks like it's only supporting pulling from one helm repository though right for all Helm addons?

faiq: yeah, exactly the idea is that this source will have all of the helm charts the user needs

PR #498, opened 2024-04-11
feat: adds a validation hook

faiq:
the new version of https://github.com/d2iq-labs/cluster-api-runtime-extensions-nutanix/pull/56

adds a validation hook for https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/497

### DIFF
--- a/api/variables/variables.go
+++ b/api/variables/variables.go
@@ -1,0 +1,33 @@
+// Copyright 2023 D2iQ, Inc. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package variables
+
+import (
+	"encoding/json"
+	"fmt"
+
+	runtimehooksv1 "sigs.k8s.io/cluster-api/exp/runtime/hooks/api/v1alpha1"
+)
+
+func UnmarshalRuntimeVariable[T any](runtimeVariable *runtimehooksv1.Variable, obj *T) error {
+	err := json.Unmarshal(runtimeVariable.Value.Raw, obj)
+	if err != nil {
+		return fmt.Errorf("error unmarshalling variable: %w", err)
+	}
+
+	return nil
+}
+
+//nolint:gocritic // no need for named results
+func GetRuntimhookVariableByName(
+	name string,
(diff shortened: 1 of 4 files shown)


api/variables/variables.go
```diff
@@ -0,0 +1,33 @@
...
+	}
+
+	return nil
+}
+
+//nolint:gocritic // no need for named results
+func GetRuntimhookVariableByName(
+	name string,
+	variables []runtimehooksv1.Variable,
+) (*runtimehooksv1.Variable, int) {
```
supershal, line 26: is there any usecase where returned index of the variable can be useful instead of just returning (nil) or (nil, false) ?

pkg/handlers/generic/validation/helm/handler.go
```diff
@@ -0,0 +1,87 @@
...
+		client:       c,
+		variableName: clusterconfig.MetaVariableName,
+	}
+}
+
+func (h *HelmRegistryValidator) Name() string {
+	return "HelmRegistryValidator"
+}
+
+func (h *HelmRegistryValidator) ValidateTopology(
```
supershal, line 39: can we please add a unit test for this

api/variables/variables.go
```diff
@@ -0,0 +1,33 @@
...
+	err := json.Unmarshal(runtimeVariable.Value.Raw, obj)
+	if err != nil {
+		return fmt.Errorf("error unmarshalling variable: %w", err)
+	}
+
+	return nil
+}
+
+//nolint:gocritic // no need for named results
+func GetRuntimhookVariableByName(
```
faiq, line 23:
```suggestion
func GetRuntimehookVariableByName(
```

pkg/handlers/generic/validation/helm/handler.go
faiq: Im not sure about this particular function, but we can split it out to be more testable

api/variables/variables.go
faiq: the variable could exist and  just be nil

PR #499, opened 2024-04-11
feat: Update CAPX to v1.4.0-alpha.2

deepakm-ntnx:


hack/third-party/capx/go.mod
```diff
@@ -3,67 +3,41 @@
 
 module github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/external/capx
 
-go 1.21
+go 1.22.1
```
jimmidyson, line 6: nit: Please keep as `1.22`
```suggestion
go 1.22
```

jimmidyson (approved):
One nit but LGTM!

hack/third-party/capx/go.mod
jimmidyson: I think you can fix here separately though right?

deepakm-ntnx: sure. was trying to avoid  make apis.sync to overwrite it again. sure we can do it separately

deepakm-ntnx: seems like pre-commit overwrites it so we need to fix it in capx only first

jimmidyson: Oh interesting because it's declared as such in CAPX it's the minimum this module can use too... Still learning about these go module intricacies 😅

dkoshkin: Looks like we do need to fix it upstream first :( 
The tests are failing

jimmidyson: 🤔 can't we just add 1.22.1 in this module for now? Shouldn't affect anything else once the types are generated. Revert my suggestion basically 😅

deepakm-ntnx: https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pull/410/files

jimmidyson:
Replaced by https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/512.

PR #500, opened 2024-04-11
refactor: standardize the code for getting Helm values

dkoshkin:
I noticed there were 3 copies of `retrieveValuesTemplateConfigMap()` in addition to the one generic one in `lifecycleutils`, refactored all of these methods to reuse the existing function.

~While refactoring I saw that for NFD the handler was adding a value, moved the templating to the helm values. This copies the pattern in other addons, but also makes it easier to override at cluster creation time (kind of as a last resort if there is a bug in the handler code)~ Reverted in 5f24ee0e2acc99e383e021bc5203994fc53c2308, we don't get the standard`.Chart.Version` when CAAPH templates the values, will file a separate PR to gotemplate this in the handler.

I also added `RetrieveValuesTemplate` to get the data instead of having `valuesTemplateConfigMap.Data["values.yaml"]` in every handler.

### DIFF
--- a/pkg/handlers/generic/lifecycle/utils/utils.go
+++ b/pkg/handlers/generic/lifecycle/utils/utils.go
@@ -122,11 +122,11 @@ func RetrieveValuesTemplateConfigMap(
 	ctx context.Context,
 	c ctrlclient.Client,
 	configMapName,
-	defaultsNamespace string,
+	namespace string,
 ) (*corev1.ConfigMap, error) {
 	configMap := &corev1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
-			Namespace: defaultsNamespace,
+			Namespace: namespace,
 			Name:      configMapName,
 		},
 	}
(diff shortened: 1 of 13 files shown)


faiq (approved):
great changes!

pkg/handlers/generic/lifecycle/utils/utils.go
```diff
@@ -144,6 +144,19 @@ func RetrieveValuesTemplateConfigMap(
...
+func RetrieveValuesTemplate(
+	ctx context.Context,
+	c ctrlclient.Client,
+	configMapName,
+	namespace string,
+) (string, error) {
+	configMap, err := RetrieveValuesTemplateConfigMap(ctx, c, configMapName, namespace)
+	if err != nil {
+		return "", err
+	}
```
faiq, line 156: just a nit, wrap if you can.

pkg/handlers/generic/lifecycle/utils/utils.go
```diff
@@ -122,11 +122,11 @@ func RetrieveValuesTemplateConfigMap(
 	ctx context.Context,
```
faiq, line 122: I guess its ok to leave both

supershal (approved):

jimmidyson (approved):

PR #505, opened 2024-04-11
feat: use CAREN to set default cert SANS

faiq:
This replaces https://github.com/d2iq-labs/cluster-api-runtime-extensions-nutanix/pull/20

This adds the logic to add necessary default Extra Sans to the `KubeadmControlPlane` for Docker Clusters via a mutation hook.


faiq:
Copying a comment over, for posterity and to continue discussion 


@supershal 
>  > WDYT of creating new separate patch dockerExtraAPIServerCertSANsPatch for Docker MetaVariable. Ensure that both lifecycle patch extraAPIServerCertSANsPatch and docker patchdockerExtraAPIServerCertSANsPatch preserves existing list of certSANs. i.e. something like

> A bit more details: I think it would be cleaner to have provider specific logic in the provider's handler. i.e. pkg/handlers/docker/mutation and generic handlers can be free of any logic that is based on provider. Having a docker mutator that always adds default SANs could be cleaner IMHO

@faiq 
> I disagree. I think adding another handler that touches the same exact fields on objects can be more prone to errors and harder for a developer to discover what gets applied when

pkg/handlers/generic/mutation/extraapiservercertsans/inject.go
```diff
@@ -64,11 +79,13 @@ func (h *extraAPIServerCertSANsPatchHandler) Mutate(
...
 		return err
 	}
-	if !found {
-		log.V(5).Info("Extra API server cert SANs variable not defined")
+	if !found && len(defaultAPICertSANs) == 0 {
+		log.V(5).Info("No Extra API server cert SANs needed to be added")
 		return nil
 	}
 
+	extraSans := append(extraAPIServerCertSANsVar, defaultAPICertSANs...)
```
dkoshkin, line 87: nit, the "extraSans" being a combination of "extraAPIServerCertSANsVar" and "defaultAPICertSANs" is a little confusing, wdyt of just `apiServerCertSANsVar`?

```suggestion
	apiServerCertSANsVar := append(extraAPIServerCertSANsVar, defaultAPICertSANs...)
```

pkg/handlers/generic/mutation/extraapiservercertsans/inject.go
```diff
@@ -83,15 +100,27 @@ func (h *extraAPIServerCertSANsPatchHandler) Mutate(
...
 			}
-			obj.Spec.Template.Spec.KubeadmConfigSpec.ClusterConfiguration.APIServer.CertSANs = extraAPIServerCertSANsVar
-
+			obj.Spec.Template.Spec.KubeadmConfigSpec.ClusterConfiguration.APIServer.CertSANs = extraSans
 			return nil
 		},
 	)
 }
+
+func getDefaultAPIServerSANs(cluster *capiv1.Cluster) []string {
```
dkoshkin, line 115: We now also have it in the Nutanix CC, please handle that one too

https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/15224874e594e057403b7d521aabd40a44dc5053/charts/cluster-api-runtime-extensions-nutanix/defaultclusterclasses/nutanix-cluster-class.yaml#L124-L128

pkg/handlers/generic/mutation/extraapiservercertsans/inject.go
```diff
@@ -83,15 +100,27 @@ func (h *extraAPIServerCertSANsPatchHandler) Mutate(
...
+	provider, ok := cluster.Labels[capiv1.ProviderNameLabel]
+	if !ok {
+		return []string{}
+	}
+	switch provider {
+	case "docker":
```
dkoshkin, line 121:
```suggestion
	case v1alpha1.DockerProvider:
```

pkg/handlers/generic/mutation/extraapiservercertsans/inject.go
faiq: I'll stack to this implementation.

pkg/handlers/generic/mutation/extraapiservercertsans/inject.go
faiq: i like it!

supershal:
>I disagree. I think adding another handler that touches the same exact fields on objects can be more prone to errors and harder for a developer to discover what gets applied when

Please check an alternative: https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/compare/faiq/docker-sans...shalin/docker-sans . It was able to add default cert sans for docker even if `clusterConfig.docker`  variable was not set for a docker cluster.  
Also changed existing `extraAPIServerCertSANsPatchHandler` to ensure it preserves any cert SANs set by other handlers.
We should discuss if multiple handlers should mutate same fields on object or not

PR #514, opened 2024-04-15
feat: give mutators a clusterGetter function

faiq:
Adds a reusable function across mutators to return a cluster. This is lazily evaluated and happens once per mutation

tested manually be creating a docker cluster 

```
apiVersion: v1
items:
- apiVersion: controlplane.cluster.x-k8s.io/v1beta1
  kind: KubeadmControlPlane
  metadata:
    annotations:
      cluster.x-k8s.io/cloned-from-groupkind: KubeadmControlPlaneTemplate.controlplane.cluster.x-k8s.io
      cluster.x-k8s.io/cloned-from-name: docker-quick-start-control-plane
    creationTimestamp: "2024-04-15T21:20:06Z"
    finalizers:
    - kubeadm.controlplane.cluster.x-k8s.io
    generation: 1
    labels:
      cluster.x-k8s.io/cluster-name: docker-cluster-cilium-helm-addon
      topology.cluster.x-k8s.io/owned: ""
    name: docker-cluster-cilium-helm-addon-kr6lc
    namespace: default
    ownerReferences:
    - apiVersion: cluster.x-k8s.io/v1beta1
      blockOwnerDeletion: true
      controller: true
      kind: Cluster
      name: docker-cluster-cilium-helm-addon
      uid: d7d3b9b5-3a4c-4a2f-b5b1-1b46b3c6f9e9
    resourceVersion: "2512"
    uid: 11809f5a-5202-4fac-bd7b-7c96b17d437c
  spec:
    kubeadmConfigSpec:
      clusterConfiguration:
        apiServer:
          certSANs:
          - localhost
          - 127.0.0.1
          - 0.0.0.0
          - host.docker.internal
```

### DIFF
--- a/common/pkg/capi/clustertopology/handlers/mutation/meta.go
+++ b/common/pkg/capi/clustertopology/handlers/mutation/meta.go
@@ -5,10 +5,13 @@ package mutation
 
 import (
 	"context"
+	"fmt"
+	"sync"
 
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	runtimehooksv1 "sigs.k8s.io/cluster-api/exp/runtime/hooks/api/v1alpha1"
 	"sigs.k8s.io/cluster-api/exp/runtime/topologymutation"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -24,6 +27,8 @@ type MutateFunc func(
 	clusterKey client.ObjectKey,
 ) error
 
+type ClusterGetter func(context.Context) (*clusterv1.Cluster, error)
+
 type MetaMutator interface {
 	Mutate(
 		ctx context.Context,
@@ -31,20 +36,24 @@ type MetaMutator interface {
 		vars map[string]apiextensionsv1.JSON,
 		holderRef runtimehooksv1.HolderReference,
 		clusterKey client.ObjectKey,
+		getCluster ClusterGetter,
 	) error
 }
 
 type metaGeneratePatches struct {
 	name     string
 	mutators []MetaMutator
+	cl       client.Client
 }
 
 func NewMetaGeneratePatchesHandler(
 	name string,
+	cl client.Client,
 	mutators ...MetaMutator,
 ) handlers.Named {
 	return metaGeneratePatches{
 		name:     name,
+		cl:       cl,
 		mutators: mutators,
 	}
 }
@@ -53,13 +62,32 @@ func (mgp metaGeneratePatches) Name() string {
 	return mgp.name
 }
 
+func (mgp metaGeneratePatches) CreateClusterGetter(
+	clusterKey client.ObjectKey,
+) func(context.Context) (*clusterv1.Cluster, error) {
+	return func(ctx context.Context) (*clusterv1.Cluster, error) {
+		var (
+			cluster clusterv1.Cluster
+			err     error
+			once    sync.Once
+		)
+		once.Do(func() {
+			err = mgp.cl.Get(ctx, clusterKey, &cluster)
+		})
+		if err != nil {
+			return nil, fmt.Errorf("failed to fetch cluster %w", err)
+		}
+		return &cluster, nil
+	}
+}
+
 func (mgp metaGeneratePatches) GeneratePatches(
 	ctx context.Context,
 	req *runtimehooksv1.GeneratePatchesRequest,
 	resp *runtimehooksv1.GeneratePatchesResponse,
 ) {
 	clusterKey := handlers.ClusterKeyFromReq(req)
-
+	clusterGetter := mgp.CreateClusterGetter(clusterKey)
 	topologymutation.WalkTemplates(
 		ctx,
 		unstructured.UnstructuredJSONScheme,
@@ -72,7 +100,7 @@ func (mgp metaGeneratePatches) GeneratePatches(
 			holderRef runtimehooksv1.HolderReference,
 		) error {
 			for _, h := range mgp.mutators {
-				if err := h.Mutate(ctx, obj.(*unstructured.Unstructured), vars, holderRef, clusterKey); err != nil {
+				if err := h.Mutate(ctx, obj.(*unstructured.Unstructured), vars, holderRef, clusterKey, clusterGetter); err != nil {
 					return err
 				}
 			}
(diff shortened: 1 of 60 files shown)


common/pkg/capi/clustertopology/handlers/mutation/meta.go
```diff
@@ -59,6 +68,18 @@ func (mgp metaGeneratePatches) GeneratePatches(
 	resp *runtimehooksv1.GeneratePatchesResponse,
 ) {
 	clusterKey := handlers.ClusterKeyFromReq(req)
+	getCluster := func(ctx context.Context) (*clusterv1.Cluster, error) {
```
dkoshkin, line 71: Can you please provide the motivation I'm not sure I understand.


We already have the client and the clusterKey (from the request), would it not be simpler to have a helper function similar to `ClusterKeyFromReq` that can be used only in the handlers that need it?

common/pkg/capi/clustertopology/handlers/mutation/meta.go
```diff
@@ -59,6 +68,18 @@ func (mgp metaGeneratePatches) GeneratePatches(
...
+		cluster := &clusterv1.Cluster{}
+		var err error
+		fn := sync.OnceFunc(func() {
+			err = mgp.cl.Get(ctx, clusterKey, cluster)
+		})
+		fn()
```
dkoshkin, line 77: Why is `sync.OnceFunc` needed?

faiq: This would be so it would happen once per cluster mutation.

common/pkg/capi/clustertopology/handlers/mutation/meta.go
faiq: We don't have the client for all the handlers it's something optional that isn't required. This makes it so that getting the cluster is called once across all the providers and so we don't run into concurrency issues

https://nutanix.slack.com/archives/C069QMDB2QM/p1713193239746189?thread_ts=1712965566.988039&cid=C069QMDB2QM

dkoshkin, line 71: nit: match what is in the handlers?
```suggestion
	clusterGetter := func(ctx context.Context) (*clusterv1.Cluster, error) {
```

common/pkg/capi/clustertopology/handlers/mutation/meta_test.go
```diff
@@ -37,6 +37,7 @@ func (h *testHandler) Mutate(
 	_ map[string]apiextensionsv1.JSON,
 	holderRef runtimehooksv1.HolderReference,
 	_ client.ObjectKey,
+	_ ClusterGetter,
```
dkoshkin, line 40: To me this really smells like an interface, something like:

```
type ClusterInfoGetter interface {
  func Cluster(context.Context) (*clusterv1.Cluster, error)
  func InfrastructureRefKind(context.Context) string
}
```

WDYT?

dkoshkin: I think it would simplify writing unit tests too.

faiq: an interface would make writing unit tests easier. however this won't be that straight forward.

every single unit test creates a NewMetaGeneratePatchesHandler for its test. 

```golang
    return mutation.NewMetaGeneratePatchesHandler("", cl, NewPatch()).(mutation.GeneratePatches)
```

to simplify the tests to allow mocks we would need to change the function definition to take in a constructor to create this interface

changing this to (not sure about this `Factory` name

```golang

type ClusterInfoGetterFactory interface {
  CreateClusterInfoGetter() (ClusterInfoGetter, error)
}

type metaGeneratePatches struct {
	name     string
	mutators []MetaMutator
	clusterInfoGetter ClusterInfoGetter
}

func NewMetaGeneratePatchesHandler(
	name string,
	factory ClusterInfoGetterFactory,
	mutators ...MetaMutator,
) (handlers.Named, error) {
        getter, err := factory.CreateClusterInfoGetter()
        if err != nil {
           return nil, err
        }
	return metaGeneratePatches{
		name:     name,
		mutators: mutators,
		clusterInfoGetter: getter,
	}, nil
}
```
the metaMutator interface might change as follows

```golang
type MetaMutator interface {
	Mutate(
		ctx context.Context,
		obj *unstructured.Unstructured,
		vars map[string]apiextensionsv1.JSON,
		holderRef runtimehooksv1.HolderReference,
               clusterGetter ClusterInfoGetter,
	) error
}
```

I think this would allow us to pass in our implementations of CreateClusterInfoGetters and also allow us to have a default one from client.

With all this being said, I don't really know if all this effort is going to be worth it.

.golangci.yml
```diff
@@ -30,6 +30,7 @@ linters:
     - gosimple
     - govet
     - ineffassign
+    - importas
```
jimmidyson, line 33: I really like this change so I pulled it out to #518 to reduce the changes in this PR unrelated to the functionality it's providing.

common/pkg/capi/clustertopology/handlers/mutation/meta.go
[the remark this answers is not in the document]
jimmidyson: Oh I see what you're doing now! Howeever it's still not quite right, because every call to `clusterGetter` is going to call `sync.OnceFunc`, which won't have the laziness you're after: it will always be run...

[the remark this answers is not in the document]
jimmidyson: It's all about the scoping of the `sync.Once` call and the returned variable. Here's a repro of this problem - https://go.dev/play/p/x0pnX273hhd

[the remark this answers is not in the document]
jimmidyson: I think what you want to do is like I shared in Slack, here it is again (slightly adapted), remembering we need it this way to pass the context properly for timeouts, cancellations, etc:

```
type ClusterGetterFn func(context.Context) (*capiv1.Cluster, error)

func ClusterGetterForObjectKey(cl client.Client, clusterKey client.ObjectKey) ClusterGetterFn {
    var (
        getOnce sync.Once
        cluster *capiv1.Cluster
        err error
    )
    return func(ctx context.Context) (*capiv1.Cluster, error) {
        getOnce.Do(func() {
          innerCluster := &clusterv1.Cluster{}
          innerErr := client.Get(ctx, innerCluster, clusterKey)
          if innerErr != nil {
            err = fmt.Errorf("failed to retrieve cluster: %w" innerErr)
          } else {
            cluster = innerCluster
          }
        }
        return cluster, err
    }
}
```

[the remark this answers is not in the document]
faiq: Gotcha! makes sense. Thank you :bow:

.golangci.yml
faiq: yeah just added this because at some point a lot of the packages were referencing v1cluster1.Cluster

common/pkg/capi/clustertopology/handlers/mutation/meta.go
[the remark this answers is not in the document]
faiq: done in the latest commit

jimmidyson (approved):

supershal (approved):
LGTM. Thanks for generic solution.

PR #517, opened 2024-04-16
build: Specify go1.22.2 as toolchain to fix govulncheck issues

jimmidyson:
Nix (and therefore devbox) has been slow in rolling out go1.22.2, which
contains CVE fixes. Current version go1.22.1 causes govulncheck to
report valid vulnerabilities in `net/http` package. go1.21 introduced
toolchain management via `go.mod` file with `toolchain` directive. This
commit specifies go1.22.2 as the toolchain to use and hence fixes the
govulncheck issues.

This does mean that go versions have to be managed in multiple places so
this is a stop-gap until Nix releases go1.22.2 to nixpkgs-unstable
channel.

### DIFF
--- a/api/go.mod
+++ b/api/go.mod
@@ -5,6 +5,8 @@ module github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/ap
 
 go 1.21
 
+toolchain go1.22.2
+
 require (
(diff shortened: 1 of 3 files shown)


mhrabovcin (approved):

deepakm-ntnx (approved):
can we move the go.mod to use go 1.22.1 as well?

PR #522, opened 2024-04-16
feat: adds nutanix SANs via patchHandler

faiq:

https://jira.nutanix.com/browse/D2IQ-100519

### DIFF
--- a/api/v1alpha1/clusterconfig_types.go
+++ b/api/v1alpha1/clusterconfig_types.go
@@ -38,6 +38,12 @@ var DefaultDockerCertSANs = []string{
 	"host.docker.internal",
 }
 
+var DefaultNutanixCertSANs = []string{
+	"localhost",
+	"127.0.0.1",
+	"0.0.0.0",
+}
+
 // +kubebuilder:object:root=true
 
 // ClusterConfig is the Schema for the clusterconfigs API.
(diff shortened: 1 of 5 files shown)


supershal (approved):
LGTM

api/v1alpha1/clusterconfig_types.go
```diff
@@ -37,6 +37,12 @@ var DefaultDockerCertSANs = []string{
 	"host.docker.internal",
 }
 
+var DefaultNutanixCertSANs = []string{
+	"localhost",
+	"127.0.0.1",
+	"0.0.0.0",
+}
```
jimmidyson, line 44: nit: let's group all these vars.:

```suggestion
var (
	DefaultDockerCertSANs = []string{
		"localhost",
		"127.0.0.1",
		"0.0.0.0",
		"host.docker.internal",
	}

	DefaultNutanixCertSANs = []string{
		"localhost",
		"127.0.0.1",
		"0.0.0.0",
	}
)
```

api/v1alpha1/clusterconfig_types.go
```diff
@@ -271,8 +277,10 @@ func (ExtraAPIServerCertSANs) VariableSchema() clusterv1.VariableSchema {
 	return clusterv1.VariableSchema{
 		OpenAPIV3Schema: clusterv1.JSONSchemaProps{
 			Description: fmt.Sprintf(
-				"Extra Subject Alternative Names for the API Server signing cert. For Docker %v are injected automatically.",
+				//nolint:lll // its a user facing message
+				"Subject Alternative Names for the API Server signing cert. For Docker %v are injected automatically. For Nutanix %v are injected automatically.",
```
jimmidyson, line 281: This will start to become unwieldy when more providers are added. I wonder if we can follow up and provide only a description for the provider this variable schema is a part of? Considering variable schema is ultimately provider specific, this should be possible.

pkg/handlers/generic/mutation/extraapiservercertsans/inject.go
```diff
@@ -94,6 +94,10 @@ func (h *extraAPIServerCertSANsPatchHandler) Mutate(
 		"variableValue",
 		apiCertSANs,
 	)
+	if len(apiCertSANs) == 0 {
+		log.Info("No APIServerSANs to apply")
```
jimmidyson, line 98: Do we need to this to be always output?
```suggestion
		log.V(4).Info("No APIServerSANs to apply")
```

dkoshkin (approved):

supershal (approved):

PR #525, opened 2024-04-16
feat: Add additionalCategories field to Nutanix machine details patch

dlipovetsky:
Adds the `additionalCategories` field to the Nutanix machine details patch.

Fixes https://jira.nutanix.com/browse/D2IQ-100464

I amended the machine details patch test.


api/v1alpha1/nutanix_node_types.go
```diff
@@ -54,6 +54,10 @@ type NutanixMachineDetails struct {
 	// or using the prism_central API.
 	Subnets NutanixResourceIdentifiers `json:"subnets"`
 
+	// List of categories that need to be added to the machines. Categories must already exist in Prism Central
+	//+optional
```
dkoshkin, line 58: nit: I don't think it makes a difference, but I see most of the other optional fields have the space, plus my IDE complains about the no space
>Comment should start with a whitespace 
```suggestion
	// +optional
```

docs/content/customization/nutanix/machine-details.md
```diff
@@ -50,6 +50,9 @@ spec:
               subnets:
               - name: subnet-name
                 type: name
+              additionalCategories:
+              - key: example-key
+                value: example-value
```
dkoshkin, line 55: Assuming the ordering does not matter, what do we think about simplifying this api in CAREN to a map?
```
additionalCategories:
  example-key: example-value
```

dlipovetsky: As long as there is no technical reason to use maps.

One problem that came to mind be that some category key is not a valid JSON key. Since a category key is a string, and a JSON key can be any string (even an escaped one), that's not a problem.

dlipovetsky: I'd like to know what @deepakm-ntnx thinks.

dlipovetsky: We talked offline. @deepakm-ntnx made a great observation: users must be able to define multiple values for the same _category key_, which would not be possible if we use JSON keys for category keys, because the JSON keys must be unique.

jimmidyson: Why not map[string][]string? Having the same key multiple times looks a bit odd to me.

deepakm-ntnx: sure, map[string][]string also makes sense but we need to be careful as to what is an acceptable value for key in json map vs in category keys. as long as that matches we can do whatever is shorter. eg i can create a category with key -1234 (please dont ask why but allowed thru categories UI) and value as xyz. but per https://google.github.io/styleguide/jsoncstyleguide.xml?showone=Property_Name_Format#Property_Name_Format this is not a valid key format. hence existing way makes more sense

deepakm-ntnx (approved):

jimmidyson (approved):
Let's get this merged, although I'm not totally sold on the ability to specify the same key multiple times (key in my mind is a unique identifier). However, I can't think of a different solution, other than adding value as an array, which would make the (possibly more common) use case of single value per key to be less user friendly.

api/v1alpha1/nutanix_node_types.go
```diff
@@ -175,3 +187,38 @@ func (NutanixResourceIdentifiers) VariableSchema() clusterv1.VariableSchema {
...
+			Type:        "object",
+			Properties: map[string]clusterv1.JSONSchemaProps{
+				"key": {
+					Type:        "string",
+					Description: "key is the Key of category in PC.",
+				},
+				"value": {
+					Type:        "string",
+					Description: "value is the category value linked to the category key in PC.",
+				},
```
jimmidyson, line 206: Let's add validation constraints via format and/or pattern

dkoshkin:
I just tried creating a duplicate key and thats not allowed:

image
Although you can have multiple values for a key:
image

But, from what I can see the CAPX controller will actually just take whatever the last value for any duplicate keys, 
https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/blob/f6e079c5dd386256a27974338a111da3b628fb9b/controllers/helpers.go#L602-L614

In which case a `map[string]string` would give us this enforcement for "free".

dkoshkin:
Ha! It goes deeper, a VM can be created with 1 key but multiple values for that key, so maybe there is a bug in CAPX in how it handles duplicate keys?

image

deepakm-ntnx:
> 
> But, from what I can see the CAPX controller will actually just take whatever the last value for any duplicate keys, https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/blob/f6e079c5dd386256a27974338a111da3b628fb9b/controllers/helpers.go#L602-L614
> 
> In which case a `map[string]string` would give us this enforcement for "free".

we should fix the CAPX to allow being able to specify same key with multiple values. cc @yannick

docs/content/customization/nutanix/machine-details.md
dlipovetsky: Thanks for the discussion, though!

api/v1alpha1/nutanix_node_types.go
```diff
@@ -54,6 +55,11 @@ type NutanixMachineDetails struct {
 	// or using the prism_central API.
 	Subnets NutanixResourceIdentifiers `json:"subnets"`
 
+	// List of categories that need to be added to the machines. Categories must already
+	// exist in Prism Central. One category key can have more than one value.
+	// +optional
+	AdditionalCategories NutanixCategoryIdentifiers `json:"additionalCategories,omitempty"`
```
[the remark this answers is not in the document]
dlipovetsky: Done

dkoshkin (approved):

docs/content/customization/nutanix/machine-details.md
dlipovetsky: I experimented a little bit more. As far as I can tell, the set of JSON strings is a subset of Nutanix category keys.

So, in theory, we can support something like
```
additionalCategories:
  "-1234":
    - value1
    - value2
  anotherkey:
    - value1
    - value2
```

That's shorter than:

```
additionalCategories:
  - key: "-1234"
    value: value1
  - key: "-1234"
    value: value2
  - key: anotherkey
    value: value1
  - key: anotherkey
    value: value2
```

On the other hand, explicit fields might be easier to understand.

api/v1alpha1/nutanix_node_types.go
dlipovetsky: I can find no limitations on either key or value.

- https://www.nutanixbible.com/classic#categories

dlipovetsky:
I rebased on main to resolve conflicts, and force-pushed

api/v1alpha1/nutanix_node_types.go
```diff
@@ -6,6 +6,7 @@ package v1alpha1
 import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/utils/ptr"
+	"k8s.io/utils/ptr"
```
dkoshkin, line 9:
```suggestion
```

dlipovetsky: Thanks. I also need to re-generate code.

dlipovetsky:
I had some issues rebasing on main while preserving the fixups, so I squashed locally.

dkoshkin (approved):

deepakm-ntnx:
@dlipovetsky pls add docs too so that it shows up here https://d2iq-labs.github.io/cluster-api-runtime-extensions-nutanix/customization/nutanix/

PR #533, opened 2024-04-18
refactor: Use separate types for provider cluster configs

jimmidyson:
This commit also switches to using controller-gen generated CRDs
for the OpenAPI schema for the variable definitions.

A lot of this commit is refactoring tests.


dkoshkin:
@jimmidyson any chance you can split this PR into? 😬 
1. The refactor
2. autogenerating openapi
3. fix the domain and api groups

jimmidyson:
It's actually a but tricky to split it all up but i'll try my best. Any thoughts on the approach though?

PR #550, opened 2024-04-22
build: Export variable schema to YAML files

jimmidyson:
This makes it easiet to compare schema changes in pull requests, rather than
relying on review of quite opaque variable schema definitions in code.


jimmidyson:
I think we can close this now we have CRD YAML generation.

PR #554, opened 2024-04-22
refactor: provider an entrypoint to the infra provider meta handlers

dkoshkin:
This aligns the structure to how the lifecycle handlers are created. More importantly it enables us to pass globalOptions to those infra handlers that need it.

The goal of this is to use for the eventual refactor to support a more generic kube-vip implementation. But for that I need to be able to read a ConfigMap from the "default" namespace where all of these CMs are created. Again similar to how its done for the CAAPH addons https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/33b29dc368622777d8dbea004077532c2a84d27a/pkg/handlers/generic/lifecycle/nfd/strategy_helmaddon.go#L56-L61

Refactor to support work for https://jira.nutanix.com/browse/D2IQ-100364


deepakm-ntnx:
Could you please add some sample output of how the options are passed as a test case

jimmidyson (approved):

deepakm-ntnx (approved):
pls see my comment above.

dkoshkin:
> Could you please add some sample output of how the options are passed as a test case

There are new unit tests using this in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/558/commits

PR #556, opened 2024-04-23
feat: Preserve user-managed fields when applying resources

dlipovetsky:
We want to preserve user-managed fields when we create/update specific resources. If a value in a user-managed field conflicts with our value, we want to alert the user by returning an error, rather than overwriting the value.

This PR exposes the `ForceOwnership` option to callers, but does not change the behavior of existing callers. 

Fixes #

### DIFF
--- a/common/pkg/k8s/client/apply.go
+++ b/common/pkg/k8s/client/apply.go
@@ -7,27 +7,33 @@ import (
 	"context"
 	"fmt"
 
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-// ServerSideApply will apply (i.e. create or update) objects via server-side apply. This will overwrite any changes
-// that have been manually applied to fields managed by CAREN.
+const (
+	// FieldOwner is the field manager name used for server-side apply.
+	FieldOwner = "d2iq-cluster-api-runtime-extensions-nutanix"
+)
+
+// ForceOwnership is an convenience alias of the same option in the controller-runtime client.
+var ForceOwnership = client.ForceOwnership
+
+// ServerSideApply will apply (i.e. create or update) objects via server-side apply.
 func ServerSideApply(
 	ctx context.Context,
 	c ctrlclient.Client,
-	objs ...ctrlclient.Object,
+	obj ctrlclient.Object,
+	opts ...ctrlclient.PatchOption,
 ) error {
-	for i := range objs {
-		err := c.Patch(
-			ctx,
-			objs[i],
-			ctrlclient.Apply,
-			ctrlclient.FieldOwner("d2iq-cluster-api-runtime-extensions-nutanix"),
-		)
-		if err != nil {
-			return fmt.Errorf("server-side apply failed: %w", err)
-		}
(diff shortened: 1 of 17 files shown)


common/pkg/k8s/client/apply.go
```diff
@@ -22,7 +22,6 @@ func ServerSideApply(
 			ctx,
 			objs[i],
 			ctrlclient.Apply,
-			ctrlclient.ForceOwnership,
```
jimmidyson, line 25: AFAIK `controller-runtime` doesn't actually fully support SSA - see discussion in https://github.com/kubernetes-sigs/controller-runtime/issues/347#issuecomment-1470092104.

There is work ongoing to support `ApplyConfigurations` but right now if you do not send all fields and force ownership then there are issues determining whether fields are unset orset-and-zero'd. I'm not completely clear on the implications of this, but I would prefer to keep the existing `ForceOwnership` behaviour for current usage and add the ability to not force ownership for your namespace use case. WDYT?

jimmidyson: I pushed https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/559 to prove to myself that this does indeed work and it does... which is good, but I still have concerns about this for the general SSA case.

jimmidyson: Sounds good to me! Let's make it opt-in to `ForceOwnership` by requiring to pass in that option too. It will require changes to existing calls to this function but that's OK.

common/pkg/k8s/client/apply.go
```diff
@@ -7,28 +7,33 @@ import (
 	"context"
 	"fmt"
 
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
```
jimmidyson, line 11: Duplicate imports
```suggestion
	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
```

common/pkg/k8s/client/apply.go
```diff
@@ -7,28 +7,33 @@ import (
...
 
-// ServerSideApply will apply (i.e. create or update) objects via server-side apply. This will overwrite any changes
-// that have been manually applied.
+const (
+	// FieldOwner is the field manager name used for server-side apply.
+	FieldOwner = "d2iq-cluster-api-runtime-extensions-nutanix"
+)
+
+// ForceOwnership is an convenience alias of the same option in the controller-runtime client.
+var ForceOwnership = client.ForceOwnership
```
jimmidyson, line 20:
```suggestion
var ForceOwnership = ctrlclient.ForceOwnership
```

common/pkg/k8s/client/apply.go
dlipovetsky: Thanks! I blame gopls :smile:

common/pkg/k8s/client/apply.go
dlipovetsky: Thanks, fixed!

jimmidyson (approved):

dkoshkin (approved):

PR #557, opened 2024-04-23
feat: Preserve user-managed fields when creating namespace

dlipovetsky:
We want to ensure the namespace exists, while preserving user-managed fields (i.e. labels and annotations) namespace exists.

Currently, if the namespace already exists, we do not update it. This prevents us from applying our own labels and annotations to an existing namespace, something we need to do in certain cases. This PR allows us to, in a future PR, apply labels and annotations to an existing namespace.

Fixes #

I added a new integration (envtest-based) test that verifies the desired behavior.

Depends on #556.

### DIFF
--- a/pkg/handlers/generic/lifecycle/utils/utils.go
+++ b/pkg/handlers/generic/lifecycle/utils/utils.go
@@ -93,8 +93,8 @@ func EnsureCRSForClusterFromObjects(
 	return nil
 }
 
-// EnsureNamespace will create the namespace if it does not exist.
-func EnsureNamespace(ctx context.Context, c ctrlclient.Client, name string) error {
+// EnsureNamespaceWithName will create the namespace with the specified name if it does not exist.
+func EnsureNamespaceWithName(ctx context.Context, c ctrlclient.Client, name string) error {
 	ns := &corev1.Namespace{
 		TypeMeta: metav1.TypeMeta{
 			APIVersion: corev1.SchemeGroupVersion.String(),
@@ -105,6 +105,17 @@ func EnsureNamespace(ctx context.Context, c ctrlclient.Client, name string) erro
 		},
 	}
 
+	return EnsureNamespace(ctx, c, ns)
+}
+
+// EnsureNamespace will create the namespace if it does not exist.
+func EnsureNamespace(ctx context.Context, c ctrlclient.Client, ns *corev1.Namespace) error {
+	if ns.TypeMeta.APIVersion == "" {
+		ns.TypeMeta.APIVersion = corev1.SchemeGroupVersion.String()
+	}
+	if ns.TypeMeta.Kind == "" {
+		ns.TypeMeta.Kind = "Namespace"
+	}
 	err := client.ServerSideApply(ctx, c, ns)
 	if err != nil {
 		return fmt.Errorf("failed to server side apply %w", err)
--- a/pkg/handlers/generic/lifecycle/utils/utils.go
+++ b/pkg/handlers/generic/lifecycle/utils/utils.go
@@ -93,7 +93,7 @@ func EnsureCRSForClusterFromObjects(
 	return nil
 }
 
-// EnsureNamespace will create the namespece if it does not exist.
+// EnsureNamespace will create the namespace if it does not exist.
 func EnsureNamespace(ctx context.Context, c ctrlclient.Client, name string) error {
 	ns := &corev1.Namespace{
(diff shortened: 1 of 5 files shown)


pkg/handlers/generic/lifecycle/utils/utils_suite_test.go
```diff
@@ -0,0 +1,16 @@
...
+package utils
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestAMIPatch(t *testing.T) {
```
dkoshkin, line 13: Did you forget to change the name?

dlipovetsky: Are you suggesting I copy-pasted this and forgot to change the name? How dare you! :joy:

dlipovetsky: Thanks for catching this. I fixed the function, and moved the integration tests to their own file to decrease confusion with the unit tests.

pkg/handlers/generic/lifecycle/utils/utils_integration_test.go
```diff
@@ -0,0 +1,62 @@
...
+		Expect(err).To(BeNil())
+
+		namespaceName := "new"
+
+		Expect(EnsureNamespace(ctx, c, namespaceName)).To(Succeed())
+		Expect(c.Delete(ctx, &corev1.Namespace{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: namespaceName,
+			},
+		})).To((Succeed()))
```
jimmidyson, line 28: Deleting namespaces in envtest does not work due to missing kube controllers. Let's remove this and ensure we use a new namespace name for each spec.

```suggestion
```

dlipovetsky: Thanks. I think you fixed this in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/559/commits/07543a8204a96403a35bf24bc10cd534642c5d26, which I'll append here,

dlipovetsky:
I've rebased on #556, added https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/559/commits/07543a8204a96403a35bf24bc10cd534642c5d26, and force-pushed

dlipovetsky:
#556 merged, and had an additional minor fixup, so I've rebased on main and force-pushed.

jimmidyson (approved):

pkg/handlers/generic/lifecycle/utils/utils.go
```diff
@@ -105,12 +105,18 @@ func EnsureNamespace(ctx context.Context, c ctrlclient.Client, name string) erro
...
 
-	err := client.ServerSideApply(ctx, c, ns, client.ForceOwnership)
+// EnsureNamespace will create the namespace if it does not exist.
+func EnsureNamespace(ctx context.Context, c ctrlclient.Client, ns *corev1.Namespace) error {
+	if ns.TypeMeta.APIVersion == "" {
+		ns.TypeMeta.APIVersion = corev1.SchemeGroupVersion.String()
+	}
+	if ns.TypeMeta.Kind == "" {
+		ns.TypeMeta.Kind = "Namespace"
+	}
```
dkoshkin, line 118: nit: we can use Go 1.22 magic :) 
```suggestion
	ns.TypeMeta.APIVersion = cmp.Or(ns.TypeMeta.APIVersion, corev1.SchemeGroupVersion.String())
	ns.TypeMeta.Kind = cmp.Or(ns.TypeMeta.APIVersion, "Namespace")
```

dkoshkin (approved):

pkg/handlers/generic/lifecycle/utils/utils.go
jimmidyson: Oh I like this!

dlipovetsky: I appreciate the suggestion. It is more concise, but I find it harder to understand, because you have to (a) know `cmp.Or` semantics, and (b) the zero values of the fields.

Apparently, `cmp.Or` [originated](https://blog.carlana.net/post/2024/golang-cmp-or-uses-and-history/) with a proposal for a `strings.First` function, which seems easier to understand.

PR #558, opened 2024-04-23
feat: add kube-vip static Pod in a Nutanix handler

dkoshkin:
Stacked on https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/554

This PR moves the responsibility of adding the control-plan kube-vip to CAREN away from infra providers.
This has a big benefit of being able to keep this implementation and the image version consistent across the different providers.

~Because CAPX is always requires kube-vip to be installed there are no API changes, but in the future we can add support to disable or change the provider.~
Added an empty `virtualIP{}` struct to Nutanix `controlPlaneEndpoint`.
```
spec:
  topology:
    variables:
      - name: clusterConfig
        value:
          nutanix:
            controlPlaneEndpoint:
              host: x.x.x.x
              port: 6443
              virtualIP: {}
```

Fixes https://jira.nutanix.com/browse/D2IQ-100364

New unit tests, and tested manually too:
```
clusterctl describe cluster $CLUSTER_NAME    
NAME                                                                                              READY  SEVERITY  REASON  SINCE  MESSAGE 
Cluster/dkkonvoy-042324-nutanix-03                                                                True                     76m             
├─ClusterInfrastructure - NutanixCluster/dkkonvoy-042324-nutanix-03-wvv6d                                                                  
├─ControlPlane - KubeadmControlPlane/dkkonvoy-042324-nutanix-03-sv2zb                             True                     76m             
│ └─Machine/dkkonvoy-042324-nutanix-03-sv2zb-ds5mv                                                True                     76m             
│   └─MachineInfrastructure - NutanixMachine/dkkonvoy-042324-nutanix-03-sv2zb-ds5mv                                                        
└─Workers                                                                                                                                  
  └─MachineDeployment/dkkonvoy-042324-nutanix-03-md-0-w6mls                                       True                     74m             
    └─Machine/dkkonvoy-042324-nutanix-03-md-0-w6mls-979mw-c582s                                   True                     74m             
      └─MachineInfrastructure - NutanixMachine/dkkonvoy-042324-nutanix-03-md-0-w6mls-979mw-c582s                                      
```


dkoshkin:
Tested manually too:
```
clusterctl describe cluster $CLUSTER_NAME    
NAME                                                                                              READY  SEVERITY  REASON  SINCE  MESSAGE 
Cluster/dkkonvoy-042324-nutanix-03                                                                True                     76m             
├─ClusterInfrastructure - NutanixCluster/dkkonvoy-042324-nutanix-03-wvv6d                                                                  
├─ControlPlane - KubeadmControlPlane/dkkonvoy-042324-nutanix-03-sv2zb                             True                     76m             
│ └─Machine/dkkonvoy-042324-nutanix-03-sv2zb-ds5mv                                                True                     76m             
│   └─MachineInfrastructure - NutanixMachine/dkkonvoy-042324-nutanix-03-sv2zb-ds5mv                                                        
└─Workers                                                                                                                                  
  └─MachineDeployment/dkkonvoy-042324-nutanix-03-md-0-w6mls                                       True                     74m             
    └─Machine/dkkonvoy-042324-nutanix-03-md-0-w6mls-979mw-c582s                                   True                     74m             
      └─MachineInfrastructure - NutanixMachine/dkkonvoy-042324-nutanix-03-md-0-w6mls-979mw-c582s                                      
```

pkg/common/controlplaneendpoint/virtualip/kubevip_test.go
```diff
@@ -0,0 +1,192 @@
...
+				configMapKey: client.ObjectKey{
+					Name:      "default-kube-vip-template",
+					Namespace: "default",
+				},
+			},
+		},
+	}
+
+	for idx := range tests {
+		tt := tests[idx] // Capture range variable
```
jimmidyson, line 132: No need to do this any more with go 1.22 🎉

jimmidyson, line 132: No need to do this any more with go 1.22 🎉 

```suggestion
	for _, tt := range tests {
```

dkoshkin: thanks, good to know!

dkoshkin:
> > Because CAPX always requires kube-vip to be installed
> 
> Is this true? Is there no way a customer could deploy a different load balancer for this? Should this be opt-in (but default true in our templates so effectively opt-out)?

We just got some feedback from an existing customer that is using an external LB where kube-vip would not be needed.
I will update this PR and put it behind an API that defaults to enabled and kube-vip provider.

hack/examples/bases/nutanix/clusterclass/kustomization.yaml.tmpl
```diff
@@ -30,12 +30,25 @@ patches:
...
       path: "/spec/template/spec/kubeadmConfigSpec/clusterConfiguration/apiServer/certSANs"
 
+# Delete the kube-vip file.
+# Will be templated and added back in the handler.
+- target:
+    kind: KubeadmControlPlaneTemplate
+  patch: |-
+    - op: test
+      path: /spec/template/spec/kubeadmConfigSpec/files/0/path
+      value: "/etc/kubernetes/manifests/kube-vip.yaml"
```
deepakm-ntnx, line 48: could you please elaborate on what does this test op do.

pkg/common/controlplaneendpoint/virtualip/fromreader.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	return &fromReaderProvider{
+		template: buf.String(),
+	}, nil
+}
+
+func (p *fromReaderProvider) GetFile(
+	_ context.Context,
+	spec v1alpha1.ControlPlaneEndpointSpec,
```
deepakm-ntnx, line 35: should we make this a kube-vip config structure with one of the value as CPEndpoint. we can set default values if needed but all the parameters are technically modifiable. sure capx does not support it yet but here we can allow to modify any parameters of kube-vip

deepakm-ntnx:
@dkoshkin did this work with k8s 1.29.x?

hack/examples/bases/nutanix/clusterclass/kustomization.yaml.tmpl
jimmidyson: https://jsonpatch.com/#test

dkoshkin:
> @dkoshkin did this work with k8s 1.29.x?

Good question, the pre/postKubeadmCommands are still in the templates so it should but let me try that.

pkg/common/controlplaneendpoint/virtualip/fromreader.go
dkoshkin: @deepakm-ntnx sorry I'm not sure if understand the question, can you please provide and example.

deepakm-ntnx: what i meant was technically all of the following env var which we set in the yaml are variables and user can change them based on need. we can make it all variables in capx as well (we already have a tracker). we can make it variables in here as well (could be with default values to start with) so that say user can update svc_enable etc
<pre>
            - name: vip_arp
              value: "true"
            - name: address
              value: "{{ `{{ .ControlPlaneEndpoint.Host }}` }}"
            - name: port
              value: "{{ `{{ .ControlPlaneEndpoint.Port }}` }}"
            - name: vip_cidr
              value: "32"
            - name: cp_enable
              value: "true"
            - name: cp_namespace
              value: kube-system
            - name: vip_ddns
              value: "false"
            - name: vip_leaderelection
              value: "true"
            - name: vip_leaseduration
              value: "15"
            - name: vip_renewdeadline
              value: "10"
            - name: vip_retryperiod
              value: "2"
            - name: svc_enable
              value: "false"
            - name: lb_enable
              value: "false"
            - name: enableServicesElection
              value: "false"
</pre>

deepakm-ntnx: https://jira.nutanix.com/browse/KRBN-8114

dkoshkin:
Tested with both Kubernetes v1.28
```
NAME                                                                                                 READY  SEVERITY  REASON  SINCE  MESSAGE                                                                                       
Cluster/dkkonvoy-042424-nutanix-128-2                                                                True                     31s                                                                                                   
├─ClusterInfrastructure - NutanixCluster/dkkonvoy-042424-nutanix-128-2-bc7z9                                                                                                                                                        
├─ControlPlane - KubeadmControlPlane/dkkonvoy-042424-nutanix-128-2-9mrkj                             True                     31s                                                                                                   
│ └─3 Machines...                                                                                    True                     3m22s  See dkkonvoy-042424-nutanix-128-2-9mrkj-hphdt, dkkonvoy-042424-nutanix-128-2-9mrkj-ktxjs, ...  
└─Workers                                                                                                                                                                                                                           
  └─MachineDeployment/dkkonvoy-042424-nutanix-128-2-md-0-mnm7c                                       True                     2m8s                                                                                                  
    └─Machine/dkkonvoy-042424-nutanix-128-2-md-0-mnm7c-qmwq8-bf7mx                                   True                     2m52s                                                                                                 
      └─MachineInfrastructure - NutanixMachine/dkkonvoy-042424-nutanix-128-2-md-0-mnm7c-qmwq8-bf7mx                                                                                                                                 


NAME                                                   CLUSTER                         NODENAME                                               PROVIDERID                                       PHASE     AGE     VERSION
dkkonvoy-042424-nutanix-128-2-9mrkj-hphdt              dkkonvoy-042424-nutanix-128-2   dkkonvoy-042424-nutanix-128-2-9mrkj-hphdt              nutanix://063a872a-fab1-4121-8692-c6a42cbaf75b   Running   71s     v1.28.7
dkkonvoy-042424-nutanix-128-2-9mrkj-ktxjs              dkkonvoy-042424-nutanix-128-2   dkkonvoy-042424-nutanix-128-2-9mrkj-ktxjs              nutanix://f1453c2b-4199-48e2-bc9b-80a312f34a98   Running   3m47s   v1.28.7
dkkonvoy-042424-nutanix-128-2-9mrkj-pm77z              dkkonvoy-042424-nutanix-128-2   dkkonvoy-042424-nutanix-128-2-9mrkj-pm77z              nutanix://af8d6dd7-8edb-4762-9c32-3de93c0ddb6f   Running   2m27s   v1.28.7
dkkonvoy-042424-nutanix-128-2-md-0-mnm7c-qmwq8-bf7mx   dkkonvoy-042424-nutanix-128-2   dkkonvoy-042424-nutanix-128-2-md-0-mnm7c-qmwq8-bf7mx   nutanix://23cb029f-f783-4c75-be6c-bd37c9bf8f6e   Running   3m50s   v1.28.7


      postKubeadmCommands:
      - echo export KUBECONFIG=/etc/kubernetes/admin.conf >> /root/.bashrc
      - echo "after kubeadm call" > /var/log/postkubeadm.log
      preKubeadmCommands:
      - echo "before kubeadm call" > /var/log/prekubeadm.log
      - hostnamectl set-hostname "{{ ds.meta_data.hostname }}"
      - echo "::1         ipv6-localhost ipv6-loopback" >/etc/hosts
      - echo "127.0.0.1   localhost" >>/etc/hosts
      - echo "127.0.0.1   kubernetes" >>/etc/hosts
      - echo "127.0.0.1   {{ ds.meta_data.hostname }}" >> /etc/hosts
      - /bin/bash /etc/cre/install-kubelet-credential-providers.sh
      - /bin/bash /etc/containerd/restart.sh
```

And v1.29
```
NAME                                                                                                 READY  SEVERITY  REASON  SINCE  MESSAGE                                                                                       
Cluster/dkkonvoy-042424-nutanix-129-7                                                                True                     8m43s                                                                                                 
├─ClusterInfrastructure - NutanixCluster/dkkonvoy-042424-nutanix-129-7-pwzt5                                                                                                                                                        
├─ControlPlane - KubeadmControlPlane/dkkonvoy-042424-nutanix-129-7-7mvfv                             True                     8m43s                                                                                                 
│ └─3 Machines...                                                                                    True                     10m    See dkkonvoy-042424-nutanix-129-7-7mvfv-2vj6x, dkkonvoy-042424-nutanix-129-7-7mvfv-b2g4p, ...  
└─Workers                                                                                                                                                                                                                           
  └─MachineDeployment/dkkonvoy-042424-nutanix-129-7-md-0-9wfdz                                       True                     9m16s                                                                                                 
    └─Machine/dkkonvoy-042424-nutanix-129-7-md-0-9wfdz-5mnjp-zf5rd                                   True                     10m                                                                                                   
      └─MachineInfrastructure - NutanixMachine/dkkonvoy-042424-nutanix-129-7-md-0-9wfdz-5mnjp-zf5rd                                                                                                                                 


NAME                                                   CLUSTER                         NODENAME                                               PROVIDERID                                       PHASE     AGE     VERSION
dkkonvoy-042424-nutanix-129-7-7mvfv-2vj6x              dkkonvoy-042424-nutanix-129-7   dkkonvoy-042424-nutanix-129-7-7mvfv-2vj6x              nutanix://ac70a906-76fe-4e5a-b10e-22253abc356d   Running   10m     v1.29.2
dkkonvoy-042424-nutanix-129-7-7mvfv-b2g4p              dkkonvoy-042424-nutanix-129-7   dkkonvoy-042424-nutanix-129-7-7mvfv-b2g4p              nutanix://57d8d55f-fb18-4e2d-9089-b76f1e79e433   Running   11m     v1.29.2
dkkonvoy-042424-nutanix-129-7-7mvfv-m52br              dkkonvoy-042424-nutanix-129-7   dkkonvoy-042424-nutanix-129-7-7mvfv-m52br              nutanix://98320226-0e9b-4c2c-82a8-b327ae8ce00f   Running   9m18s   v1.29.2
dkkonvoy-042424-nutanix-129-7-md-0-9wfdz-5mnjp-zf5rd   dkkonvoy-042424-nutanix-129-7   dkkonvoy-042424-nutanix-129-7-md-0-9wfdz-5mnjp-zf5rd   nutanix://62c16adb-780c-48e6-9473-646e016d2763   Running   12m     v1.29.2

      postKubeadmCommands:
      - echo export KUBECONFIG=/etc/kubernetes/admin.conf >> /root/.bashrc
      - echo "after kubeadm call" > /var/log/postkubeadm.log
      - |-
        if [ -f /run/kubeadm/kubeadm.yaml ]; then
          sed -i 's#path: /etc/kubernetes/super-admin.conf#path: /etc/kubernetes/admin.conf#' /etc/kubernetes/manifests/kube-vip.yaml;
        fi
      preKubeadmCommands:
      - echo "before kubeadm call" > /var/log/prekubeadm.log
      - hostnamectl set-hostname "{{ ds.meta_data.hostname }}"
      - echo "::1         ipv6-localhost ipv6-loopback" >/etc/hosts
      - echo "127.0.0.1   localhost" >>/etc/hosts
      - echo "127.0.0.1   kubernetes" >>/etc/hosts
      - echo "127.0.0.1   {{ ds.meta_data.hostname }}" >> /etc/hosts
      - |-
        if [ -f /run/kubeadm/kubeadm.yaml ]; then
          sed -i 's#path: /etc/kubernetes/admin.conf#path: /etc/kubernetes/super-admin.conf#' /etc/kubernetes/manifests/kube-vip.yaml;
        fi
      - /bin/bash /etc/cre/install-kubelet-credential-providers.sh
      - /bin/bash /etc/containerd/restart.sh
```

dkoshkin:
Thanks @deepakm-ntnx for the 1.29 question, it made me realize that CAREN should also be responsible for adding the commands, done in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/558/commits/a2a4ba7b509966256e3017d851ee31332bd4c577

I did test without this commit and it didn't work because it expects `KUBERNETES_VERSION` to be set to a value.
I will fix the CAPX templates to be more generic and instead only apply this patch based on the built in variable `.builtin.controlPlane.version`.

pkg/common/controlplaneendpoint/virtualip/fromreader.go
dkoshkin: Ahh got it. With https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/558/commits/c8c9c176801bdafafe13188382525fd5ae63d062 I've extended an API where in the future we can support additional configuration. IMO we should wait and only expose the configuration that we are ready to test/support.

For now though, its possible to override the full Pod spec by changing the ConfigMap. (similar pattern to the addons)

pkg/handlers/options/global.go
```diff
@@ -33,7 +35,7 @@ func (o *GlobalOptions) AddFlags(flags *pflag.FlagSet) {
 }
 
 func (o *GlobalOptions) DefaultsNamespace() string {
-	return o.defaultsNamespace
+	return cmp.Or(o.defaultsNamespace, corev1.NamespaceDefault)
```
dkoshkin, line 38: Defaulting here too because unit tests are now calling `NewGlobalOptions()` directly without calling  `AddFlags`.

charts/cluster-api-runtime-extensions-nutanix/templates/kube-vip/manifests/kube-vip-configmap.yaml
```diff
@@ -0,0 +1,74 @@
...
+    spec:
+      containers:
+        - name: kube-vip
+          image: ghcr.io/kube-vip/kube-vip:v0.6.4
+          imagePullPolicy: IfNotPresent
+          args:
+            - manager
+          env:
+            - name: vip_arp
+              value: "true"
```
jimmidyson, line 22: Why do we want to use ARP rather than BGP?

hack/examples/bases/nutanix/clusterclass/kustomization.yaml.tmpl
```diff
@@ -30,12 +30,36 @@ patches:
...
+# Delete the kube-vip related pre and postKubeadmCommands.
+# Will be added back in the handler if enabled.
+# If the index of these changes upstream this will need to change, but will show up as a git diff.
+- target:
+    kind: KubeadmControlPlaneTemplate
+  patch: |-
+    - op: "remove"
+      path: "/spec/template/spec/kubeadmConfigSpec/preKubeadmCommands/6"
+    - op: "remove"
+      path: "/spec/template/spec/kubeadmConfigSpec/postKubeadmCommands/1"
```
jimmidyson, line 61: Could we use JSONPatch tests here too?

charts/cluster-api-runtime-extensions-nutanix/templates/kube-vip/manifests/kube-vip-configmap.yaml
```diff
@@ -0,0 +1,74 @@
+# Copyright 2023 D2iQ, Inc. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+{{- if .Values.hooks.kubeVip.defaultTemplateConfigMap.create }}
+apiVersion: v1
+data:
+  kube-vip.yaml: |
```
jimmidyson, line 7: Can we generate the content of this using https://kube-vip.io/docs/installation/daemonset/#creating-the-manifest instead of just hard-coding them in here? Might make upgrades easier and hopefully not miss anything.

pkg/handlers/options/global.go
jimmidyson: Instead of this, we can set the default in `NewGlobalOptions` and use that property in the default for `defaults-namespace` flag.

hack/examples/bases/nutanix/clusterclass/kustomization.yaml.tmpl
jimmidyson: Forget that, one of the scripts is too long to check the value for.

charts/cluster-api-runtime-extensions-nutanix/templates/kube-vip/manifests/kube-vip-configmap.yaml
```diff
@@ -0,0 +1,74 @@
...
+              value: "false"
+            - name: lb_enable
+              value: "false"
+            - name: enableServicesElection
+              value: "false"
+          securityContext:
+            capabilities:
+              add:
+                - NET_ADMIN
+                - SYS_TIME
```
jimmidyson, line 53: What is this capability required for?

charts/cluster-api-runtime-extensions-nutanix/templates/kube-vip/manifests/kube-vip-configmap.yaml
jimmidyson: Pushed https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/562 to do that, let me know what you think.

pkg/handlers/options/global.go
jimmidyson: Done in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/562

charts/cluster-api-runtime-extensions-nutanix/templates/kube-vip/manifests/kube-vip-configmap.yaml
jimmidyson: Don't think this is required and is in fact not included when generating the manifests with kube-vip now so i think this is safe to remove.

dkoshkin: Thanks for looking into this 🙏

dkoshkin:
Tested with the latest changes after rebase

```
NAME                                                                                                 READY  SEVERITY  REASON  SINCE  MESSAGE                                                                                       
Cluster/dkkonvoy-042524-nutanix-128-1                                                                True                     36s                                                                                                   
├─ClusterInfrastructure - NutanixCluster/dkkonvoy-042524-nutanix-128-1-w9grt                                                                                                                                                        
├─ControlPlane - KubeadmControlPlane/dkkonvoy-042524-nutanix-128-1-pxzp4                             True                     36s                                                                                                   
│ └─3 Machines...                                                                                    True                     2m     See dkkonvoy-042524-nutanix-128-1-pxzp4-jpbkp, dkkonvoy-042524-nutanix-128-1-pxzp4-k86kz, ...  
└─Workers                                                                                                                                                                                                                           
  └─MachineDeployment/dkkonvoy-042524-nutanix-128-1-md-0-fwksc                                       True                     106s                                                                                                  
    └─Machine/dkkonvoy-042524-nutanix-128-1-md-0-fwksc-np89s-kb94k                                   True                     2m1s                                                                                                  
      └─MachineInfrastructure - NutanixMachine/dkkonvoy-042524-nutanix-128-1-md-0-fwksc-np89s-kb94k                                                                                                                                 
```

pkg/common/controlplaneendpoint/virtualip/fromreader.go
jimmidyson: Thinking about this, I wonder if this is better to add kube-vip as an addon API rather than under controlplane here? This would allow for configuring kube-vip service LB, etc rather than just saying it's for CP.

charts/cluster-api-runtime-extensions-nutanix/defaultclusterclasses/nutanix-cluster-class.yaml
```diff
@@ -212,14 +147,6 @@ spec:
...
         - echo export KUBECONFIG=/etc/kubernetes/admin.conf >> /root/.bashrc
-        - |
-          KUBERNETES_VERSION_NO_V=${KUBERNETES_VERSION#v}
-          VERSION_TO_COMPARE=1.29.0
-          if [ "$(printf '%s\n' "$KUBERNETES_VERSION_NO_V" "$VERSION_TO_COMPARE" | sort -V | head -n1)" != "$KUBERNETES_VERSION_NO_V" ]; then
-            if [ -f /run/kubeadm/kubeadm.yaml ]; then
-              sed -i 's#path: /etc/kubernetes/super-admin.conf#path: /etc/kubernetes/admin.conf#' /etc/kubernetes/manifests/kube-vip.yaml;
-            fi
-          fi
         - echo "after kubeadm call" > /var/log/postkubeadm.log
```
faiq, line 150: Can we get rid of these?

api/v1alpha1/common_types.go
```diff
@@ -37,4 +37,10 @@ type ControlPlaneEndpointSpec struct {
 	// +kubebuilder:validation:Minimum=1
 	// +kubebuilder:validation:Maximum=65535
 	Port int32 `json:"port"`
+
+	// Configuration for the virtual IP provider.
+	// +optional
+	VirtualIPSpec *ControlPlaneVirtualIPSpec `json:"virtualIP,omitempty"`
 }
+
+type ControlPlaneVirtualIPSpec struct{}
```
faiq, line 46: Can we put information like provider name or anything like that?

hack/addons/update-kube-vip-manifests.sh
```diff
@@ -0,0 +1,51 @@
...
+  --address='{{ `{{ .ControlPlaneEndpoint.Host }}` }}' \
+  --port=-99999 \
+  --controlplane \
+  --leaderElection \
+  --leaseDuration=15 \
+  --leaseRenewDuration=10 \
+  --leaseRetry=2 \
+  --prometheusHTTPServer='' |
+  gojq --yaml-input --yaml-output 'del(.metadata.creationTimestamp, .status) | .spec.containers[].imagePullPolicy |= "IfNotPresent"' |
+  sed "s/\"-99999\"/'{{ \`{{ .ControlPlaneEndpoint.Port }}\` }}'/" >"${ASSETS_DIR}/${FILE_NAME}"
```
faiq, line 34: Nice work

dkoshkin: Thanks @jimmidyson 🙏

api/v1alpha1/common_types.go
dkoshkin: I made it similar to some of the [other addons ](https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/808dae79f2b018999371ad4cb061b8c77bb3fd62/api/v1alpha1/addon_types.go#L52C6-L56) where if there is only a single option we don't expose it. I think we can keep it simple and have an empty struct, wdyt?

pkg/handlers/generic/mutation/controlplanevirtualip/providers/kubevip_test.go
```diff
@@ -0,0 +1,192 @@
...
+				client.ObjectKeyFromObject(tt.configMap),
+			)
+			require.Equal(t, tt.expectedErr, err)
+			assert.Equal(t, tt.expectedData, data)
+		})
+	}
+}
+
+var (
+	validKubeVIPTemplate = `
```
faiq, line 149: Lets put this template in the configMap as well.

api/v1alpha1/common_types.go
faiq: It's just not that clear to me its going to be kube-vip to me so maybe we just say `ControlPlaneVirtualIPSpec.providerName` = kube-vip`?

pkg/handlers/generic/mutation/controlplanevirtualip/inject.go
```diff
@@ -0,0 +1,192 @@
...
+	if controlPlaneEndpointVar.VirtualIPSpec == nil {
+		log.V(5).Info("ControlPlane VirtualIP not set")
+		return nil
+	}
+
+	cluster, err := clusterGetter(ctx)
+	if err != nil {
+		log.Error(
+			err,
+			"failed to get cluster from extraAPIServerCertSANs mutation handler",
```
supershal, line 115:
```suggestion
			"failed to get cluster from ControlPlaneVirtualIP mutation handler",
```

api/v1alpha1/common_types.go
dkoshkin: Ok I see your point, added it in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/558/commits/8f86f9280cc5dcea503371891e1b6c90e486b86a, thanks for the suggestion.

api/v1alpha1/common_types.go
```diff
@@ -37,4 +39,16 @@ type ControlPlaneEndpointSpec struct {
 	// +kubebuilder:validation:Minimum=1
 	// +kubebuilder:validation:Maximum=65535
 	Port int32 `json:"port"`
+
+	// Configuration for the virtual IP provider.
+	// +optional
+	VirtualIPSpec *ControlPlaneVirtualIPSpec `json:"virtualIP,omitempty"`
```
supershal, line 45: wdyt if we use the `*corev1.LocalObjectReference ` object reference here and default it to the kubevip configmap reference that we use in the chart? I think it could make the schema generic and the handler can use virtual ip specific logic at runtime. 
This way we dont have to provide name of the configmap as arg to CAREN server.

PS: I understand you just changed the schema to include `Provider` string.

dkoshkin: I thought about this at first but wanted to keep it similar to the addons.
If its part of the API it would be pretty easy to end up with different clusters with different config over a period of time since it would no longer come from a global config.

If we do prefer that approach though, IMO it should be a more general change across the whole API.

pkg/common/controlplaneendpoint/virtualip/fromreader.go
supershal: kubevip pod has to start along with first control plane for capi controllers to connect with api server. 
configuring kube-vip as service LB is a good idea, but i am not sure if we should use the same kube-vip pods for CP.

api/v1alpha1/common_types.go
supershal: ok. So we want that a CAREN instance can only support one VirtualIP Provider and all the clusters will use the same provider. The CAREN client has to make sure that they use correct value for `Provider` Enum that matches with what configmap they provided when deploying CAREN. Is that well established contract?

dkoshkin: Ahh I think I understand where the original question is coming from. When we support additional providers there would be [more flags](https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/558/files#diff-7a859c2cff68db150a13157650eadaa81d36ff680116ed54d910acad6fe8ab6eR62-R63) and charts/cluster-api-runtime-extensions-nutanix/templates/kube-vip/manifests/kube-vip-configmap.yaml deployed for the new providers.
So the user would have an option on which provider to deploy for different clusters, but what the config for that specific provider will be the same across all clusters.

dkoshkin: Let me change up the chart flags to make that a little more clear.

supershal: Thanks for making it clear. Flags for provider would work.

supershal (approved):

PR #559, opened 2024-04-24
feat: Preserve user-managed fields when applying resources

jimmidyson:
Added to prove (to myself more than anything!) that using SSA would properly preserve values.

Depends on #557.


dlipovetsky:
Thanks for verifying the SSA behavior. I pulled your changes into #557

PR #564, opened 2024-04-26
feat: AWS CCM for Kubernetes v1.29

dkoshkin:
Adds AWS CCM support for Kubernetes v1.29. While testing I also found an upstream change https://github.com/kubernetes/kubernetes/pull/121028 and to fix that added `hostNetwork: true` to the CCM pods. Upstream CAPA templates and the Nutanix CCM already use `hostNetwork: true`, so it made sense to go with that approach here too.

Fixes #

Created an AWS v1.29.4 cluster with the following lookup envs:
```
AMI_LOOKUP_BASEOS=rocky-9.1
AMI_LOOKUP_FORMAT=konvoy-ami-{{.BaseOS}}-release-?{{.K8sVersion}}-*
AMI_LOOKUP_ORG=999867407951
```


faiq (approved):

hack/addons/kustomize/aws-ccm/helm-values.yaml
```diff
@@ -0,0 +1,12 @@
+# Copyright 2024 D2iQ, Inc. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+---
+# Starting in Kubernetes v1.29 the Kubelet no longer adds temporary addresses to the Node.
```
supershal, line 5: thanks for the comment

supershal (approved):

supershal (approved):

PR #588, opened 2024-04-29
refactor: Clean up API constants, and explain usage

dlipovetsky:
We have not explained how we use API constants. We also have good reasons to remove some constants. This PR does the former, and gets us closer to the latter.

Fixes #


jimmidyson (approved):
Thanks!

supershal (approved):
Thank you for the code comments.

PR #590, opened 2024-04-29
build: fix outdated devbox.lock file

dkoshkin:
The release failed because of a dirty `git` state with `devbox.lock`. https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/actions/runs/8882515772/job/24387382324

Fixes #

### DIFF
--- a/devbox.lock
+++ b/devbox.lock
@@ -41,26 +41,6 @@
         }
       }
     },
-    "clusterctl@latest": {
-      "last_modified": "2024-03-15T20:27:35Z",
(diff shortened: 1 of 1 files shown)


jimmidyson (approved):
Sorry this was my bad, I didn't change the lock file because some people had problems with this new format and didn't think about it causing failed releases. But we should all be on new enough devbox now to handle this so let's get it merged.

mhrabovcin (approved):

PR #598, opened 2024-04-30
fix: Fix error messages returned by HelmChartGetter

dlipovetsky:
Puts the namespace and name in the right order in a log message, and includes these in the corresponding error message, instead of a data value that's not human-readable.

Fixes #

### DIFF
--- a/pkg/handlers/generic/lifecycle/config/cm.go
+++ b/pkg/handlers/generic/lifecycle/config/cm.go
@@ -76,10 +76,10 @@ func (h *HelmChartGetter) For(
 	name Component,
 ) (*HelmChart, error) {
 	log.Info(
-		fmt.Sprintf("Fetching HelmChart info for %s from configmap %s/%s",
+		fmt.Sprintf("Fetching HelmChart info for %q from configmap %s/%s",
 			string(name),
-			h.cmName,
-			h.cmNamespace),
+			h.cmNamespace,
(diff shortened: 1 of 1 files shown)


dkoshkin (approved):
🙇

faiq (approved):

pkg/handlers/generic/lifecycle/config/cm.go
```diff
@@ -76,18 +76,18 @@ func (h *HelmChartGetter) For(
 	name Component,
 ) (*HelmChart, error) {
 	log.Info(
-		fmt.Sprintf("Fetching HelmChart info for %s from configmap %s/%s",
+		fmt.Sprintf("Fetching HelmChart info for %q from configmap %s/%s",
 			string(name),
-			h.cmName,
-			h.cmNamespace),
+			h.cmNamespace,
+			h.cmName),
```
jimmidyson, line 82: Really nitty but interesting all the same imo: I'm starting to prefer creating `client.ObjectKey` and relying on the stringifying of that rather than relying on ordering like this - I've seen it mixed up so many times. Also no need to cast `name` to a string.

```suggestion
		fmt.Sprintf(
			"Fetching HelmChart info for %q from configmap %s",
			name,
			client.ObectKey{Namespace: h.cmNamespace, h.cmName},
		),
```

PR #592, opened 2024-04-29
feat: Add the ServiceLoadbalancer Addon, with MetalLB as first provider

dlipovetsky:
**What problem does this PR solve?**:
Adds the `ServiceLoadbalancer` addon, and support for MetalLB as the first provider.

### TODO:
- [x] API type
- [x] Generic handler
- [x] MetalLB handler
- [x] Deploy MetalLB Helm chart 
- [x] Default configuration for MetalLB Helm chart
- [ ] Document the ServiceLoadbalancer Addon.

**Which issue(s) this PR fixes**:
Fixes https://jira.nutanix.com/browse/D2IQ-100444

**How Has This Been Tested?**:

**Special notes for your reviewer**:

PR #603, opened 2024-05-01
fix: Correct the CSI handler logic

dlipovetsky:
- Return an error when the CSI variable is defined, but the
  Providers list is empty.
- Return an error when the CSI provider is defined, but its
  StorageClassConfig list is empty.
- Return an error when any CSI provider is unknown.
- Return an error when DefaultStorage is empty.
- Return an error when the DefaultStorage Provider Name does not match a set provider.
- Return an error when the DefaultStorage StorageClassConfig Name does not match the name of a StorageClassConfig for the default provider.   

Adds unit tests to verify this logic, and more. Covers the majority of the generic CSI handler `AfterControlPlaneInitialized` hook.

Fixes #

Adds a unit test; I think this is the first unit test in CAREN to exercise a handler request-response flow.
 
I noticed these issues as I was studying the CSI handler in order to write the ServiceLoadbalancer handler. I'll be adding a unit test to #592 as well.


pkg/handlers/generic/lifecycle/csi/handler.go
```diff
@@ -74,68 +74,103 @@ func (c *CSIHandler) AfterControlPlaneInitialized(
...
 		return
 	}
-	if len(csiProviders.Providers) == 1 &&
-		len(csiProviders.Providers[0].StorageClassConfig) == 1 &&
-		csiProviders.DefaultStorage == nil {
-		csiProviders.DefaultStorage = &v1alpha1.DefaultStorage{
-			ProviderName:           csiProviders.Providers[0].Name,
-			StorageClassConfigName: csiProviders.Providers[0].StorageClassConfig[0].Name,
+
+	if len(csi.Providers) == 0 {
```
jimmidyson, line 99: This can be added as `minItems` in the API schema can't it? I don't think we need custom logic for this. Happy to leave this here as well as defensive programming but let's also update the API to prevent bad data getting in.

pkg/handlers/generic/lifecycle/csi/handler.go
```diff
@@ -74,68 +74,103 @@ func (c *CSIHandler) AfterControlPlaneInitialized(
...
+			err,
+			"The list of CSI providers must include at least one provider.",
+		)
+		resp.SetStatus(runtimehooksv1.ResponseStatusFailure)
+		resp.SetMessage("the list of CSI providers must include at least one provider")
+		return
+	}
+
+	for _, provider := range csi.Providers {
+		if len(provider.StorageClassConfig) == 0 {
```
jimmidyson, line 110: Again, we can prevent this happening via OpenAPI validation using `minItems`.

pkg/handlers/generic/lifecycle/csi/handler.go
```diff
@@ -74,68 +74,103 @@ func (c *CSIHandler) AfterControlPlaneInitialized(
...
+			)
+			return
+		}
+	}
+
+	if csi.DefaultStorage == nil {
+		if len(csi.Providers) > 1 {
+			log.Error(
+				err,
+				"A CSI configuration with two or more providers must configure the default storage.",
```
jimmidyson, line 130: Why must it do this? Couldn't a user configure multiple CSI providers, not set a default, and explicitly add required storage class to PVCs?

jimmidyson:
Pushed a PR targeting your PR branch 😅 https://github.com/dlipovetsky/capi-runtime-extensions/pull/1

pkg/handlers/generic/lifecycle/csi/handler.go
dlipovetsky: IIUC, Kommander uses only the default storage class.

pkg/handlers/generic/lifecycle/csi/handler.go
dlipovetsky: Yes, I meant to add that, but got distracted by the test :smile:

pkg/handlers/generic/lifecycle/csi/handler.go
jimmidyson: It does, but that can be set in our usage of CAREN in NKP, not necessary to enforce that in this project i don't think.

dlipovetsky:
> Pushed a PR targeting your PR branch 😅 [dlipovetsky#1](https://github.com/dlipovetsky/capi-runtime-extensions/pull/1)

Thank you. Merged!

pkg/handlers/generic/lifecycle/csi/handler.go
dlipovetsky: I remembered the main reason I added this logic.

Right now, we want the default storage to be defined, but out of convenience, we derive it, if we can.

That is, if the user does not define default storage, then:

- 1 provider: we derive the default storage 
- 2 providers: we cannot derive the default storage, so we return an error

You're proposing we change this to:

- 1 provider: we derive the default storage
- 2 providers: we leave the default storage undefined

I think this is inconsistent. I don't think we should derive the default storage in the first case, but not try to derive it in the second.

pkg/handlers/generic/lifecycle/csi/handler.go
dlipovetsky: Let's use validation and keep this.

pkg/handlers/generic/lifecycle/csi/handler.go
dlipovetsky: Let's use validation and keep this.

pkg/handlers/generic/lifecycle/csi/handler.go
jimmidyson: Good analysis! I feel we should require explicit configuration of default storage. This would enable users to configure a different CSI provider that we don't support in addition to one we do and make that default storage. I know this may sound like additional burden on the user, but an explicit opt-in UX is clearest here i think. We can make that easier with our cluster create clients in our product, while still allowing flexibility here.

faiq: In practical terms, does this mean dropping the behavior of defaulting at all?

dlipovetsky: I made the default storage a required field.

dlipovetsky:
I rebased on main and force-pushed. The conflict was with kubebuilder tags, nothing related to these changes.

I added changes in a new commit, so previous reviews are still valid.

I added stricter validation for the default storage configuration as well.

The unit test covers all of the handler code, with the exception of two error cases, one for failing to read the CSI variable, and another for failing to apply the driver. I'll see about handling these as well.

dlipovetsky:
I added a test to confirm we return an error on handler failure. I'm not going to add a test for failing to read the CSI variable for now.

dkoshkin (commented):
Thanks for the changes, the unit tests are really helpful to understand these complicated API interactions.

dkoshkin (approved):

pkg/handlers/generic/lifecycle/csi/handler.go
```diff
@@ -74,68 +74,118 @@ func (c *CSIHandler) AfterControlPlaneInitialized(
...
 		return
 	}
-	if len(csiProviders.Providers) == 1 &&
-		len(csiProviders.Providers[0].StorageClassConfig) == 1 &&
-		csiProviders.DefaultStorage == nil {
-		csiProviders.DefaultStorage = &v1alpha1.DefaultStorage{
-			ProviderName:           csiProviders.Providers[0].Name,
-			StorageClassConfigName: csiProviders.Providers[0].StorageClassConfig[0].Name,
+
+	// This is defensive, because the API validation requires at least provider.
```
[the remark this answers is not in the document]
dlipovetsky: Thanks, fixed!

pkg/handlers/generic/lifecycle/csi/handler.go
```diff
@@ -74,68 +74,117 @@ func (c *CSIHandler) AfterControlPlaneInitialized(
...
-				fmt.Sprintf(
-					"Skipping CSI handler, for provider given in %s. Provider handler not given.",
-					provider.Name,
-				),
+			log.Error(
+				err,
+				"CSI provider is unknown",
+				"name",
+				provider.Name,
 			)
```
faiq, line 168: a typo could lead to a caren cluster not getting the rest of the things deployed. i left this not as an error explicitly,

api/variables/util.go
```diff
@@ -0,0 +1,32 @@
+// Copyright 2024 D2iQ, Inc. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package variables
```
dkoshkin, line 4: WDYT moving these into https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/main/common/pkg/capi/clustertopology/variables/variable.go, seems like that package is already doing something similar related to `clusterv1.ClusterVariable`

pkg/handlers/generic/lifecycle/csi/handler.go
```diff
@@ -74,68 +74,117 @@ func (c *CSIHandler) AfterControlPlaneInitialized(
...
+				"name",
+				provider.Name,
+			)
+			resp.SetStatus(runtimehooksv1.ResponseStatusFailure)
+			resp.SetMessage(
+				fmt.Sprintf("the CSI provider %q must configure at least one storage class",
+					provider.Name),
+			)
+			return
+		}
```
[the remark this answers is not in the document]
dlipovetsky: I think we can delegate to the client (CLI/UI). The client will create at least one storage class, so the user won't need to do additional work.

Is there a good reason to not have a storage class? If there is, then we can only require a storage class to satisfy a reference in DefaultStorage.

pkg/handlers/generic/lifecycle/csi/handler.go
dlipovetsky: You're right; since this hook runs after the control plane is initialized, this would block the cluster from deploying other addons.

However, once the user remedies the issue, the hook should succeed, and unblock. I think that's the UX we want.

pkg/handlers/generic/lifecycle/csi/handler.go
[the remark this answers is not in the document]
faiq: Sorry I might be misreading this, but it looks like we need to have a storage class for every CSI provider. I have a stronger disagreement with that, because just deploying a CSI provider seems like something a user could want to do.

api/variables/util.go
dlipovetsky: The rest of the code in this package appears to be dead. I'll check about removing it.

dlipovetsky: Dead code removed in #623

pkg/handlers/generic/lifecycle/csi/handler.go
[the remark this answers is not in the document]
dlipovetsky: That's a good point!

How about we require only the storage class that is referenced by `DefaultStorage.StorageClassConfigName`? Since we require `DefaultStorage`, we create at least one storage class on the cluster, and make it the default one. Additional storage classes are optional.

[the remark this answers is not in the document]
dlipovetsky: Done in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/603/commits/e53524694936c0d242e3cb06c6164b4eef84dbd3

pkg/handlers/generic/lifecycle/csi/handler.go
```diff
@@ -27,7 +27,7 @@ type CSIProvider interface {
 	Apply(
 		context.Context,
 		v1alpha1.CSIProvider,
-		*v1alpha1.DefaultStorage,
+		v1alpha1.DefaultStorage,
```
faiq, line 30: Why not a pointer anymore?

pkg/handlers/generic/lifecycle/csi/handler.go
```diff
@@ -74,68 +74,107 @@ func (c *CSIHandler) AfterControlPlaneInitialized(
...
+			"The list of CSI providers must include at least one provider.",
+		)
+		resp.SetStatus(runtimehooksv1.ResponseStatusFailure)
+		resp.SetMessage("the list of CSI providers must include at least one provider")
+		return
+	}
+
+	// Verify that the default storage references a defined provider, and one of the
+	// storage class configs for that provider.
+	{
```
faiq, line 112: why is this in a block?

dlipovetsky: This logic is 30+ lines, so I put it in its own block to be easier to read. I considered putting it in a separate function, too.

pkg/handlers/generic/lifecycle/csi/handler.go
dlipovetsky: This change makes DefaultStorage a required field. It can't be nil, and so there's no need to represent that possibility.

dlipovetsky:
@faiq noticed some misleading `log.Error` calls. Thanks for catching that! Fixed in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/603/commits/c4fa6746612d31fa75698975b27e8f794540309b

faiq (approved):
looks great. thanks for taking in the feedback

pkg/handlers/generic/lifecycle/csi/handler.go
dlipovetsky: At your suggestion, factored this out to its own function in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/603/commits/e0f134ddb164c09fbdf12f9cbc721737eb5e0f8d. Thanks!

PR #604, opened 2024-05-02
refactor: Apply kubebuilder annotations for required/optional everywhere

jimmidyson:
Following advice from an upstream PR to explicitly mark fields as optional/required
to ensure there are no surprises/bugs that arise through tooling.

Also, add API conventions doc to contributing section for reference.


api/v1alpha1/addon_types.go
```diff
@@ -94,69 +97,79 @@ type NFD struct {
...
 	DefaultStorage *DefaultStorage `json:"defaultStorage,omitempty"`
 }
 
 type CSIProvider struct {
 	// Name of the CSI Provider.
+	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:Enum=aws-ebs;nutanix
 	Name string `json:"name"`
 
 	// +optional
```
dkoshkin, line 131:
```suggestion
	// +kubebuilder:validation:Optional
```

api/v1alpha1/clusterconfig_types.go
```diff
@@ -147,107 +146,112 @@ type NutanixClusterConfigSpec struct {
...
 }
 
 type ImageRegistry struct {
-	// Registry URL.
 	// +kubebuilder:validation:Format=`uri`
 	// +kubebuilder:validation:Pattern=`^https?://`
+	// +kubebuilder:validation:Required
+
+	// Registry URL.
 	URL string `json:"url"`
```
dkoshkin, line 224:
```suggestion
	// Registry URL.
	// +kubebuilder:validation:Format=`uri`
	// +kubebuilder:validation:Pattern=`^https?://`
	// +kubebuilder:validation:Required
	URL string `json:"url"`
```

api/v1alpha1/clusterconfig_types.go
```diff
@@ -147,107 +146,112 @@ type NutanixClusterConfigSpec struct {
...
+	// +kubebuilder:validation:Optional
 	Credentials *RegistryCredentials `json:"credentials,omitempty"`
 }
 
 // User defines the input for a generated user in cloud-init.
 type User struct {
+	// +kubebuilder:validation:Required
+
 	// Name specifies the user name.
 	Name string `json:"name"`
```
dkoshkin, line 236:
```suggestion
	// Name specifies the user name.
	// +kubebuilder:validation:Required
	Name string `json:"name"`
```

api/v1alpha1/common_types.go
```diff
@@ -13,36 +13,48 @@ type ObjectMeta struct {
...
 	// More info: http://kubernetes.io/docs/user-guide/annotations
-	// +optional
+	// +kubebuilder:validation:Optional
 	Annotations map[string]string `json:"annotations,omitempty"`
 }
 
 type ControlPlaneEndpointSpec struct {
 	// The hostname on which the API server is serving.
+	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:MinLength=1
```
dkoshkin, line 30:
```suggestion
	// The hostname on which the API server is serving.
	// +kubebuilder:validation:MinLength=1
	// +kubebuilder:validation:Required
```

api/v1alpha1/common_types.go
```diff
@@ -13,36 +13,48 @@ type ObjectMeta struct {
...
 	Host string `json:"host"`
 
 	// The port on which the API server is serving.
+	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:Minimum=1
 	// +kubebuilder:validation:Maximum=65535
```
dkoshkin, line 36: nit: to be consistent with the rest of the API
```suggestion
	// +kubebuilder:validation:Minimum=1
	// +kubebuilder:validation:Maximum=65535
	// +kubebuilder:validation:Required
```

api/v1alpha1/common_types.go
```diff
@@ -13,36 +13,48 @@ type ObjectMeta struct {
...
 	Provider string `json:"provider,omitempty"`
 }
+
+// LocalObjectReference contains enough information to let you locate the
+// referenced object inside the same namespace.
+type LocalObjectReference struct {
+	// Name of the referent.
+	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
```
dkoshkin, line 58:
```suggestion
	// Name of the referent.
	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
	// +kubebuilder:validation:MinLength=1
	// +kubebuilder:validation:Required
```

api/v1alpha1/nutanix_clusterconfig_types.go
```diff
@@ -19,31 +17,35 @@ const (
...
 	// Nutanix Prism Central endpoint configuration.
+	// +kubebuilder:validation:Required
 	PrismCentralEndpoint NutanixPrismCentralEndpointSpec `json:"prismCentralEndpoint"`
 }
 
 type NutanixPrismCentralEndpointSpec struct {
 	// The URL of Nutanix Prism Central, can be DNS name or an IP address.
+	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:Format=`uri`
 	// +kubebuilder:validation:Pattern=`^https://`
```
dkoshkin, line 32:
```suggestion
	// The URL of Nutanix Prism Central, can be DNS name or an IP address.
	// +kubebuilder:validation:Format=`uri`
	// +kubebuilder:validation:Pattern=`^https://`
	// +kubebuilder:validation:Required
```

api/v1alpha1/nutanix_clusterconfig_types.go
```diff
@@ -19,31 +17,35 @@ const (
...
 	Insecure bool `json:"insecure"`
 
 	// A base64 PEM encoded x509 cert for the RootCA that was used to create
 	// the certificate for a Prism Central that uses certificates that were issued by a non-publicly trusted RootCA.
 	// The trust bundle is added to the cert pool used to authenticate the TLS connection to the Prism Central.
 	// +kubebuilder:validation:Format=`byte`
-	// +optional
-	AdditionalTrustBundle *string `json:"additionalTrustBundle,omitempty"`
+	// +kubebuilder:validation:Optional
+	AdditionalTrustBundle string `json:"additionalTrustBundle,omitempty"`
```
dkoshkin, line 44: 👍

api/v1alpha1/nutanix_node_types.go
```diff
@@ -15,50 +15,57 @@ type NutanixNodeSpec struct {
...
 	// exist in Prism Central. One category key can have more than one value.
-	// +optional
+	// +kubebuilder:validation:Optional
 	AdditionalCategories []NutanixCategoryIdentifier `json:"additionalCategories,omitempty"`
 
 	// Defines the boot type of the virtual machine. Only supports UEFI and Legacy
-	BootType NutanixBootType `json:"bootType,omitempty"`
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Enum:=legacy;uefi
+	BootType NutanixBootType `json:"bootType"`
```
dkoshkin, line 52:
```suggestion
	// +kubebuilder:validation:Enum:=legacy;uefi
	// +kubebuilder:validation:Optional
	BootType NutanixBootType `json:"bootType"`
```

api/v1alpha1/common_types.go
jimmidyson: Why do you want it this way round? Alphabetically ordered? I'd prefer to have Required/Optional first and validation after. I can go through and make that consistent if you agree.

api/v1alpha1/clusterconfig_types.go
jimmidyson: Done.

api/v1alpha1/common_types.go
jimmidyson: I made required/optional consistently first.

api/v1alpha1/addon_types.go
```diff
@@ -94,69 +97,79 @@ type NFD struct {
...
 
 type StorageClassConfig struct {
 	// Name of storage class config.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
 	Name string `json:"name"`
 
 	// Parameters passed into the storage class object.
-	// +optional
+	// +kubebuilder:validation:Optional
```
deepakm-ntnx, line 151: Per https://book.kubebuilder.io/reference/markers.html
<pre>
If you’re using controller-gen only then they’re redundant, but if you’re using other generators or you want developers that need to build their own clients for your API, you’ll want to also include +optional.
</pre>
do we expect anyone building their own clients? ofcourse following also says
<pre>
The most reliable way in 1.x to get +optional is omitempty.
</pre>
which we do.

api/v1alpha1/addon_types.go
```diff
@@ -94,69 +97,79 @@ type NFD struct {
...
 	Parameters map[string]string `json:"parameters,omitempty"`
 
+	// +kubebuilder:validation:Optional
 	// +kubebuilder:validation:Enum=Delete;Retain;Recycle
 	// +kubebuilder:default=Delete
-	// +optional
 	ReclaimPolicy corev1.PersistentVolumeReclaimPolicy `json:"reclaimPolicy,omitempty"`
```
deepakm-ntnx, line 157: should this be pointer? per https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required
<pre>
Optional fields have the following properties:

They have the +optional comment tag in Go.
They are a pointer type in the Go definition (e.g. AwesomeFlag *SomeFlag) or have a built-in nil value (e.g. maps and slices).
The API server should allow POSTing and PUTing a resource with this field unset.
</pre>

api/v1alpha1/addon_types.go
```diff
@@ -94,69 +97,79 @@ type NFD struct {
...
 
 	// +kubebuilder:validation:Enum=Immediate;WaitForFirstConsumer
+	// +kubebuilder:validation:Optional
 	// +kubebuilder:default=WaitForFirstConsumer
-	// +optional
 	VolumeBindingMode storagev1.VolumeBindingMode `json:"volumeBindingMode,omitempty"`
```
deepakm-ntnx, line 162: same as above. should be pointer?

api/v1alpha1/aws_clusterconfig_types.go
```diff
@@ -9,28 +9,32 @@ import (
...
 type Region string
 
 type AWSNetwork struct {
-	// +optional
+	// +kubebuilder:validation:Optional
 	VPC *VPC `json:"vpc,omitempty"`
 
 	// AWS Subnet configuration.
-	// +optional
+	// +kubebuilder:validation:Optional
```
deepakm-ntnx, line 30: Same as above, should be a pointer?

api/v1alpha1/addon_types.go
jimmidyson: I don't expect any clients to be built from this, especially as these are not CRDs to have a client built for.

> The most reliable way in 1.x to get +optional is omitempty.

We're on 3.x though 😉

api/v1alpha1/addon_types.go
jimmidyson: This is a slice (aliased type) so is effectively already a pointer.

api/v1alpha1/addon_types.go
jimmidyson: Good catch! Thanks. Updated.

api/v1alpha1/aws_clusterconfig_types.go
jimmidyson: This is a slice (aliased type) so is effectively already a pointer.

api/v1alpha1/addon_types.go
jimmidyson: I think it's ok to leave with `+kubebuilder:validation:Optional` as a companion of `+kubebuilder:validation:Required` here. Are you OK with that @deepakm-ntnx ? We can add `+optional` back later if we find we need to, but I don't expect we will.

deepakm-ntnx: sure makes sense. thanks

deepakm-ntnx (approved):

dkoshkin (approved):

PR #607, opened 2024-05-02
fix: Fix the internal ClusterConfig type used for provider-agnostic logic

jimmidyson:
Spotted in discussions with @dkoshkin about internal usage of types like
this as opposed to type switches.


dkoshkin (approved):
Thanks. I'll create a follow up PR with some additional changes to support clients unmarshaling these types.

faiq (approved):

PR #610, opened 2024-05-03
feat: API for encryption at-rest

supershal:
- CAREN API for encryption at-rest
```
topology:
    variables:
      - name: clusterConfig
        value:
          encryption:
            providers: 
              - aescbc: {}
```
- [x] Define API schema
- [x] Implement mutator handler
- [x] unit tests
- [x] e2e tests
- [x] documentation

Fixes # https://jira.nutanix.com/browse/D2IQ-100555


api/v1alpha1/clusterconfig_types.go
```diff
@@ -266,6 +269,18 @@ type User struct {
...
+// Encryption defines the configuration to enable encryption at REST
+// This configuration is used by API server to encrypt data before storing it in ETCD.
+// Currently the encryption only enabled for secrets and configmaps.
+type Encryption struct {
+	// Encryption providers
+	// +kubebuilder:validation:UniqueItems=true
+	// +kubebuilder:validation:Enum=aescbc;aesgcm
+	// +kubebuilder:default=aescbc
+	// +optional
+	Providers []string `json:"providers"`
```
supershal, line 281: The API supports multiple providers. https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/#understanding-the-encryption-at-rest-configuration
should we support multiple providers or single provider? 

The first provider in the list will be used for encryption.

jimmidyson: I think the only reason for multiple providers here is if a user wants to change providers, they will be tried in order for decryption, the first provider being used for encryption on write. I think that use case is quite advanced/unusual so I would advocate for only a single provider at this point. We can extend the API later if we need to support multiple. Wdyt?

api/v1alpha1/crds/caren.nutanix.com_awsclusterconfigs.yaml
```diff
@@ -298,6 +298,23 @@ spec:
...
+                description: |-
+                  Encryption defines the configuration to enable encryption at REST
+                  This configuration is used by API server to encrypt data before storing it in ETCD.
+                  Currently the encryption only enabled for secrets and configmaps.
+                properties:
+                  providers:
+                    default: aescbc
+                    description: Encryption providers
+                    enum:
+                    - aescbc
```
jimmidyson, line 311: Can we support kms V2 as well? Both these providers have issues that may not be acceptable to users.

api/v1alpha1/crds/caren.nutanix.com_awsclusterconfigs.yaml
```diff
@@ -298,6 +298,23 @@ spec:
...
+                    - aesgcm
+                    items:
+                      type: string
```
jimmidyson, line 314: Don't these providers require keys to be provided too?

supershal: in konvoy we generate the keys and store in secrets. Should we do the same here?

api/v1alpha1/crds/caren.nutanix.com_awsclusterconfigs.yaml
supershal: KMS V2 requires KMS configuration that we do not support at the moment. https://kubernetes.io/docs/tasks/administer-cluster/kms-provider/#encrypting-your-data-with-the-kms-provider-kms-v2

api/v1alpha1/clusterconfig_types.go
supershal: Agree. We can keep it limited to one provider. However I think we should keep type as array and add max length to 1 to allow adding more providers in future.

api/v1alpha1/crds/caren.nutanix.com_awsclusterconfigs.yaml
jimmidyson: Can we have a think how we can support it? Not a blocker for this pr but something we should support in future.

api/v1alpha1/clusterconfig_types.go
jimmidyson: Ok that sounds good to me.

api/v1alpha1/crds/caren.nutanix.com_awsclusterconfigs.yaml
dkoshkin: @supershal are you suggesting that clients(konvoy, kommander UI) need to create these Secrets?

supershal: @dkoshkin I am proposing that CAREN creates the Encryption key secrets and Create a secret that contains `EncryptionConfig` . Similar as how konvoy does it.

dkoshkin: 👍 SGTM, thanks for clarifying @supershal

supershal: > Ok let's do the same for first cut, but allow for potential extension later by making this an object with a single name field? Or a map from string to object (with no properties for now).

Let me try this out.

jimmidyson: I think this should be explicit in the API and it should be possible to specify an existing secret or secret name. We can always default to generate as long as it can be disabled.

api/v1alpha1/crds/caren.nutanix.com_awsclusterconfigs.yaml
supershal: Created https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/issues/624 to track it. We can add it to our JIRA backlog if it becomes requirement.

api/v1alpha1/crds/caren.nutanix.com_awsclusterconfigs.yaml
supershal: Both `aesgcm` and `aescbc` requires a secret key or a set of secret keys in terms of rotations. The KMS v2 would require configuration to support KMS plugin's socket location and timeouts and cache size.

A one time secret key (a unique slice of bytes) for `aesgcm` and `aescbc` can be generated from CAREN.
So far we do not allow passing secret in the CAREN API. Providing K8s Secret object's reference would make more sense if we decide to go that route. Then we have to come up with secret format and lifecycle of the secret, specially when we introduce rotating of the keys. 

Given the fact that
1) `aescbc` and `aesgcm` are not recommended because former is weak and latter requires rotation implemented as it must rotate key after 200k writes.
2) we have not implemented or planning to implement rotation of the keys. If we plan to implement rotation at some point, getting secret keys from user could become complex.
3) The recommended provider KMS V2 provides key rotation

Should we provide a way for user to create a secret and provide its reference in CAREN API?
In future, what KMS V2 configuration we would allow user to modify. (cache, socket path, timeouts) Or CAREN will handle it using some sane defaults in its internal configuration.

@jimmidyson we briefly discussed some of it in scrum. I will setup a huddle tomorrow so that we can finalize it.

api/v1alpha1/crds/caren.nutanix.com_awsclusterconfigs.yaml
```diff
@@ -298,6 +298,23 @@ spec:
...
+                  Encryption defines the configuration to enable encryption at REST
+                  This configuration is used by API server to encrypt data before storing it in ETCD.
+                  Currently the encryption only enabled for secrets and configmaps.
+                properties:
+                  providers:
+                    default: aescbc
+                    description: Encryption providers
+                    enum:
+                    - aescbc
+                    - aesgcm
```
supershal, line 312: @dlipovetsky @dkoshkin the result of the SPIKE recommends that we should also allow `aesgcm` as one of the provider and default provider. https://jira.nutanix.com/browse/D2IQ-100160?focusedId=6347378&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-6347378

The `aesgcm` requires rotation of the keys every 200k writes. https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/#providers
Given that we do not rotate keys, should we support it in CAREN API. 
If not, then it leaves only `aescbc` as a provider that we can support right now to support Konvoy usecase. 
We can keep the API flexible, I am wondering if we should allow `aesgcm` as one of the option without providing a way to rotate the keys before 200k writes.

dkoshkin: Good catch, IMO CAREN should not allow `aesgcm` since it would be too easy to have a vulnerable cluster over time, and should be secure by default. I think these kinds of details is another reason for CAREN to take care of token creation, and can extend it with rotation in the future.

jimmidyson: `aescbc` is also not recommended `due to CBC's vulnerability to padding oracle attacks`...

How about we support `aescbc` and `secretbox` as at least that doesn't seem to have any known vulnerabilities (yet)?

jimmidyson: The only way to `be secure by default` is to use KMS v2 right now i think, with `secretbox` a second place alternative.

jimmidyson: Note that OpenShift only supports `aescbc` and `aesgcm` (with weekly key rotation) - https://docs.openshift.com/container-platform/4.15/security/encrypting-etcd.html#etcd-encryption-types_encrypting-etcd.

supershal: We can add `secretbox` provider. I will add it to the API.

pkg/handlers/generic/mutation/encryption/inject_test.go
```diff
@@ -0,0 +1,233 @@
...
+	// 		{
+	// 			Operation:    "",
+	// 			Path:         "",
+	// 			ValueMatcher: Equal(nil),
+	// 		},
+	// 	},
+	// 	UnexpectedPatchMatchers: encryptionMatchers,
+	// }
+
+	// Context("Default encryption provider secret already exists", func() {
```
supershal, line 170: I was not able to get existing test helper to work with use case where the Response contains empty patch.
```
 <[]v1alpha1.GeneratePatchesResponseItem | len:1, cap:1>: [
          {
              UID: "4fa296dd-abbb-4629-9afd-c8457890248f",
              PatchType: "JSONPatch",
              Patch: "[]",
          },
      ]
```

api/v1alpha1/clusterconfig_types.go
```diff
@@ -200,6 +200,9 @@ type GenericClusterConfigSpec struct {
 
 	// +kubebuilder:validation:Optional
 	Users []User `json:"users,omitempty"`
+
+	// +optional
+	Encryption *Encryption `json:"encryption,omitempty"`
```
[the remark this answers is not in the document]
supershal: This came up during spike. https://jira.nutanix.com/browse/NCN-100160#:~:text=added%20a-,comment,-%2D%2019/Apr/24%209%3A16 . We concluded to not put it under `etcd`. @dlipovetsky provided following reasoning.

> Nothing here touches etcd in any way. [This is for the apiserver](https://kubernetes.io/docs/reference/config-api/apiserver-encryption.v1/) alone. The apiserver reads the EncryptionConfiguration, and encrypts/decrypts the values in etcd. Etcd has no knowledge of this encryption scheme at all.
I know that we historically referred to this as "etcd encryption," but I think that might have been a mistake.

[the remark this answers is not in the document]
jimmidyson: That's a good point. Maybe `DataEncryption` or `EncryptionAtRest`? Or even include `APIServer` in the name somehow? My concern is that `Encryption` is just too generic here.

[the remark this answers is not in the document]
supershal: in that case, I think `EncryptionAtRest` would be most accurate that reflects the patch variable.

api/v1alpha1/clusterconfig_types.go
```diff
@@ -279,6 +282,28 @@ type User struct {
...
+// Currently the encryption only enabled for secrets and configmaps.
+type EncryptionAtRest struct {
+	// Encryption providers
+	// +kubebuilder:default={{aescbc:{}}}
+	// +kubebuilder:maxItems=1
+	// +kubebuilder:validation:Optional
+	Providers []EncryptionProviders `json:"providers"`
+}
+
+type EncryptionProviders struct {
```
supershal, line 296: The apiserver allows only one type of the encryption provider. https://github.com/kubernetes/apiserver/blob/master/pkg/apis/apiserver/validation/validation_encryption.go#L311
`"more than one provider specified in a single element, should split into different list elements"`
`oneof` kubebuilder validation is still not implemented. https://github.com/kubernetes-sigs/controller-tools/issues/461

Should we implement the same validation in the patch handler?

jimmidyson: We can do, but I also think we can patch the CRD post generation to add the oneOf validation.

jimmidyson: We can also use cel validation once my upstream pr is finally reviewed and merged...

common/pkg/k8s/client/create.go
```diff
@@ -0,0 +1,30 @@
...
+package client
+
+import (
+	"context"
+	"fmt"
+
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func Create(
```
dkoshkin, line 13: Why do we need this new Create func over [Apply](https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/4b5cfde67d21d16542408082ae62f15dea86cd25/common/pkg/k8s/client/apply.go#L22)?

pkg/handlers/generic/mutation/encryption/inject.go
```diff
@@ -0,0 +1,307 @@
...
+	)
+
+	return patches.MutateIfApplicable(
+		obj, vars, &holderRef, selectors.ControlPlane(), log,
+		func(obj *controlplanev1.KubeadmControlPlaneTemplate) error {
+			cluster, err := clusterGetter(ctx)
+			if err != nil {
+				log.Error(err, "failed to get cluster from encryption mutation handler")
+				return err
+			}
```
dkoshkin, line 101: nit: in the other mutators this function is called early outside of the `MutateIfApplicable`, maybe we can do the same here?

pkg/handlers/generic/mutation/encryption/inject.go
```diff
@@ -0,0 +1,307 @@
...
+			if err != nil {
+				log.WithValues(
+					"defaultEncryptionSecret", defaultEncryptionSecretName(cluster.Name),
+				).Error(err, "failed to find default encryption configuration secret")
+				return err
+			}
+
+			// we do not rotate or override the secret keys for encryption configuration
+			if !found {
+				encConfig, err := h.generateEncryptionConfiguration(encryptionVariable.Providers)
```
dkoshkin, line 113: nit: I had to reread it a couple of times and though it was a typo for "etcd"
```suggestion
				encryptionConfig, err := h.generateEncryptionConfiguration(encryptionVariable.Providers)
```

pkg/handlers/generic/mutation/encryption/inject.go
```diff
@@ -0,0 +1,307 @@
...
+	return &apiserverv1.EncryptionConfiguration{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: apiserverv1.SchemeGroupVersion.String(),
+			Kind:       "EncryptionConfiguration",
+		},
+		Resources: resourceConfigs,
+	}, nil
+}
+
+func (h *encryptionPatchHandler) DefaultEncryptionSecretExists(
```
dkoshkin, line 186:
```suggestion
func (h *encryptionPatchHandler) defaultEncryptionSecretExists(
```

pkg/handlers/generic/mutation/encryption/inject.go
```diff
@@ -0,0 +1,307 @@
...
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return false, nil
+		}
+		return false, err
+	}
+	return true, nil
+}
+
+func (h *encryptionPatchHandler) CreateEncryptionConfigurationSecret(
```
dkoshkin, line 211:
```suggestion
func (h *encryptionPatchHandler) createEncryptionConfigurationSecret(
```

pkg/handlers/generic/mutation/encryption/inject.go
```diff
@@ -0,0 +1,307 @@
...
+
+	// We only support creating encryption config in BeforeClusterCreate hook and ensure that the keys are immutable.
+	if err := k8sClientUtil.Create(ctx, h.client, encryptionConfigSecret); err != nil {
+		return fmt.Errorf("failed to create encryption configuration secret: %w", err)
+	}
+	return nil
+}
+
+// We only support encryption for "secrets" and "configmaps".
+func encryptionConfigForSecretsAndConfigMaps(
```
dkoshkin, line 254: nit: maybe something a little more generic so it doesn't need to change if the details change?
```suggestion
func defaultEncryptionConfiguration(
```

pkg/handlers/generic/mutation/encryption/encryptionprovider_test.go
```diff
@@ -0,0 +1,87 @@
+// Copyright 2024 Nutanix. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package encryption
```
dkoshkin, line 4: nit: please rename the package and the file names
```suggestion
package encryptionatrest
```

pkg/handlers/generic/mutation/encryption/inject_test.go
```diff
@@ -0,0 +1,224 @@
...
+				testEncryptionSecretObj(),
+			)).To(BeNil())
+		})
+		It(noOpEncryptionConfigDef.Name, func(ctx SpecContext) {
+			capitest.AssertGeneratePatches(GinkgoT(), patchGenerator, &noOpEncryptionConfigDef)
+		})
+	})
+
+	// Test that encryption configuration secret is generated and patched on kubeadmconfig spec.
+	patchEncryptionConfigDef := capitest.PatchTestDef{
```
dkoshkin, line 188: Thanks for the unit test!

dkoshkin (approved):
Looks good, just some minor comments.

api/v1alpha1/clusterconfig_types.go
```diff
@@ -279,6 +282,27 @@ type User struct {
...
 }
 
+// Encryption defines the configuration to enable encryption at REST
+// This configuration is used by API server to encrypt data before storing it in ETCD.
+// Currently the encryption only enabled for secrets and configmaps.
+type Encryption struct {
+	// Encryption providers
+	// +kubebuilder:default={aescbc:{}}
+	// +kubebuilder:validation:Optional
+	Providers *EncryptionProviders `json:"providers"`
```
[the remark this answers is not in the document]
supershal: fixed.

common/pkg/k8s/client/create.go
supershal: This was to ensure that we dont update the secret with new randomly generated key. Once I added check for existing secret, `create` or `apply` would not matter, however I kept there to be clear that we do not update this secret in CAREN once it is generated.

pkg/handlers/generic/mutation/encryption/inject.go
supershal: This came up during pairing with Daniel when reviewing other PRs. The mutation handler is called for all the objects so clusterGetter will be called multiple times. Even though it is cached, I think we should only make Get call if the `obj` is the target object. i. .e in `MutateIfApplicable` method. 
If you agree, we want to change the same for other handlers.

api/v1alpha1/clusterconfig_types.go
supershal: > We can do, but I also think we can patch the CRD post generation to add the oneOf validation.
Thanks. I can try it out. is it okay if I file followup PR for this validation.
> We can also use cel validation once my upstream pr is finally reviewed and merged...
We can remove the workaround once we start using CEL validation is merged

pkg/handlers/generic/mutation/encryption/inject.go
supershal: fixed https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/commit/89d1408e8533448aa0168d904a7add4896d6150b

pkg/handlers/generic/mutation/encryption/inject.go
supershal: fixed https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/commit/89d1408e8533448aa0168d904a7add4896d6150b

pkg/handlers/generic/mutation/encryption/encryptionprovider_test.go
supershal: fixed https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/commit/89d1408e8533448aa0168d904a7add4896d6150b

api/v1alpha1/clusterconfig_types.go
jimmidyson: I looked into using `oneOf` and unfortunately cluster-api CC variables don't currently support it (see https://github.com/kubernetes-sigs/cluster-api/blob/main/api/v1beta1/clusterclass_types.go#L424 for what is currently supported). I've asked in [k8s slack](https://kubernetes.slack.com/archives/C8TSNPY4T/p1715774690914879) and will follow up to see if we can add upstream.

pkg/handlers/generic/mutation/encryption/inject.go
dkoshkin: sg, we can follow up in a separate PR for the other handlers.

jimmidyson (approved):

api/v1alpha1/clusterconfig_types.go
```diff
@@ -279,6 +282,28 @@ type User struct {
...
 
+// EncryptionAtRest defines the configuration to enable encryption at REST
+// This configuration is used by API server to encrypt data before storing it in ETCD.
+// Currently the encryption only enabled for secrets and configmaps.
+type EncryptionAtRest struct {
+	// Encryption providers
+	// +kubebuilder:default={{aescbc:{}}}
+	// +kubebuilder:validation:MaxItems=1
+	// +kubebuilder:validation:Optional
+	Providers []EncryptionProviders `json:"providers"`
```
jimmidyson, line 293:
```suggestion
	Providers []EncryptionProviders `json:"providers,omitempty"`
```

supershal: 7caae5872e7d10a939ac5040108cd161e64a8ad4

api/v1alpha1/clusterconfig_types.go
```diff
@@ -279,6 +282,28 @@ type User struct {
...
+}
+
+type EncryptionProviders struct {
+	// +kubebuilder:validation:Optional
+	AESCBC *AESConfiguration `json:"aescbc"`
```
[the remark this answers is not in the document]
supershal: fixed. 7caae5872e7d10a939ac5040108cd161e64a8ad4

api/v1alpha1/clusterconfig_types.go
```diff
@@ -279,6 +282,28 @@ type User struct {
...
+	// +kubebuilder:validation:Optional
+	Secretbox *SecretboxConfiguration `json:"secretbox"`
```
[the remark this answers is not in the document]
supershal: fixed. 7caae5872e7d10a939ac5040108cd161e64a8ad4

jimmidyson (approved):

PR #613, opened 2024-05-04
refactor: rename docker-quick-start to just docker

dkoshkin:
Alternative name ideas are welcome!

Wanted to propose this change, to me "quick start" sounds like non-production and implies you need something else to get full functionality. Thats not really the case here since these will be the default templates that all patches apply to. 

Keep in mind that clients will automatically get these templates when deploying with the helm chart, so it should probably be pretty generic.

**We can also think about a version suffix to these.**
Fixes #

If we agree, I will follow up with additional PRs for the other providers.

### DIFF
--- a/charts/cluster-api-runtime-extensions-nutanix/defaultclusterclasses/docker-cluster-class.yaml
+++ b/charts/cluster-api-runtime-extensions-nutanix/defaultclusterclasses/docker-cluster-class.yaml
@@ -3,23 +3,23 @@ kind: ClusterClass
 metadata:
   labels:
     cluster.x-k8s.io/provider: docker
-  name: docker-quick-start
+  name: docker
 spec:
   controlPlane:
     machineInfrastructure:
       ref:
         apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
         kind: DockerMachineTemplate
-        name: docker-quick-start-control-plane
+        name: docker-control-plane
     ref:
       apiVersion: controlplane.cluster.x-k8s.io/v1beta1
       kind: KubeadmControlPlaneTemplate
-      name: docker-quick-start-control-plane
+      name: docker
   infrastructure:
     ref:
       apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
       kind: DockerClusterTemplate
-      name: docker-quick-start-cluster
+      name: docker
   patches:
   - external:
       discoverVariablesExtension: dockerclusterconfigvars.cluster-api-runtime-extensions-nutanix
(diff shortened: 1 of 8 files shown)


jimmidyson:
Thank you for doing this! 

I think we can rename these to match the flavours that clusterctl accepts - this will also make using them through clusterctl easier.. I'm not sure what the names should be though!

dkoshkin:
> Thank you for doing this!
> 
> I think we can rename these to match the flavours that clusterctl accepts - this will also make using them through clusterctl easier.. I'm not sure what the names should be though!

Good idea. I also want to publish these as part of releases so it can be consumed by clients separate from the CAREN webhook deployment. Let me play around with this a little.

charts/cluster-api-runtime-extensions-nutanix/defaultclusterclasses/docker-cluster-class.yaml
```diff
@@ -3,23 +3,23 @@ kind: ClusterClass
 metadata:
   labels:
     cluster.x-k8s.io/provider: docker
-  name: docker-quick-start
+  name: docker
```
faiq, line 6: `docker-base` 
`docker-default`
`docker-caren`

dkoshkin:
> > Thank you for doing this!
> > I think we can rename these to match the flavours that clusterctl accepts - this will also make using them through clusterctl easier.. I'm not sure what the names should be though!
> 
> Good idea. I also want to publish these as part of releases so it can be consumed by clients separate from the CAREN webhook deployment. Let me play around with this a little.

Looked more into this, the `--flavor` currently [only works for an infra provider](https://github.com/kubernetes-sigs/cluster-api/blob/fd328d724e9ac22abf4f37df0606a9db0009da3d/cmd/clusterctl/cmd/generate_cluster.go#L186-L191) and I don't think we should fake this as an infra provider just for this clusterctl functionality.

The better option IMO is to also publish all the examples with the `Cluster` objects and all the templates so a cluster can be created using the `--from` flag, like this:
```
  # Generates a yaml file for creating workload clusters using a template from a specific URL.
  clusterctl generate cluster my-cluster --from https://github.com/foo-org/foo-repository/blob/main/cluster-template.yaml
```

jimmidyson:
We already publish the clusterclasses https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/main/.goreleaser.yml#L30 and the docs already use that approach to create clusters https://nutanix-cloud-native.github.io/cluster-api-runtime-extensions-nutanix/getting-started/create-cluster/nutanix/

PR #625, opened 2024-05-07
build: Latest devbox update

jimmidyson:

Fixes #


faiq (approved):
thank you. 

it would be good to get dependabot to do it

jimmidyson:
> it would be good to get dependabot to do it

Dependabot doesn't handle devbox tools... I did have a workflow but it can't create PRs any more so have to do it manually.

PR #637, opened 2024-05-08
feat: adds a mindthegap container and deployment 

faiq:

This mindthegap container is going to serve the helm charts for any helm addons we have.

https://jira.nutanix.com/browse/NCN-100596
https://jira.nutanix.com/browse/NCN-100597


make/addons.mk
```diff
@@ -3,12 +3,12 @@
 
 export CALICO_VERSION := v3.26.4
 export CILIUM_VERSION := 1.15.0
-export NODE_FEATURE_DISCOVERY_VERSION := 0.15.2
+export NODE_FEATURE_DISCOVERY_VERSION := v0.15.2
```
faiq, line 6: These two values changed because mindthegap wasn't able to get the tags as they were previously, but with these changes it was

faiq:
```
01:23 PM  faiqus @ archlinux  ~/go/src/github.com/d2iq-labs/cluster-api-runtime-extensions-nutanix  ⏎ 130  faiq/adds-mindthegap-container⚡?+ 
$ kubectl get po
NAME                                                      READY   STATUS    RESTARTS   AGE
cluster-api-runtime-extensions-nutanix-68f76fb8cf-6zngc   1/1     Running   0          5m51s
mindthegap-6ff84f74cc-68x6d                               1/1     Running   0          21m

01:23 PM  faiqus @ archlinux  ~/go/src/github.com/d2iq-labs/cluster-api-runtime-extensions-nutanix  faiq/adds-mindthegap-container⚡?+ 
$ kubectl get cm ^C

01:23 PM  faiqus @ archlinux  ~/go/src/github.com/d2iq-labs/cluster-api-runtime-extensions-nutanix  ⏎ 130  faiq/adds-mindthegap-container⚡?+ 
$ kubectl get cm default-helm-addons-config -o yaml
apiVersion: v1
data:
  cilium: |
    ChartName: cilium
    ChartVersion: 1.15.0
    RepositoryURL: registry.default.svc
  cluster-autoscaler: |
    ChartName: cluster-autoscaler
    ChartVersion: 9.35.0
    RepositoryURL: registry.default.svc
  metallb: |
    ChartName: metallb
    ChartVersion: v0.14.5
    RepositoryURL: registry.default.svc
  nfd: |
    ChartName: node-feature-discovery
    ChartVersion: v0.15.2
    RepositoryURL: registry.default.svc
  nutanix-ccm: |
    ChartName: nutanix-cloud-provider
    ChartVersion: 0.3.3
    RepositoryURL: registry.default.svc
  nutanix-snapshot-csi: |
    ChartName: nutanix-csi-snapshot
    ChartVersion: 6.3.2
    RepositoryURL: registry.default.svc
  nutanix-storage-csi: |
    ChartName: nutanix-csi-storage
    ChartVersion: v3.0.0-beta.1912
    RepositoryURL: registry.default.svc
  tigera-operator: |
    ChartName: tigera-operator
    ChartVersion: v3.26.4
    RepositoryURL: registry.default.svc
kind: ConfigMap
```

hack/addons/mindthegap-helm-registry/repos.yaml
```diff
@@ -0,0 +1,48 @@
...
+  cilium:
+    repoURL: https://helm.cilium.io/
+    charts:
+      cilium:
+      - 1.15.0
+  cluster-autoscaler:
+    repoURL: https://kubernetes.github.io/autoscaler
+    charts:
+      cluster-autoscaler:
+      - 9.35.0
```
deepakm-ntnx, line 18: anyway better way to keep it in sync with one in chart https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/637/files#diff-3b9b812f97c38c8ede8337eabf11d60d59b53e6ce54b66db1f2dced13035a5e6R16

deepakm-ntnx: same here too. https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/637/files#diff-4307f3fa02a9e4bd4c0051d5c9910ce9c8863cb0725f6c4e6bc2f29b7fd3d094R7

can we replace the same from addon.mk file everywhere so that we keep it at once place

deepakm-ntnx:
Could you please document in the description, general workflow with and without mindthegap from customer perspective. need a writeup to undestand from newbie perspective

hack/addons/mindthegap-helm-registry/repos.yaml
faiq: Thats a good question, the values for repos.yaml and the helm-config.yaml are derived from the makefile.

faiq:
the workflow should be transparent to the user. this is an internal implementation detail of how helm charts are served

.github/workflows/checks.yml
```diff
@@ -224,6 +224,14 @@ jobs:
               --name chart-testing \
               "ko.local/cluster-api-runtime-extensions-nutanix:$(devbox run -- gojq -r .version dist/metadata.json)"
 
+      - if: steps.list-changed.outputs.changed == 'true'
+        name: Sideload docker image caren-helm
+        run: |
+          devbox run -- \
+            kind load docker-image \
+            --name chart-testing \
+            "ghcr.io/nutanix-cloud-native/caren-helm-reg:$(devbox run -- gojq -r .version dist/metadata.json)"
```
jimmidyson, line 233: You can combine loading both docker images into the same step:

```suggestion
              "ko.local/cluster-api-runtime-extensions-nutanix:$(devbox run -- gojq -r .version dist/metadata.json)" \
              "ghcr.io/nutanix-cloud-native/caren-helm-reg:$(devbox run -- gojq -r .version dist/metadata.json)"
```

make/addons.mk
```diff
@@ -7,9 +7,9 @@ export NODE_FEATURE_DISCOVERY_VERSION := 0.15.2
 export CLUSTER_AUTOSCALER_VERSION := 9.37.0
 export AWS_CSI_SNAPSHOT_CONTROLLER_VERSION := v6.3.3
 export AWS_EBS_CSI_CHART_VERSION := v2.28.1
-export NUTANIX_STORAGE_CSI_CHART_VERSION := v3.0.0-beta.1912
-export NUTANIX_SNAPSHOT_CSI_CHART_VERSION := v6.3.2
-export LOCAL_PATH_CSI_CHART_VERSION := v0.0.29
+export NUTANIX_STORAGE_CSI_CHART_VERSION := 3.0.0-beta.1912
+export NUTANIX_SNAPSHOT_CSI_CHART_VERSION := 6.3.2
+export LOCAL_PATH_CSI_CHART_VERSION := 0.0.29
```
dkoshkin, line 12: Any way we can avoid changing these since the upstream versions do have the "v"?

pkg/handlers/generic/lifecycle/cni/calico/strategy_helmaddon.go
```diff
@@ -102,7 +102,7 @@ func (s helmAddonStrategy) apply(
 			ValuesTemplate:   values,
 		},
 	}
-
```
dkoshkin, line 105: nit: please keep the spacing consistent either with a newline with these changes across the different files.

pkg/handlers/generic/lifecycle/utils/utils.go
```diff
@@ -219,3 +222,14 @@ func CreateConfigMapForCRS(configMapName, configMapNamespace string,
 	cm.Data[defaultCRSConfigMapKey] = string(utilyaml.JoinYaml(l...))
 	return cm, nil
 }
+
+func SetTLSConfigForHelmChartProxyIfNeeded(hcp *caaphv1.HelmChartProxy) {
+	if strings.Contains(hcp.Spec.RepoURL, "mindthegap") {
+		hcp.Spec.TLSConfig = &caaphv1.TLSConfig{
+			CASecretRef: &corev1.SecretReference{
+				Name:      "mindthegap-tls",
```
dkoshkin, line 230: Maybe we can get this value from some env via the [downward API](https://kubernetes.io/docs/concepts/workloads/pods/downward-api/)

make/dev.mk
```diff
@@ -9,12 +9,17 @@ ifndef SKIP_BUILD
...
 	helm upgrade --install cluster-api-runtime-extensions-nutanix ./charts/cluster-api-runtime-extensions-nutanix \
+		--create-namespace \
+		--namespace caren-system \
 		--set-string image.repository=ko.local/cluster-api-runtime-extensions-nutanix \
 		--set-string image.tag=$$(gojq -r .version dist/metadata.json) \
+		--set-string mindthegapImage.tag=$$(gojq -r .version dist/metadata.json) \
 		--wait --wait-for-jobs
-	kubectl rollout restart deployment cluster-api-runtime-extensions-nutanix
-	kubectl rollout status deployment cluster-api-runtime-extensions-nutanix
+	kubectl rollout restart deployment -n caren-system cluster-api-runtime-extensions-nutanix
```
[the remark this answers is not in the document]
faiq: yes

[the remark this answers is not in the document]
faiq: the way that its deployed in dev doesn't reflect how it's deployed via clusterctl or helm either.

pkg/handlers/generic/lifecycle/utils/utils.go
faiq: the namespace - sure. although this is a fixed namespace.

pkg/handlers/generic/lifecycle/cni/calico/strategy_helmaddon.go
faiq: the spacing is consistent.

make/dev.mk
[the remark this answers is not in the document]
jimmidyson: It can be deployed in any namespace via helm. I'd like to keep it namespace agnostic if possible.

charts/cluster-api-runtime-extensions-nutanix/templates/helm-config.yaml
```diff
@@ -10,39 +10,39 @@ data:
   cilium: |
     ChartName: cilium
     ChartVersion: 1.15.5
-    RepositoryURL: https://helm.cilium.io/
+    RepositoryURL: {{ if .Values.selfHostedRegistry }}oci://mindthegap.{{ .Release.Namespace }}.svc/charts{{ else }}https://helm.cilium.io/{{ end }}
```
jimmidyson, line 13: You can just use `oci://mindthegap` for the address, you don't need the namespace or svc suffix. This will resolve to that service in the same namespace as the pod.

pkg/handlers/generic/lifecycle/utils/utils.go
faiq: added a change

charts/cluster-api-runtime-extensions-nutanix/templates/helm-config.yaml
faiq: these were some extra steps probably from the trust manager version of this. i like this simplifications

make/addons.mk
faiq: ill see if we can - mindthegap was being fussy about these previously

jimmidyson: Mindthegap fussy? Never!

hack/addons/mindthegap-helm-registry/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM ghcr.io/mesosphere/mindthegap:v1.13.4
+# this gets called by goreleaser so the copy source has to be the path relative to the repo root.
+RUN --mount=source=./hack/addons/mindthegap-helm-registry/repos.yaml,target=/repos.yaml ["/ko-app/mindthegap", "create", "helm-bundle", "--helm-charts-file=/repos.yaml", "--output-file=/tmp/helm-charts.tar"]
+VOLUME /certs
+CMD ["serve", "bundle", "--bundle=/tmp/helm-charts.tar", "--listen-port=5000", "--listen-address=0.0.0.0", "--tls-private-key-file=/certs/tls.key", "--tls-cert-file=/certs/tls.crt"]
```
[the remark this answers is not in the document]
faiq: i guess. this is already very tightly coupled with the build process and release and deployments can always override this so i don't really see a good reason to remove it.

charts/cluster-api-runtime-extensions-nutanix/templates/helm-config.yaml
faiq: hmm now that i think about it- i think i need the namespace here or add the namespace in code because Repository URL is eventually going to be used by CAAPH. 

lets keep this explicit for right now. i appreciate the suggestion for trying to simplify this!

charts/cluster-api-runtime-extensions-nutanix/templates/mindthegap.yaml
```diff
@@ -0,0 +1,78 @@
...
+  - mindthegap.{{ .Release.Namespace }}.svc.cluster.local
+  issuerRef:
+    name: mindthegap-selfsigned
+    kind: Issuer
+  secretName: mindthegap-tls
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: mindthegap
```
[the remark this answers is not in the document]
supershal: It would be good to use generic name like `chart-repository` instead of `mindthdgap` in deployments, cert resources etc, so that we can change the backend from mindthegap to something else if needed.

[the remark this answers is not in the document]
faiq: yep- working on changing to just `chart-repository`

hack/addons/mindthegap-helm-registry/Dockerfile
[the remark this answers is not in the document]
supershal: > I think this would fit better in the Deployment since it ultimately makes the decision about where to mount the volume and the port to run on

+1

.goreleaser.yml
```diff
@@ -92,6 +95,15 @@ archives:
     builds:
       - cluster-api-runtime-extensions-nutanix
 
+dockers:
+  - id: helm-registry-container
+    use: buildx
```
supershal, line 100: Do we need buildx to build this image?

.goreleaser.yml
```diff
@@ -92,6 +95,15 @@ archives:
     builds:
       - cluster-api-runtime-extensions-nutanix
 
+dockers:
```
supershal, line 98: why we decided to not use `ko` to build the image? i.e. building in similar way as CAREN docker image?

faiq: ko is a tool used for go binaries. i don't think it would be applicable here.

supershal: we create CAREN image using `kos` goreleaser plugin. https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/main/.goreleaser.yml#L95 
so i thought it would be consistent to use the same for building chart-repo image

charts/cluster-api-runtime-extensions-nutanix/templates/mindthegap.yaml
```diff
@@ -0,0 +1,78 @@
...
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: mindthegap
```
supershal, line 35: what CAREN clients have to do to move the chart-repository resources to the management cluster? should we apply `clusterctl.cluster.x-k8s.io/move` label ?

.goreleaser.yml
faiq: this is not a go binary so i dont think it works here.

make/addons.mk
faiq: i looked into this and its not the fault of mindthegap. for example

with this repos.yaml

```yaml
repositories:
  metallb:
    repoURL: https://metallb.github.io/metallb
    charts:
      metallb:
      - v0.14.5
```

mindthegap is able to find it correctly and create the oci archive. 

but since the index.yaml for metallb does not have the v 

```yaml 
apiVersion: v1
entries:
  metallb:
  - apiVersion: v2
    appVersion: v0.14.5
    created: "2024-04-19T14:44:27.58227863Z"
    dependencies:
    - condition: crds.enabled
      name: crds
      repository: ""
      version: 0.14.5
    - condition: frrk8s.enabled
      name: frr-k8s
      repository: https://metallb.github.io/frr-k8s
      version: 0.0.11
    description: A network load-balancer implementation for Kubernetes using standard
      routing protocols
    digest: 12e8d6a6ef2bf0d6450c10fba55c94282ef2d3559f78d2fbf626000d13baa9ef
    home: https://metallb.universe.tf
    icon: https://metallb.universe.tf/images/logo/metallb-white.png
    kubeVersion: '>= 1.19.0-0'
    name: metallb
    sources:
    - https://github.com/metallb/metallb
    type: application
    urls:
    - https://github.com/metallb/metallb/releases/download/metallb-chart-0.14.5/metallb-0.14.5.tgz
    version: 0.14.5
```

we don't see that when we list the tags on the chart 

```
$ curl -k https://localhost:5000/v2/charts/metallb/tags/list | jq .
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    44  100    44    0     0   7604      0 --:--:-- --:--:-- --:--:--  8800
{
  "name": "charts/metallb",
  "tags": [
    "0.14.5"
  ]
}

``` 

with all of this being said - i think stripping the v from these makes sense because we need to be able to access this as its served through the helm registry

hack/addons/mindthegap-helm-registry/Dockerfile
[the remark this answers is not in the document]
faiq: either is fine because you can override entrypoints and commands in any container. this seems like a very small nit which has no implications on the product or extensibility of it.

charts/cluster-api-runtime-extensions-nutanix/templates/mindthegap.yaml
faiq: im not sure any more this would get published and get installed by clusterctl so i dont think this is needed.

make/dev.mk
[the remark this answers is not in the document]
faiq: removed!

supershal (approved):
🎉  Thanks for solving very hard problem with airgap usecase.

.github/workflows/checks.yml
```diff
@@ -222,7 +222,8 @@ jobs:
           devbox run -- \
             kind load docker-image \
               --name chart-testing \
-              "ko.local/cluster-api-runtime-extensions-nutanix:$(devbox run -- gojq -r .version dist/metadata.json)"
+              "ko.local/cluster-api-runtime-extensions-nutanix:$(devbox run -- gojq -r .version dist/metadata.json)"\
```
dkoshkin, line 225: nit
```suggestion
              "ko.local/cluster-api-runtime-extensions-nutanix:$(devbox run -- gojq -r .version dist/metadata.json)" \
```

pkg/handlers/generic/lifecycle/utils/utils.go
```diff
@@ -219,3 +223,19 @@ func CreateConfigMapForCRS(configMapName, configMapNamespace string,
 	cm.Data[defaultCRSConfigMapKey] = string(utilyaml.JoinYaml(l...))
 	return cm, nil
 }
+
+func SetTLSConfigForHelmChartProxyIfNeeded(hcp *caaphv1.HelmChartProxy) {
+	// this is set as an environment variable from the downward API on deployment
+	deploymentNS := os.Getenv("POD_NAMESPACE")
```
dkoshkin, line 229: Is this env always set by some Kubernetes controller or something thats needs to be in the Pod spec?
(I couldn't find it in https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/)

api/v1alpha1/constants.go
```diff
@@ -27,4 +27,6 @@ const (
 	ClusterAutoscalerVariableName = "clusterAutoscaler"
 	// ServiceLoadBalancerVariableName is the Service LoadBalancer config patch variable name.
 	ServiceLoadBalancerVariableName = "serviceLoadBalancer"
+	// CarenNamespace is the namespace that caren is deployed in.
+	CarenNamespace = "caren-system"
```
dkoshkin, line 31: Having a default here is a little tricky here since `caren-system` is the namespace when publishing the charts, but `default` is the default when installing via the Helm chart.
Would it be alot of work to be more explicit and fail if the env is not set? Or maybe sticking with `default` since thats closer to what Kubernetes behaves like, ie an empty namespace == `default`?

WDYT?

pkg/handlers/generic/lifecycle/ccm/nutanix/handler.go
```diff
@@ -148,6 +149,7 @@ func (p *provider) Apply(
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: cluster.Namespace,
 			Name:      "nutanix-ccm-" + cluster.Name,
+			Labels:    utils.NewLabels(utils.WithMove()),
```
dkoshkin, line 152: Do these not get moved with `clusterctl move` like ClusterResourceSets do?

dkoshkin (approved):
Great job getting this across the line. and thanks for taking in my previous suggestions!

api/v1alpha1/constants.go
faiq: actually we don't even use this any more so i'll just remove it

pkg/handlers/generic/lifecycle/ccm/nutanix/handler.go
faiq: actually, they should be moved if they are owned by the cluster object. nice catch removing

pkg/handlers/generic/lifecycle/utils/utils.go
faiq: Its being set by the downward API already by the helm chart 

https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/9498fa43f0d0fcaa567dc2b776ca409a593f52d5/charts/cluster-api-runtime-extensions-nutanix/templates/deployment.yaml#L56

api/v1alpha1/constants.go
faiq: :thinking: it should be there because of the downward API. i suppose we can use default as a back up here then.

pkg/handlers/generic/lifecycle/utils/utils.go
dkoshkin: Ahh thats an existing env, got it 👍

supershal (approved):

charts/cluster-api-runtime-extensions-nutanix/values.yaml
```diff
@@ -84,6 +84,9 @@ hooks:
 
 helmAddonsConfigMap: default-helm-addons-config
 
+selfHostedRegistry: true
```
jimmidyson, line 87: Could we combine this with `helmRegistryImage` please? So the config looks like:

```
inClusterHelmOCIRegistry:
  enabled: true
  image:
    repository: repository: ghcr.io/nutanix-cloud-native/caren-helm-reg
    tag: ""
    pullPolicy: IfNotPresent
```

hack/tools/mindthegap-helm-reg/main.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	f, err = os.OpenFile(fullOutputfilePath, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0o666)
+	if err != nil {
+		log.Error(err, "failed to create file")
+	}
+	defer f.Close()
+	_, err = bytes.NewBuffer(b).WriteTo(f)
+	if err != nil {
+		log.Error(err, "failed to write to file")
+	}
```
jimmidyson, line 110: Simpler to just use `os.WriteFile`

```suggestion
	if err := os.WriteFile(fullOutputfilePath, b, 0o644); err != nil {
		log.Error(err, "failed to write mindthegap config file")
	}
```

charts/cluster-api-runtime-extensions-nutanix/templates/helm-repository.yaml
```diff
@@ -0,0 +1,78 @@
...
+  dnsNames:
+  - helm-repository.{{ .Release.Namespace }}.svc
+  - helm-repository.{{ .Release.Namespace }}.svc.cluster.local
+  issuerRef:
+    name: helm-repository-selfsigned
+    kind: Issuer
+  secretName: helm-repository-tls
+---
+apiVersion: apps/v1
+kind: Deployment
```
jimmidyson, line 30: Let's add tolerations and node affinity (`preferredDuringSchedulingIgnoredDuringExecution`) to allow running on CP nodes but prefer to run on non-CP nodes.

PR #644, opened 2024-05-09
fix: Always apply containerd patches

jimmidyson:
This commit ensures that applying containerd patches always happens. Previously,
this only happened if mirror configuration was provided, which meant that patches
such as enabling containerd metrics were never applied to containerd config.

By bundling the apply patches and restart containerd script in the same handler,
we ensure they both always get run.


dlipovetsky:
I was confused by the use of "apply patches" here, mostly because "patch" in the CAREN context usually refers to a hook handler.

IIUC, we write all of our containerd config changes as files to disk, but need to run a command to apply these changes to the actualy containerd configuration. This command was previously coupled with a specific type of containerd config change (mirror registry), and when that config change was not made, the command did not run.

jimmidyson:
@dlipovetsky That's correct, thanks for clarifying.

pkg/handlers/generic/mutation/handlers.go
```diff
@@ -41,6 +41,6 @@ func MetaMutators(mgr manager.Manager) []mutation.MetaMutator {
 		// Containerd restart and readiness altogether could take ~5s.
 		// We want to keep patch independent of each other and not share any state.
 		// Therefore, We must always apply this patch regardless any other patch modified containerd configuration.
-		containerdrestart.NewPatch(),
+		containerdapplypatchesandrestart.NewPatch(),
```
dlipovetsky, line 44: Could you add a note about applying patches to the comment above this? Because the comment hasn't changed, I can't make a suggestion to it in GitHub.

Instead of 

```
		// Some patches may have changed containerd configuration.
		// We must restart containerd for the configuration to take effect.
		// Therefore, we must apply this patch last.
```

I suggest this:

```
		// Some patches may have changed containerd configuration.
		// We write the configuration changes to disk, and must run a command
		// to apply the changes to the actual containerd configuration.
		// We then must restart containerd for the configuration to take effect.
		// Therefore, we must apply this patch last.
```

dlipovetsky (approved):
Thanks for fixing this!

pkg/handlers/generic/mutation/handlers.go
jimmidyson: Thanks! Updated.

PR #645, opened 2024-05-09
feat: Enable unprivileged ports sysctl in containerd config

jimmidyson:
This enabled pods to run as non-root and bind to privileged ports
as long as they have the necessary capability, `CAP_NET_BIND_SERVICE`
added.

This fixes an issue on AWS when bringing up coredns which binds to port
53 but runs as an unprivileged user.

Overall this is a net security improvement for clusters, meaning users
can stop giving too many privileged to pods - see
https://github.com/kubernetes/kubernetes/issues/102612 for discussion.

Depends on #644.

### DIFF
--- a/pkg/handlers/generic/mutation/containerdunprivilegedports/unprivileged_ports.go
+++ b/pkg/handlers/generic/mutation/containerdunprivilegedports/unprivileged_ports.go
@@ -1,0 +1,32 @@
+// Copyright 2024 Nutanix. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+package containerdunprivilegedports
+
+import (
+	_ "embed"
+	"path"
+
+	cabpkv1 "sigs.k8s.io/cluster-api/bootstrap/kubeadm/api/v1beta1"
+)
+
+const (
+	// TODO Factor out this constant to a common package.
+	containerdPatchesDirOnRemote = "/etc/containerd/cre.d"
+)
+
+var (
+	//go:embed files/unprivileged-ports-config.toml
+	unprivilegedPortsConfigDropIn             []byte
+	unprivilegedPortsConfigDropInFileOnRemote = path.Join(
+		containerdPatchesDirOnRemote,
+		"unprivileged-ports-config.toml",
+	)
+)
+
+func generateUnprivilegedPortsConfigDropIn() cabpkv1.File {
+	return cabpkv1.File{
+		Path:        unprivilegedPortsConfigDropInFileOnRemote,
+		Content:     string(unprivilegedPortsConfigDropIn),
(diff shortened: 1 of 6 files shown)


dlipovetsky:
CoreDNS v1.11.1 is deployed with Kubernetes v1.29.0.

Its securityContext stays the same as before:

```yaml
    securityContext:
      allowPrivilegeEscalation: false
      capabilities:
        add:
        - NET_BIND_SERVICE
        drop:
        - all
      readOnlyRootFilesystem: true
```

But it now run as a nonroot user (https://github.com/coredns/coredns/pull/5969).

However, even with this capability, containerd does not allow binding to an unprivileged port. For that, containerd must be configured as this PR does it.

For comparison, see an identical change to containerd configuration in https://github.com/kubernetes/minikube/pull/18421

dlipovetsky (approved):

faiq (dismissed):

faiq (approved):

test/e2e/config/caren.yaml
```diff
@@ -184,15 +184,15 @@ variables:
   AMI_LOOKUP_ORG: "999867407951"
   # To run Nutanix provider tests, set following variables here or as an env var
   # # IP/FQDN of Prism Central.
-  # NUTANIX_ENDPOINT: ""
+  NUTANIX_ENDPOINT: ""
```
dkoshkin, line 187: Was this change intentional?

jimmidyson: Not for this PR 😅 will revert

jimmidyson: Fixed.

pkg/handlers/generic/mutation/containerdunprivilegedports/unprivileged_ports.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	_ "embed"
+	"path"
+
+	cabpkv1 "sigs.k8s.io/cluster-api/bootstrap/kubeadm/api/v1beta1"
+)
+
+const (
+	// TODO Factor out this constant to a common package.
```
dkoshkin, line 13: Is this PR enough https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/656/files?

dkoshkin (approved):
Thanks for figuring this out!

pkg/handlers/generic/mutation/containerdunprivilegedports/unprivileged_ports.go
jimmidyson: Good catch! Thanks! Updated to use new common function to derive patch file path.

PR #648, opened 2024-05-09
fix: Make Cluster the owner of image registry credential secret

dlipovetsky:
Makes the Cluster resource the owner of image registry credential Secret. CAREN creates this Secret for each Cluster, and it should be deleted when the Cluster is deleted. 

Fixes https://jira.nutanix.com/browse/D2IQ-100572

All test cases pass. The test is updated to create a Cluster resource, and to provide the handler a client that can read a Cluster resource.

### DIFF
--- a/pkg/handlers/generic/mutation/imageregistries/credentials/inject.go
+++ b/pkg/handlers/generic/mutation/imageregistries/credentials/inject.go
@@ -12,11 +12,13 @@ import (
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	bootstrapv1 "sigs.k8s.io/cluster-api/bootstrap/kubeadm/api/v1beta1"
 	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
 	runtimehooksv1 "sigs.k8s.io/cluster-api/exp/runtime/hooks/api/v1alpha1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	"github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/api/v1alpha1"
 	"github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/common/pkg/capi/clustertopology/handlers/mutation"
@@ -69,7 +71,7 @@ func (h *imageRegistriesPatchHandler) Mutate(
 	vars map[string]apiextensionsv1.JSON,
 	holderRef runtimehooksv1.HolderReference,
 	clusterKey ctrlclient.ObjectKey,
-	_ mutation.ClusterGetter,
+	clusterGetter mutation.ClusterGetter,
 ) error {
 	log := ctrl.LoggerFrom(ctx).WithValues(
 		"holderRef", holderRef,
@@ -172,7 +174,16 @@ func (h *imageRegistriesPatchHandler) Mutate(
 				commands...,
 			)
 
-			generateErr = createSecretIfNeeded(ctx, h.client, registriesWithOptionalCredentials, clusterKey)
+			cluster, err := clusterGetter(ctx)
+			if err != nil {
+				log.Error(
+					err,
+					"failed to get cluster from Image Registry Credentials mutation handler",
+				)
+				return err
+			}
+
+			generateErr = createSecretIfNeeded(ctx, h.client, registriesWithOptionalCredentials, cluster)
 			if generateErr != nil {
 				return generateErr
 			}
@@ -216,7 +227,16 @@ func (h *imageRegistriesPatchHandler) Mutate(
 			).Info("adding PreKubeadmCommands to worker node kubeadm config template")
 			obj.Spec.Template.Spec.PreKubeadmCommands = append(obj.Spec.Template.Spec.PreKubeadmCommands, commands...)
 
-			generateErr := createSecretIfNeeded(ctx, h.client, registriesWithOptionalCredentials, clusterKey)
+			cluster, err := clusterGetter(ctx)
+			if err != nil {
+				log.Error(
+					err,
+					"failed to get cluster from Image Registry Credentials mutation handler",
+				)
+				return err
+			}
+
+			generateErr := createSecretIfNeeded(ctx, h.client, registriesWithOptionalCredentials, cluster)
 			if generateErr != nil {
 				return generateErr
 			}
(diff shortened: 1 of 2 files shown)


pkg/handlers/generic/mutation/imageregistries/credentials/inject.go
```diff
@@ -172,9 +200,8 @@ func (h *imageRegistriesPatchHandler) Mutate(
 				commands...,
 			)
 
-			generateErr = createSecretIfNeeded(ctx, h.client, registriesWithOptionalCredentials, clusterKey)
-			if generateErr != nil {
-				return generateErr
+			if err := client.ServerSideApply(ctx, h.client, credentialsSecret, client.ForceOwnership); err != nil {
+				return fmt.Errorf("failed to apply Image Registry Credentials Secret: %w", err)
```
jimmidyson, line 204: Why not do this when setting the owner ref above outside of the mutation?

jimmidyson:
Are there any changes to e2e needed for this too? eg ownership checking of secrets in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/main/test/e2e/ownerreference_helpers.go#L162 is currently disabled

pkg/handlers/generic/mutation/imageregistries/credentials/inject_test.go
```diff
@@ -392,22 +394,44 @@ var _ = Describe("Generate Image registry patches", func() {
...
-		client, err := helpers.TestEnv.GetK8sClient()
+		client, err := helpers.TestEnv.GetK8sClientWithScheme(clientScheme)
 		gomega.Expect(err).To(gomega.BeNil())
+
 		gomega.Expect(client.Create(
 			ctx,
 			newRegistryCredentialsSecret(validSecretName, request.Namespace),
 		)).To(gomega.BeNil())
+
+		gomega.Expect(client.Create(
```
supershal, line 405: I like explicitly creating cluster in unit test. I could use this pattern in encryption handler unit tests to verify secret ownership.

pkg/handlers/generic/mutation/imageregistries/credentials/inject.go
```diff
@@ -151,6 +152,33 @@ func (h *imageRegistriesPatchHandler) Mutate(
 		return generateErr
 	}
 
+	credentialsSecret, generateErr := generateCredentialsSecret(
```
supershal, line 155: `generateCredentialsSecret` has side effects. it returns `nil,nil` if `registriesWithOptionCredentials` is empty. https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/main/pkg/handlers/generic/mutation/imageregistries/credentials/credentials_secret.go#L57.
```
if len(configs) == 0 {
		return nil, nil
	}
```
This will break this mutator. 
It think we need to refactor that function to move that condition outside.

pkg/handlers/generic/mutation/imageregistries/credentials/inject.go
dlipovetsky: IIUC, in theory, both mutations may be skipped.

The handler should have side effects only if some mutation is executed.

Creating a resource is a side effect, so it must be done only if the mutation is executed.

jimmidyson: Ok could we make sure it's only applied a maximum of once then?

pkg/handlers/generic/mutation/imageregistries/credentials/inject_test.go
supershal: wdyt moving the cluster create and delete to common test utils package?

pkg/handlers/generic/mutation/imageregistries/credentials/inject_test.go
```diff
@@ -392,22 +394,44 @@ var _ = Describe("Generate Image registry patches", func() {
...
 			ctx,
 			newRegistryCredentialsSecret(validSecretName, request.Namespace),
 		)).To(gomega.BeNil())
+
+		gomega.Expect(client.Delete(
+			ctx,
+			&clusterv1.Cluster{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      request.ClusterName,
+					Namespace: metav1.NamespaceDefault,
```
supershal, line 431:
```suggestion
					Namespace: request.Namespace
```

pkg/handlers/generic/mutation/imageregistries/credentials/inject_test.go
```diff
@@ -392,22 +394,44 @@ var _ = Describe("Generate Image registry patches", func() {
...
 			ctx,
 			newRegistryCredentialsSecret(validSecretName, request.Namespace),
 		)).To(gomega.BeNil())
+
+		gomega.Expect(client.Create(
+			ctx,
+			&clusterv1.Cluster{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      request.ClusterName,
+					Namespace: metav1.NamespaceDefault,
```
supershal, line 410:
```suggestion
					Namespace: request.Namespace,
```

pkg/handlers/generic/mutation/imageregistries/credentials/inject.go
dlipovetsky: I don't understand the motivation.

Do we want the two `patches.MutateIfApplicable` calls to be mutually exclusive?

If so, then we should make them mutually exclusive, and that would, in turn, guarantee that we apply the ownerRef at most once.

If we do not want the calls to be mutually exclusive, then I don't understand why we should apply the ownerRef at most once.

dlipovetsky:
> Are there any changes to e2e needed for this too? eg ownership checking of secrets in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/main/test/e2e/ownerreference_helpers.go#L162 is currently disabled

Good call-out. I'm investigating this, and will address it in a separate PR.

pkg/handlers/generic/mutation/imageregistries/credentials/inject_test.go
dlipovetsky: Yes, I want to do that. I tried it, then decided to do it as a follow-up PR.

pkg/handlers/generic/mutation/imageregistries/credentials/inject_test.go
dlipovetsky: Thanks!

pkg/handlers/generic/mutation/imageregistries/credentials/inject_test.go
dlipovetsky: Thanks!

pkg/handlers/generic/mutation/imageregistries/credentials/inject_test.go
supershal: sounds good. thanks for making this improvement.

pkg/handlers/generic/mutation/imageregistries/credentials/inject.go
jimmidyson: Just an optimisation, reducing unnecessary calls to the API server.

dlipovetsky:
The refactors became rather complicated, and I would like to take them up in a separate PR. With that in mind, I'm force-pushing a minimal set of changes, just enough to correctly set the owner ref, and ensure the test verifies it.

Thank you @jimmidyson and @supershal. I'll make sure to address your comments w.r.t the refactors in the separate PR.

jimmidyson (approved):

mhrabovcin (approved):

PR #656, opened 2024-05-13
refactor: Write configuration under /etc/caren

dlipovetsky:
Previously, we wrote some configuration to /etc/cre. Now that the project name is CAREN, we should use /etc/caren.

We also wrote containerd configuration to /etc/containerd/cre.d, but this configuration is not read by containerd directly. Instead, it is read by a script that merges it to the primary containerd configuration. For that reason, this configuration belongs under /etc/caren.

Fixes #

### DIFF
--- a/pkg/handlers/generic/mutation/imageregistries/credentials/inject_test.go
+++ b/pkg/handlers/generic/mutation/imageregistries/credentials/inject_test.go
@@ -161,7 +161,7 @@ var _ = Describe("Generate Image registry patches", func() {
 					Path:      "/spec/template/spec/kubeadmConfigSpec/files",
 					ValueMatcher: gomega.ContainElements(
 						gomega.HaveKeyWithValue(
-							"path", "/etc/cre/install-kubelet-credential-providers.sh",
+							"path", "/etc/caren/install-kubelet-credential-providers.sh",
 						),
 						gomega.HaveKeyWithValue(
 							"path", "/etc/kubernetes/image-credential-provider-config.yaml",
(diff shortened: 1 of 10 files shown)


pkg/handlers/generic/mutation/imageregistries/credentials/inject_test.go
```diff
@@ -161,7 +161,7 @@ var _ = Describe("Generate Image registry patches", func() {
 					Path:      "/spec/template/spec/kubeadmConfigSpec/files",
 					ValueMatcher: gomega.ContainElements(
 						gomega.HaveKeyWithValue(
-							"path", "/etc/cre/install-kubelet-credential-providers.sh",
+							"path", "/etc/caren/install-kubelet-credential-providers.sh",
```
deepakm-ntnx, line 164: should we reuse constants everywhere instead of specifying paths directly (even in tests)

dlipovetsky: In https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/656/commits/1c618d607dd6ece68645845276c9a61fbc962813, I added common code to derive the path.

However, I think it is better to not use this code in tests. If we unintentionally change the path, we want the tests to fail, and not using the code ensures this. (On the other hand, if we intentionally change the path, we can update the tests using search/replace.)

jimmidyson:
I've rebased to fix up conflicts and made a few tweaks to naming, as well as using the new config path functions where we add containerd restart and apply patches scripts. I'm going to get this merged, but if you feel we should change the paths, or naming, etc then we can do as follow up PR.

mhrabovcin (approved):

PR #660, opened 2024-05-15
fix: issues with mindthegap deplotment 

faiq:

This adds trust-manager to allow us to use self hosted helm charts in mindthegap. See: https://github.com/kubernetes-sigs/cluster-api-addon-provider-helm/issues/104

We can think about this in a couple of different parts: 

1. Installing trust-manager and CRDs. We need to install trust manager so that we can create a certificate bundle with our self signed cert to caaph-controller-manager. We pull the CRDs separately because we are unable to use them in our templates because they aren't available at that time. See the comment in the script for more details.

2. Modifying CAAPH to use the bundle. This is done through a job it has a corresponding Service Account/Role/Role Binding. 

3. Various little fixes. ie changing repository url and versions to using `v` 

I tested this by creating a docker cluster. 

The helm chart proxies use this value 

```yaml
- apiVersion: addons.cluster.x-k8s.io/v1alpha1
  kind: HelmChartProxy
  metadata:
    creationTimestamp: "2024-05-14T23:41:09Z"
    finalizers:
    - helmchartproxy.addons.cluster.x-k8s.io
    generation: 1
    name: node-feature-discovery-docker-cluster-cilium-helm-addon
    namespace: default
    ownerReferences:
    - apiVersion: cluster.x-k8s.io/v1beta1
      kind: Cluster
      name: docker-cluster-cilium-helm-addon
      uid: e0e943e4-6d63-4d85-8d37-f63160ff93e3
    resourceVersion: "3282"
    uid: e04b6b62-5f39-4528-87f7-5928d2afd4dd
  spec:
    chartName: node-feature-discovery
    clusterSelector:
      matchLabels:
        cluster.x-k8s.io/cluster-name: docker-cluster-cilium-helm-addon
    namespace: node-feature-discovery
    options:
      enableClientCache: false
      install:
        createNamespace: true
      timeout: 10m0s
      upgrade:
        maxHistory: 10
    releaseName: node-feature-discovery
    repoURL: oci://mindthegap.mindthegap.svc/charts
```

Fixes #


charts/cluster-api-runtime-extensions-nutanix/templates/job.yaml
```diff
@@ -0,0 +1,73 @@
...
+# Copyright 2023 D2iQ, Inc. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: update-trust-manager-namespace
+  namespace: caaph-system
+  annotations:
+    "helm.sh/hook": post-install,post-upgrade
```
faiq, line 13: not sure if we can keep helm hooks

charts/cluster-api-runtime-extensions-nutanix/templates/mindthegap.yaml
```diff
@@ -1,23 +1,38 @@
...
+# The pod is built via goreleaser with configuration from hack/addons.
+#
+# NOTE: We use a hard code namespace here because trust-manager
+# need to set app.trust.namespace to be able to read secrets
+# from this namespace. We then distribute the certificates
+# to CAAPH system, which is the primary user of this service.
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: mindthegap
```
faiq, line 17: this is causing the failure in the e2e tests. i think we're going to need to put caren `caren-system` and put this mindthegap registry with it

charts/cluster-api-runtime-extensions-nutanix/Chart.lock
```diff
@@ -0,0 +1,6 @@
+dependencies:
+- name: trust-manager
```
jimmidyson, line 2: I'm not keen on trust-manager being a requirement for CAREN - shouldn't this be a user deployment configuration rather than a strict requirement? It could conflict with existing trust-manager deployment for example.

faiq: We can keep this functionality until we implement https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/497

We might be able to then set this behind a boolean gate. If that is the case, i'm not sure what the implication for konvoy would be. How should we template it? 

On the other hand, CAREN is already a highly opinionated and has a ton of dependencies adding another isn't that big of a deal IMO

hack/addons/kustomize/nfd/kustomization.yaml.tmpl
```diff
@@ -18,7 +18,7 @@ helmCharts:
   valuesFile: helm-values.yaml
   valuesInline:
     image:
-      tag: "${NODE_FEATURE_DISCOVERY_VERSION}-minimal"
+      tag: "v${NODE_FEATURE_DISCOVERY_VERSION}-minimal"
```
dkoshkin, line 21: Wait was the NFD Deployment never coming up because of a wrong image tag?

charts/cluster-api-runtime-extensions-nutanix/templates/helm-config.yaml
```diff
@@ -10,36 +10,37 @@ data:
   cilium: |
     ChartName: cilium
     ChartVersion: 1.15.0
-    RepositoryURL: oci://mindthegap.{{ .Release.Namespace }}.svc
+    RepositoryURL: oci://mindthegap.caren-system.svc/charts
```
dkoshkin, line 13: This won't work if deployed in a different namespace?

dkoshkin (commented):
Did you observe a cert error? From CAAPH code it looks like its always setting [the insecure flag](https://github.com/kubernetes-sigs/cluster-api-addon-provider-helm/blob/293b35e0ede1552b960d7679b484289d4faed763/internal/helm_client.go#L70).

hack/addons/kustomize/nfd/kustomization.yaml.tmpl
jimmidyson: NFD was definitely working, e2e test validates this

charts/cluster-api-runtime-extensions-nutanix/values.yaml
```diff
@@ -1,6 +1,10 @@
 # Copyright 2023 Nutanix. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
-
+#
+trust-manager:
+  app:
+    trust:
+      namespace: caren-system
```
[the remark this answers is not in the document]
faiq: values.yaml doesn't use templates :( https://github.com/helm/helm/issues/2492

charts/cluster-api-runtime-extensions-nutanix/templates/helm-config.yaml
faiq: that is correct. it needs to be deployed in that namespace, this is a limitation of trust-manager needing trust namespace set. this comment explains it 

https://github.com/cert-manager/trust-manager/issues/120#issuecomment-1549984274

if helm was able to use templating in values.yaml we could get around this

hack/addons/kustomize/nfd/kustomization.yaml.tmpl
faiq: this had something to do with mindthegap not being able to find the tag

hack/examples/overlays/clusterclasses/aws/kustomization.yaml.tmpl
```diff
@@ -3,6 +3,7 @@
 
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
+namespace: caren-system
```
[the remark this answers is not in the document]
faiq: this can be an issue down the line for sure :thinking:

charts/cluster-api-runtime-extensions-nutanix/values.yaml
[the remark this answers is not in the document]
dkoshkin: I ran `make dev` with this PR but removed the above `trust-manager` config and it seems to work correctly and already deploys the trust-manager chart into the correct namespace:
```
kubectl get pods -n caren-system
NAME                                                      READY   STATUS    RESTARTS   AGE
cluster-api-runtime-extensions-nutanix-7f5dc9ffdd-gxl8r   1/1     Running   0          65s
mindthegap-c57f655d5-k9sp4                                1/1     Running   0          88s
trust-manager-865cff7789-c2vc6                            1/1     Running   0          88s
```

The chart handles this correctly https://github.com/cert-manager/trust-manager/blob/2cb20f7d742d44a9eb3f2ebf62c415fffb34aa57/deploy/charts/trust-manager/templates/_helpers.tpl#L55

I think we're good to remove this and not hardcode the namespace?

jimmidyson:
I've pushed to fix this issue in CAAPH without requiring trust-manager - I think this approach is very clean and easy to work with and as such feel we should rework this PR based on that approach.

charts/cluster-api-runtime-extensions-nutanix/values.yaml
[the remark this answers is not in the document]
faiq: the issue is with referencing the secret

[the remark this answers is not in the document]
faiq: this comment shows why we need to set trust namespace https://github.com/cert-manager/trust-manager/issues/120#issuecomment-1549984274

PR #663, opened 2024-05-15
refactor: Use a Credentials struct consistently

dlipovetsky:
Previously, we used a Credentials struct in some places, and a LocalObjectReference in others.  With this change, we use a Credentials struct in all places.

Fixes #


api/v1alpha1/addon_types.go
```diff
@@ -174,11 +174,23 @@ type StorageClassConfig struct {
 	AllowExpansion bool `json:"allowExpansion,omitempty"`
 }
 
+type CSICredentials struct {
```
dkoshkin, line 177: I think the examples need to change too https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/tree/main/examples/capi-quick-start, and possible some snippets in docs

dkoshkin (commented):
Thanks for making it more explicit

api/v1alpha1/addon_types.go
dlipovetsky: Good catch, I'll update that

dlipovetsky: Fixed in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/663/commits/18dbf4def55e390325a1f056a3abd9c40aeaea50

dkoshkin (approved):

supershal (approved):
Thank you keep fixing existing consistency issues.

PR #665, opened 2024-05-17
test(e2e): Fix up secret ownership checks

jimmidyson:
This commit copies the upstream check by allowing `KubeadmControlPlane` and
`KubeadmConfig` owners, but adds `Cluster` as well as the secrets that CAREN
generates are owned by the relevant `Cluster` to ensure they are cleaned up
as part of Kubernetes resource garbage collection following deletion of the
`Cluster`.

### DIFF
--- a/test/e2e/ownerreference_helpers.go
+++ b/test/e2e/ownerreference_helpers.go
@@ -11,6 +11,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/ptr"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	bootstrapv1 "sigs.k8s.io/cluster-api/bootstrap/kubeadm/api/v1beta1"
+	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
 	addonsv1 "sigs.k8s.io/cluster-api/exp/addons/api/v1beta1"
 	"sigs.k8s.io/cluster-api/test/framework"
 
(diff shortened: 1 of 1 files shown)


dkoshkin (approved):

faiq (approved):

test/e2e/ownerreference_helpers.go
```diff
@@ -167,9 +183,15 @@ var (
 	// https://github.com/kubernetes-sigs/cluster-api/tree/main/docs/book/src/reference/owner_references.md.
 	KubernetesReferenceAssertions = map[string]func([]metav1.OwnerReference) error{
 		secretKind: func(owners []metav1.OwnerReference) error {
-			// TODO:deepakm-ntnx Currently pc-creds, pc-creds-for-csi, dockerhub-credentials,
-			// registry-creds, and encryption config secrets have unexpected owners which needs more investigation.
-			return nil
+			// Secrets for cluster certificates must be owned and controlled by the KubeadmControlPlane.
+			// The bootstrap secret should be owned and controlled by a KubeadmControlPlane.
+			// Other resources can be owned by the Cluster to ensure correct GC.
+			return framework.HasOneOfExactOwners(
```
deepakm-ntnx, line 189: wouldnt this mandate that every secret should be owned by something. there are few secrets which are not. does this test pass now?

dlipovetsky: I think you're right; this check does not allow user-created Secrets in the Cluster namespace, but it needs to.

We configure CI to test only the Docker provider. I think [this](https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/c5384877cf6f4afcce3e306e658667c95047b757/examples/capi-quick-start/docker-cluster-cilium-crs.yaml#L24) is the cluster spec the test uses.

AFAICT, the test does not configure registry credentials. And the Docker provider does not require credentials for any Addons (e.g. CSI, CCM). Therefore the test framework (which acts as the user here) does not create any Secrets in the Cluster namespace; CAREN does, but these Secrets are owned by the Cluster, by design.

PR #667, opened 2024-05-17
test: Remove hard-coded text focus and label for e2e tests

dlipovetsky:
Removes hard-coded text focus and label for e2e tests.

This code took precedent over any configuration provided to ginkgo at
runtime, so only single e2e test matching both focus and labels could
run.

Fixes #

I used dry-run (added in #666) to confirm that the removed code was the root cause.


dlipovetsky:
Depends on #666

jimmidyson (approved):
Omg sorry I needed to add this to debug tests and forgot to remove it 🤦‍♂️ Thanks for fixing!

supershal (approved):
Thanks.

PR #673, opened 2024-05-20
e2e: use all CPUs running ginkgo tests

dkoshkin:
There are 4 e2e tests running in 3 tests in parallel, effectively doubling the time to run e2e tests. Testing here to use all CPUs in the runners.

Fixes #

### DIFF
--- a/make/go.mk
+++ b/make/go.mk
@@ -76,7 +76,7 @@ bench.%: ; $(info $(M) running benchmarks$(if $(GOTEST_RUN), matching "$(GOTEST_
 
 E2E_DRYRUN ?= false
 E2E_VERBOSE ?= $(filter $(E2E_DRYRUN)$(CI),true) # If dry-run or CI, enable verbosity
-E2E_PARALLEL_NODES ?= $(if $(filter $(E2E_DRYRUN),true),1,$(shell nproc --ignore=1)) # Ginkgo cannot dry-run in parallel
+E2E_PARALLEL_NODES ?= $(if $(filter $(E2E_DRYRUN),true),1,$(shell nproc)) # Ginkgo cannot dry-run in parallel
 E2E_FLAKE_ATTEMPTS ?= 1
 E2E_CONF_FILE ?= $(REPO_ROOT)/test/e2e/config/caren.yaml
 E2E_CONF_FILE_ENVSUBST ?= $(basename $(E2E_CONF_FILE))-envsubst.yaml
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -71,3 +71,10 @@ jobs:
           check_name: 'e2e test report'
           detailed_summary: true
           require_passed_tests: true
+
+      - if: success() || failure() # always run even if the previous step fails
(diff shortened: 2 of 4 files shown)


make/go.mk
```diff
@@ -74,7 +74,7 @@ bench.%: ## Runs go benchmarks for a specific module
 bench.%: ; $(info $(M) running benchmarks$(if $(GOTEST_RUN), matching "$(GOTEST_RUN)") for $* module)
 	$(if $(filter-out root,$*),cd $* && )go test $(if $(GOTEST_RUN),-run "$(GOTEST_RUN)") -race -cover -v ./...
 
-E2E_PARALLEL_NODES ?= $(shell nproc --ignore=1)
+E2E_PARALLEL_NODES ?= $(shell nproc)
```
jimmidyson, line 77: This was originally here to not lock up local dev machines by using all processors. I wonder if we can change this to just use all procs only in CI?
```suggestion
ifeq ($(CI),true)
  E2E_PARALLEL_NODES ?= $(shell nproc)
else
  E2E_PARALLEL_NODES ?= $(shell nproc --ignore 1)
endif
```

jimmidyson:
> There are 4 e2e tests running in 3 tests in parallel

Actually tests are running in Serial - see https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/main/test/e2e/quick_start_test.go#L25. This is especially (only?) important for Docker provider where we need wait times to not be impacted by number of concurrent clusters to try to prevent flaky tests by hitting contention issues.

make/go.mk
jimmidyson: Ran without this change locally and it all seems fine. I guess normal usage locally will focus tests so it will be ok to run utilising all CPUs. Let's try it out and see.

dkoshkin:
>Actually tests are running in Serial 

🤔 From this e2e the tests seem way too fast, but either way this change is not gaining us anything since the self-hosted tests still take about ~22 mins. Will close this PR and we can revisit this if we enable parallel tests.

This PR:
```
Mon, 20 May 2024 13:28:42 GMT
  cluster.cluster.x-k8s.io/quick-start-jgamzk created
Mon, 20 May 2024 13:28:45 GMT
  cluster.cluster.x-k8s.io/quick-start-hdxzq7 created
Mon, 20 May 2024 13:28:57 GMT
  cluster.cluster.x-k8s.io/quick-start-li8hxh created
Mon, 20 May 2024 13:29:07 GMT
  cluster.cluster.x-k8s.io/quick-start-hz30tt created
```

main:
```
Mon, 20 May 2024 17:48:47 GMT
  cluster.cluster.x-k8s.io/quick-start-jxtryf created
Mon, 20 May 2024 17:50:39 GMT
  cluster.cluster.x-k8s.io/quick-start-wvslt2 created
Mon, 20 May 2024 17:53:01 GMT
  cluster.cluster.x-k8s.io/quick-start-m7ko2d created
Mon, 20 May 2024 17:55:24 GMT
  cluster.cluster.x-k8s.io/quick-start-eicgp4 created
```

PR #676, opened 2024-05-20
feat: Pull calico images from quay.io instead of docker hub

jimmidyson:
This reduces need for Docker Hub credentials by pulling images from
another registry that has Calico images pushed to on new Calico
releases.

### DIFF
--- a/charts/cluster-api-runtime-extensions-nutanix/templates/cni/calico/manifests/aws/crs-installation.yaml
+++ b/charts/cluster-api-runtime-extensions-nutanix/templates/cni/calico/manifests/aws/crs-installation.yaml
@@ -32,5 +32,6 @@ data:
         bgp: Enabled
       nodeMetricsPort: 9091
       typhaMetricsPort: 9093
+      registry: quay.io/
 {{- end -}}
 {{- end -}}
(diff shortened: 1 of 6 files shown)


dlipovetsky:
Do you think this change will have a noticable impact, given the number (and size) of other images we pull from docker?

I also have no strong objections.

dlipovetsky (approved):

jimmidyson:
@dlipovetsky For CAREN it does... I've hit pull limits locally without this. For cluster lifecycle tests (konvoy) I also think it may have an impact, for Kommander no it won't help too much at all 😅

supershal (approved):

dkoshkin:
Not too much of an impact, Konvoy will need to update the [script](https://github.com/mesosphere/konvoy2/blob/051fdf94628a09028d0c4befe8d27b04876d37f7/hack/functions.sh#L112) that lists CAREN images, something similar to [this](https://github.com/mesosphere/konvoy2/blob/051fdf94628a09028d0c4befe8d27b04876d37f7/hack/functions.sh#L133C1-L133C18) but with `quay.io`.

I will restart the conversation on how CAREN can publish the images it needs [initially attempted here](https://github.com/d2iq-labs/cluster-api-runtime-extensions-nutanix/pull/2).

jimmidyson:
I think this has an impact on people running just with CAREN as no Docker Hub credentials will be required for that afaict, no images coming from there i don't think.

PR #677, opened 2024-05-20
fix: Upgrade dynamic-credential-provider to v0.5.3

jimmidyson:
Fixes falling back to origin registry if mirror does not contain
requested image.

### DIFF
--- a/pkg/handlers/generic/mutation/imageregistries/credentials/credential_provider_config_files_test.go
+++ b/pkg/handlers/generic/mutation/imageregistries/credentials/credential_provider_config_files_test.go
@@ -267,7 +267,7 @@ credentialProviders:
 kind: DynamicCredentialProviderConfig
 mirror:
   endpoint: 98765432.dkr.ecr.us-east-1.amazonaws.com
-  credentialsStrategy: MirrorCredentialsOnly
+  credentialsStrategy: MirrorCredentialsFirst
 credentialProviderPluginBinDir: /etc/kubernetes/image-credential-provider/
 credentialProviders:
   apiVersion: kubelet.config.k8s.io/v1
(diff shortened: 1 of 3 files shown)


dkoshkin:
We should update this in KIB first, otherwise the air-gapped tests using CAREN will fail.

jimmidyson:
@dkoshkin 

> We should update this in KIB first, otherwise the air-gapped tests using CAREN will fail.

We don't have air-gapped tests in CAREN and I assume the air-gapped tests you're talking about use a released version of CAREN, so let's merge this here and then when we release CAREN we can update KIB and the air-gapped tests you're talking about.

supershal:
I think @dkoshkin is referring to existing konvoy2 airgap tests that uses KIB images. We don't have env setup for automated airgap e2e tests yet. I think this PR can be merged in parallel. 
PR for KIB. https://github.com/mesosphere/konvoy-image-builder/pull/1084

Konvoy still uses `ghcr.io/mesosphere/dynamic-credential-provider:v0.2.0`.  Do we know why we havent bumped it there yet? or it needs to bumped there as well?

supershal (approved):

jimmidyson:
@supershal 

> Konvoy still uses ghcr.io/mesosphere/dynamic-credential-provider:v0.2.0. Do we know why we havent bumped it there yet? or it needs to bumped there as well?

No idea, yes it needs to be bumped there as it will include newer credential providers that support the latest kubelet credential provider APIs.

dkoshkin:
> No idea, yes it needs to be bumped there as it will include newer credential providers that support the latest kubelet credential provider APIs.

Konvoy has not migrated the API version yet.

dkoshkin:
Lets talk about it offline first before merging please. This coordination that we need across 3 repos is not great.

dkoshkin:
I was thinking about this some more, we can merge this and rebuild the AMIs in KIB with the same change which will be with v1.29.4 Kubernetes version and then use the new AMIs in Konvoy e2e tests.

dkoshkin (approved):

PR #678, opened 2024-05-20
build: Update all tools

jimmidyson:

Weekly tools update.

Fixes #

### DIFF
--- a/hack/flakes/flake.nix
+++ b/hack/flakes/flake.nix
@@ -14,16 +14,16 @@
 
           go-mod-upgrade = buildGo122Module rec {
             name = "go-mod-upgrade";
-            version = "0.9.1";
+            version = "0.10.0";
             src = fetchFromGitHub {
               owner = "oligot";
               repo = "go-mod-upgrade";
               rev = "v${version}";
-              hash = "sha256-+C0IMb7MU1fq/P0/tTUNmzznZ1q5M69491pO5yBZlVs=";
+              hash = "sha256-BuHyqv0rK1giNiPO+eCx13rJ9L6y2oCDdKW1sJXyFg4=";
             };
             doCheck = false;
             subPackages = [ "." ];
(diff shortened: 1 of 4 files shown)


hack/flakes/flake.nix
```diff
@@ -14,16 +14,16 @@
...
             name = "go-mod-upgrade";
-            version = "0.9.1";
+            version = "0.10.0";
             src = fetchFromGitHub {
               owner = "oligot";
               repo = "go-mod-upgrade";
               rev = "v${version}";
-              hash = "sha256-+C0IMb7MU1fq/P0/tTUNmzznZ1q5M69491pO5yBZlVs=";
+              hash = "sha256-BuHyqv0rK1giNiPO+eCx13rJ9L6y2oCDdKW1sJXyFg4=";
             };
```
faiq, line 23: can you tell us the command to upgrage this hash. i just use the wrong one until its right

faiq (approved):

hack/flakes/flake.nix
jimmidyson: 😂 that's what i do. I think there must be a command to calculate it, but what I end up doing is replacing this hash and vendor hash with `lib.fakeHash`, running the devbox install, and then updating the hashes after that. It's not pretty...

jimmidyson: Here ya go:

```
$ nix hash to-sri --type sha256 $(nix-prefetch-url --unpack https://github.com/oligot/go-mod-upgrade/archive/v0.10.0.tar.gz 2>&/dev/null)

sha256-BuHyqv0rK1giNiPO+eCx13rJ9L6y2oCDdKW1sJXyFg4=
```

 Not sure how to do it for vendorHash though.

jimmidyson: Official advice from https://ryantm.github.io/nixpkgs/languages-frameworks/go/#ex-buildGoModule:

> To obtain the actual hash, set vendorHash = lib.fakeSha256; and run the build

dlipovetsky (approved):

PR #679, opened 2024-05-21
build: Correct dry-run output

jimmidyson:
The existing `if` condition will always evaluate to true because it is
checking if `E2E_DRYRUN` has _any_ content. Change this to check that
`E2E_DRYRUN` is set to `true` before outputting `dry-` in this place.
This is already done correctly in other places, so just tidying up a
typo from earlier PR.


dkoshkin (approved):

dlipovetsky (approved):
Thanks.

make functions :sob:

PR #682, opened 2024-05-21
feat: implements BeforeClusterUpgrade hook

faiq:

This implements BeforeClusterUpgrade hook and executes the same logic. There's some common code that I'd like ideas on how to consolidate while also having the handlers be separate lifecycle hooks.

Fixes #


common/pkg/capi/clustertopology/handlers/lifecycle/interfaces.go
```diff
@@ -9,6 +9,11 @@ import (
 	runtimehooksv1 "sigs.k8s.io/cluster-api/exp/runtime/hooks/api/v1alpha1"
 )
 
+const (
+	// LifecycleRetryAfterSeconds is used for BeforeClusterUpgrade hooks in case of failure.
+	LifecycleRetryAfterSeconds = 30
```
dlipovetsky, line 14: Since the package is `lifecycle`, we  can simplify the name to
```suggestion
	RetryAfterSeconds = 30
```

However, _if this is value is only for BeforeClusterUpgrade_, then please consider renaming it to
```suggestion
	BeforeClusterUpgradeRetryAfterSeconds = 30
```

dlipovetsky:
> > Any reason that the existing lifecycle hooks do not retry? 🤔
> 
> They do retry as part of the reconciliation loop. The BeforeClusterUpgrade hooks just allow us to set a specific amount of time to retry after.

Thanks for explaining that!

common/pkg/server/server.go
```diff
@@ -94,9 +94,9 @@ func (s *Server) Start(ctx context.Context) error {
 		}
 
 		if t, ok := h.(lifecycle.AfterControlPlaneInitialized); ok {
-			if err := webhookServer.AddExtensionHandler(server.ExtensionHandler{
+			if err := webhookServer.AddExtensionHandler(runtimeserver.ExtensionHandler{
 				Hook:        runtimehooksv1.AfterControlPlaneInitialized,
-				Name:        strings.ToLower(h.Name()),
+				Name:        strings.ToLower(h.Name()) + "aftercontrolplaneinitialized",
```
jimmidyson, line 99: I wonder if we can shorten this suffix? And add a `-` separator
```suggestion
				Name:        strings.ToLower(h.Name()) + "-acpi",
```

common/pkg/server/server.go
```diff
@@ -105,9 +105,9 @@ func (s *Server) Start(ctx context.Context) error {
 		}
 
 		if t, ok := h.(lifecycle.BeforeClusterUpgrade); ok {
-			if err := webhookServer.AddExtensionHandler(server.ExtensionHandler{
+			if err := webhookServer.AddExtensionHandler(runtimeserver.ExtensionHandler{
 				Hook:        runtimehooksv1.BeforeClusterUpgrade,
-				Name:        strings.ToLower(h.Name()),
+				Name:        strings.ToLower(h.Name()) + "beforeclusterupgrade",
```
jimmidyson, line 110:
```suggestion
				Name:        strings.ToLower(h.Name()) + "-bcu",
```

common/pkg/server/server.go
```diff
@@ -83,7 +83,7 @@ func (s *Server) Start(ctx context.Context) error {
 		h := s.hooks[idx]
 
 		if t, ok := h.(lifecycle.BeforeClusterCreate); ok {
-			if err := webhookServer.AddExtensionHandler(server.ExtensionHandler{
+			if err := webhookServer.AddExtensionHandler(runtimeserver.ExtensionHandler{
 				Hook:        runtimehooksv1.BeforeClusterCreate,
 				Name:        strings.ToLower(h.Name()),
```
jimmidyson, line 88: We will need to add suffixes to every handler type
```suggestion
				Name:        strings.ToLower(h.Name()+"-bcc"),
```

common/pkg/server/server.go
```diff
@@ -116,7 +116,7 @@ func (s *Server) Start(ctx context.Context) error {
 		}
 
 		if t, ok := h.(lifecycle.AfterControlPlaneUpgrade); ok {
-			if err := webhookServer.AddExtensionHandler(server.ExtensionHandler{
+			if err := webhookServer.AddExtensionHandler(runtimeserver.ExtensionHandler{
 				Hook:        runtimehooksv1.AfterControlPlaneUpgrade,
 				Name:        h.Name(),
```
jimmidyson, line 121:
```suggestion
				Name:        strings.ToLower(h.Name()+"-acpu"),
```

common/pkg/server/server.go
```diff
@@ -127,7 +127,7 @@ func (s *Server) Start(ctx context.Context) error {
 		}
 
 		if t, ok := h.(lifecycle.BeforeClusterDelete); ok {
-			if err := webhookServer.AddExtensionHandler(server.ExtensionHandler{
+			if err := webhookServer.AddExtensionHandler(runtimeserver.ExtensionHandler{
 				Hook:        runtimehooksv1.BeforeClusterDelete,
 				Name:        strings.ToLower(h.Name()),
```
jimmidyson, line 132:
```suggestion
				Name:        strings.ToLower(h.Name()+"-bcd"),
```

common/pkg/server/server.go
```diff
@@ -138,7 +138,7 @@ func (s *Server) Start(ctx context.Context) error {
 		}
 
 		if t, ok := h.(mutation.DiscoverVariables); ok {
-			if err := webhookServer.AddExtensionHandler(server.ExtensionHandler{
+			if err := webhookServer.AddExtensionHandler(runtimeserver.ExtensionHandler{
 				Hook:        runtimehooksv1.DiscoverVariables,
 				Name:        strings.ToLower(h.Name()),
```
jimmidyson, line 143:
```suggestion
				Name:        strings.ToLower(h.Name()+"-dv"),
```

common/pkg/server/server.go
```diff
@@ -149,7 +149,7 @@ func (s *Server) Start(ctx context.Context) error {
 		}
 
 		if t, ok := h.(mutation.GeneratePatches); ok {
-			if err := webhookServer.AddExtensionHandler(server.ExtensionHandler{
+			if err := webhookServer.AddExtensionHandler(runtimeserver.ExtensionHandler{
 				Hook:        runtimehooksv1.GeneratePatches,
 				Name:        strings.ToLower(h.Name()),
```
jimmidyson, line 154:
```suggestion
				Name:        strings.ToLower(h.Name()+"-gp"),
```

common/pkg/server/server.go
```diff
@@ -160,7 +160,7 @@ func (s *Server) Start(ctx context.Context) error {
 		}
 
 		if t, ok := h.(mutation.ValidateTopology); ok {
-			if err := webhookServer.AddExtensionHandler(server.ExtensionHandler{
+			if err := webhookServer.AddExtensionHandler(runtimeserver.ExtensionHandler{
 				Hook:        runtimehooksv1.ValidateTopology,
 				Name:        strings.ToLower(h.Name()),
```
jimmidyson, line 165:
```suggestion
				Name:        strings.ToLower(h.Name()+"-vt"),
```

common/pkg/capi/clustertopology/handlers/lifecycle/interfaces.go
```diff
@@ -9,6 +9,11 @@ import (
 	runtimehooksv1 "sigs.k8s.io/cluster-api/exp/runtime/hooks/api/v1alpha1"
 )
 
+const (
+	// BeforeClusterUpgradeRetryAfterSeconds is used for BeforeClusterUpgrade hooks in case of failure.
+	BeforeClusterUpgradeRetryAfterSeconds = 30
```
jimmidyson, line 14: This is not just in case of failures, but also when waiting for addons to be upgraded before continuing with cluster upgrade. `BeforeClusterUpgrade` hooks can block upgrade from continuing and will be retried until long running operations have completed without this being an error.

For example, might want to wait until CNI has been successfully upgraded before doing the Kubernetes upgrade.

pkg/handlers/generic/lifecycle/ccm/handler.go
```diff
@@ -67,14 +68,39 @@ func (c *CCMHandler) AfterControlPlaneInitialized(
...
+	req *runtimehooksv1.BeforeClusterUpgradeRequest,
+	resp *runtimehooksv1.BeforeClusterUpgradeResponse,
+) {
+	commonResponse := &runtimehooksv1.CommonResponse{}
+	c.apply(ctx, &req.Cluster, commonResponse)
+	resp.Status = commonResponse.GetStatus()
+	resp.Message = commonResponse.GetMessage()
+	if resp.Status == runtimehooksv1.ResponseStatusFailure {
+		resp.SetRetryAfterSeconds(lifecycle.BeforeClusterUpgradeRetryAfterSeconds)
+	}
```
jimmidyson, line 88: Retry can be set even with success response, and should be set in the case of waiting for addons to actually be successfully deployed and running. I think we'll need a different signature for apply, something like:

```
func (c *CCMHandler) apply(
	ctx context.Context,
	cluster *clusterv1.Cluster,
	resp *runtimehooksv1.CommonResponse,
) (completed bool, err error) // No need for named returns, just here for clarity
```

The `BeforeClusterUpgrade` hook can set the retry time if completed is false, and `AfterInitializedControlPlane` can ignore `completed`.

In the `apply` function, the logic will have to check for the existence and status of the applied resources and preferably the status of the actual deployed resources in workload cluster to make the decision about whether it has `completed` or not.

common/pkg/server/server.go
faiq: sounds great to me. i just did this to see if this would solve the issue that i was running into.

pkg/handlers/generic/lifecycle/ccm/handler.go
faiq: That sounds great to me. 

thanks for the explanation.

jimmidyson (dismissed):

jimmidyson:
Rebased and fixed conflicts after merge of #637.

jimmidyson (approved):

pkg/handlers/generic/mutation/imageregistries/credentials/inject_test.go
```diff
@@ -354,7 +354,10 @@ var _ = Describe("Generate Image registry patches", func() {
 						},
 					),
 				},
-				RequestItem: request.NewKubeadmConfigTemplateRequest("", "test-kubeadmconfigtemplate"),
+				RequestItem: request.NewKubeadmConfigTemplateRequest(
+					"",
+					"test-kubeadmconfigtemplate",
+				),
```
dlipovetsky, line 360: (nit: Was this an automatic formatting change? I find it distracting, because this file is unrelated to the PR)

jimmidyson: Yes it was. Need to figure out why these changes aren't getting picked up on original pr.

jimmidyson (approved):

jimmidyson (approved):

PR #684, opened 2024-05-21
feat: Added FailureDomains as Nutanix cluster mutation

deepakm-ntnx:
This PR allows us to configure failureDomains and allows use it for control plane and worker nodes to create a capx cluster on nutanix infra.

Fixes # NCN-100421

Unit tests passed.
<pre>
make test
</pre>

Manual test passed.
<pre>
make dev.run-on-kind
clusterctl generate cluster nutanix-quick-start-helm-addon-cilium \
  --from examples/capi-quick-start/nutanix-cluster-cilium-helm-addon.yaml \
  --kubernetes-version v1.30.0 \
  --worker-machine-count 1 > c2.yaml
# update c2.yaml to have failureDomains added and set respective MD property. more info in docs
kubectl apply --server-side -f ./c2.yaml
</pre>
Then observed that respective control plane and worker nodes are created on respective FD


faiq (approved):
looks good from a high level

api/v1alpha1/nutanix_clusterconfig_types.go
```diff
@@ -76,3 +84,39 @@ func (s NutanixPrismCentralEndpointSpec) ParseURL() (string, int32, error) {
 
 	return hostname, int32(port), nil
 }
+
+// NutanixFailureDomains is a list of FDs.
+type NutanixFailureDomains []NutanixFailureDomain
+
+// NutanixFailureDomain configures failure domain information for Nutanix.
+type NutanixFailureDomain struct {
```
dkoshkin, line 92: Is there some validation we want to do with PE Clusters and Subnets, like setting either this or those values there?

dkoshkin, line 92: Do we expect clients to set the FD name in MD to use these, maybe some docs on how to use this would be good?

hack/examples/bases/nutanix/cluster/kustomization.yaml.tmpl
```diff
@@ -7,7 +7,7 @@ kind: Kustomization
 resources:
 - ../../../additional-resources/dockerhub-secret.yaml
 - ../../../additional-resources/nutanix/csi-secret.yaml
-- https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/releases/download/${CAPX_VERSION}/cluster-template-topology.yaml
```
dkoshkin, line 10: I would really prefer if we had a CAPX release with the changes

api/v1alpha1/nutanix_clusterconfig_types.go
```diff
@@ -76,3 +84,39 @@ func (s NutanixPrismCentralEndpointSpec) ParseURL() (string, int32, error) {
...
+	// +kubebuilder:validation:Required
+	Cluster NutanixResourceIdentifier `json:"cluster"`
+
+	// subnets holds a list of identifiers (one or more) of the cluster's network subnets
+	// for the Machine's VM to connect to. The subnet identifiers (uuid or name) can be
+	// obtained from the Prism Central console or using the prism_central API.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinItems=1
+	// +listType=map
+	// +listMapKey=type
```
dkoshkin, line 116: IIUC we want the name to be the unique key?
```suggestion
	// +listMapKey=name
```

api/v1alpha1/nutanix_clusterconfig_types.go
```diff
@@ -76,3 +84,39 @@ func (s NutanixPrismCentralEndpointSpec) ParseURL() (string, int32, error) {
...
+	Subnets []NutanixResourceIdentifier `json:"subnets"`
+
+	// indicates if a failure domain is suited for control plane nodes
+	// +kubebuilder:validation:Required
+	ControlPlane bool `json:"controlPlane,omitempty"`
```
dkoshkin, line 121: Is it valid to have all FD with this value set to `false`?

api/v1alpha1/nutanix_clusterconfig_types.go
deepakm-ntnx: capx handles it by checking fd set or not first. https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/blob/v1.4.0-alpha.2/controllers/nutanixmachine_controller.go#L834 need to see if we need to handle that in caren since caren does not have any notion of sequence of which patch will get called first. do you have any proposals?

api/v1alpha1/nutanix_clusterconfig_types.go
deepakm-ntnx: user can provide uuid or name. hence type is the only fields which will be available. also this is a list so why does it need to be unique?

api/v1alpha1/nutanix_clusterconfig_types.go
deepakm-ntnx: Yes, it is valid in backend but we can have UI/CLI make it mandatory to provide one fd with this to be true. Ideally it will create controlPlane on provided cluster and subnet in the respective machine template if no fd has ControlPlane is true. if nmt does not have cluster and subnet provided then it will fail with this error https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/blob/v1.4.0-alpha.2/controllers/nutanixmachine_controller.go#L836.

api/v1alpha1/nutanix_clusterconfig_types.go
deepakm-ntnx: yes we have a internal tracker for opendocs for adding capx docs. but we can also update caren docs.

deepakm-ntnx: docs added.

api/v1alpha1/nutanix_clusterconfig_types.go
dkoshkin: The value of from `listMapKey` is used when doing server side applies https://kubernetes.io/docs/reference/using-api/server-side-apply/#merge-strategy, IIUC having it `listMapKey= type` would be that any (or some) elements that has the same type would be replaced instead of appended?

We probably don't want these at all here.

jimmidyson: `listMapKey` isn't supported for ClusterClass variables iirc.

deepakm-ntnx: ok, will file a tracker to fix in capx as well https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/blob/v1.4.0-alpha.2/api/v1beta1/nutanixcluster_types.go#L132

jimmidyson: @deepakm-ntnx That's different because that is a real CRD and as such it is supported by the API server.

jimmidyson: Having said that, we should definitely re-evaluate what the key means in this context in CAPX CRDs at some point.

deepakm-ntnx:
Just a note: to be debugged
per NCN-100947,
- case 1:  if we create a cluster with failureDomains while creating the cluster then vms do get created on the respective FD
- case 2: if we create a cluster first without failureDomains and later add FD with this PR changes, then somehow VMs are not getting created on the respective FD
Will continue debugging

deepakm-ntnx:
I would recommend to approve and merge this PR so that we can allow users to enable them manually if needed in the field if they need control planes on multiple PEs in 2.12. since its option, we dont have to expose it in the documentation

dkoshkin:
The failure domain API changing, this PR will need to be reworked.
If we do decide to use the existing API for some reason, we can reopen this PR.

PR #693, opened 2024-05-30
feat: Add local-path-provisioner CSI

jimmidyson:
This enables easy testing with Docker clusters as well, but this should
not be used in production.


deepakm-ntnx (approved):
what do you plan to test with local-path csi on docker provider in general. is there a stateful app test or do you want to test csi installation workflow?

faiq (approved):

jimmidyson:
@deepakm-ntnx 

> what do you plan to test with local-path csi on docker provider

First thing I used this for was the hosted registry work I've been doing. Without this it was hard to switch to a registry backed by CSI storage. Now I have a generic way to switch to a persistent registry backed by default storage class which works across providers.

PR #694, opened 2024-05-31
fix: Remove resource name stutter in default ClusterClasses

dlipovetsky:
1. Removes sutter from resource names. The KubeadmControlPlaneTemplate for the aws-quick-start ClusterClass  is renamed from `aws-quick-start-control-plane` to `aws-quick-start`, because its Kind already makes it clear that it's for the control plane, and only one such resource exists.
2. Follows one convention for resource names in all default ClusterClasses. For example, the workers' NutanixMachineTemplate for the nutanix-quick-start ClusterClass is renamed from `nutanix-quick-start-md-nmt` to `nutanix-quick-start-workers`, because its `nmt` repeats the information in the Kind, and `-workers` clearly communicates that the resource is for workers, whereas `-md` does not.

Fixes #

### DIFF
--- a/charts/cluster-api-runtime-extensions-nutanix/defaultclusterclasses/nutanix-cluster-class.yaml
+++ b/charts/cluster-api-runtime-extensions-nutanix/defaultclusterclasses/nutanix-cluster-class.yaml
@@ -3,7 +3,7 @@ kind: KubeadmConfigTemplate
 metadata:
   labels:
     cluster.x-k8s.io/provider: nutanix
-  name: nutanix-quick-start-kcfg-0
+  name: nutanix-quick-start
 spec:
   template:
     spec:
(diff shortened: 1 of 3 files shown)


charts/cluster-api-runtime-extensions-nutanix/defaultclusterclasses/nutanix-cluster-class.yaml
```diff
@@ -3,7 +3,7 @@ kind: KubeadmConfigTemplate
 metadata:
   labels:
     cluster.x-k8s.io/provider: nutanix
-  name: nutanix-quick-start-kcfg-0
+  name: nutanix-quick-start
```
deepakm-ntnx, line 6: will this get overwritten when we do <pre>
make apis.sync
</pre>
ref https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/blob/main/templates/cluster-template-clusterclass.yaml#L4

dlipovetsky: Good point. That won't do it, but the related `make examples.sync` will.

I guess I can work around this in our kustomization, but also fix it upstream.

PR #695, opened 2024-06-02
fix: Post-process clusterconfig CRDs for supported CSI providers

jimmidyson:
Only supported CSI providers should be available on each CAPI provider. This
commit enforces that via the API, without changing handler logic to allow
for easier extensibility later, by simply updating the supported CSI provider
names in the API definitions.

This can only happen as CRD post-processing as kubebuilder annotations are
specified on API types in go. To use that approach for CSI providers here
would require changing the whole API structure to provide CAPI provider
specific types with CAPI provider specific annotations, which would be a
maintenance headache.

This issue arose for me when adding the `local-path` CSI provider for Docker
as I did not want this to be available on AWS or Nutanix clusters, as well
as the opposite way round for Docker clusters (no `aws-ebs` or `nutanix`
support).

### DIFF
--- a/make/addons.mk
+++ b/make/addons.mk
@@ -24,6 +24,10 @@ export KUBE_VIP_VERSION := v0.8.0
 
 export METALLB_CHART_VERSION := v0.14.5
 
+CSI_PROVIDERS_aws := ["aws-ebs"]
+CSI_PROVIDERS_nutanix := ["nutanix"]
+CSI_PROVIDERS_docker := ["local-path"]
+
 .PHONY: addons.sync
 addons.sync: $(addprefix update-addon.,calico cilium nfd cluster-autoscaler aws-ebs-csi aws-ccm.127 aws-ccm.128 aws-ccm.129 kube-vip)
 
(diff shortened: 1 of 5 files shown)


make/addons.mk
```diff
@@ -24,6 +24,10 @@ export KUBE_VIP_VERSION := v0.8.0
 
 export METALLB_CHART_VERSION := v0.14.5
 
+CSI_PROVIDERS_aws := ["aws-ebs"]
```
dlipovetsky, line 27:
```suggestion
# Below are the lists of CSI Providers allowed for a specific infrastructure.
# - When we support a new infrastructure, we need to a create a new list using the same convention.
# - When we support a new CSI Provider, we need to add it to one or more of these lists
CSI_PROVIDERS_aws := ["aws-ebs"]
```

make/addons.mk
```diff
@@ -59,3 +66,13 @@ update-addon.kube-vip: ; $(info $(M) updating kube-vip manifests)
 generate-helm-configmap:
 	go run hack/tools/helm-cm/main.go -kustomize-directory="./hack/addons/kustomize" -output-file="./charts/cluster-api-runtime-extensions-nutanix/templates/helm-config.yaml"
 	./hack/addons/add-warning-helm-configmap.sh
+
+# Set only the supported CSI providers for each provider.
+.PHONY: configure-csi-providers.%
```
faiq, line 71: nice - can we move all the scripting to a file and call it like the rest of this file does? it helps keep things neat

dkoshkin (approved):

make/addons.mk
jimmidyson: Done!

PR #698, opened 2024-06-03
fix: use external Nutanix API types directly

dkoshkin:
From what I can tell there is no longer a need to have these separate types.

These types get used by other projects where it generates a swagger spec. This change also fixes a swagger issue complaining about a circular reference since the types are named the same.

Fixes #


supershal (approved):

dlipovetsky:
The type aliases allow Go consumers to import only the CAREN package, so they are convenient.

I don't completey understand the swagger issue.

That said, I'm ok with this change. We can easily use type aliases again in the future, if we resolve the swagger issue.

faiq (approved):

dlipovetsky (approved):

PR #699, opened 2024-06-07
refactor: kube-vip commands

dkoshkin:
This PR refactors the kube-vip handler to:
1. Use a script file when running preKubeadmCommands/postKubeadmCommands. IMO this is simpler to maintain and change in the future instead of looking for complex multi-line strings in the slices.
2. Move `echo "127.0.0.1   kubernetes" >>/etc/hosts` out of the template and into the handler since this command is not unique to Nutanix and would be needed for other infra providers using kube-vip.

Fixes #

### DIFF
--- a/charts/cluster-api-runtime-extensions-nutanix/defaultclusterclasses/nutanix-cluster-class.yaml
+++ b/charts/cluster-api-runtime-extensions-nutanix/defaultclusterclasses/nutanix-cluster-class.yaml
@@ -153,7 +153,6 @@ spec:
         - hostnamectl set-hostname "{{ ds.meta_data.hostname }}"
         - echo "::1         ipv6-localhost ipv6-loopback" >/etc/hosts
         - echo "127.0.0.1   localhost" >>/etc/hosts
-        - echo "127.0.0.1   kubernetes" >>/etc/hosts
         - echo "127.0.0.1   {{ ds.meta_data.hostname }}" >> /etc/hosts
         useExperimentalRetryJoin: true
         verbosity: 10
--- a/pkg/handlers/generic/mutation/controlplanevirtualip/providers/templates/configure-for-kube-vip.sh
+++ b/pkg/handlers/generic/mutation/controlplanevirtualip/providers/templates/configure-for-kube-vip.sh
@@ -22,9 +22,13 @@ function use_admin_conf() {
   fi
 }
 
+function set_host_aliases() {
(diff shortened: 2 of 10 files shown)


pkg/handlers/generic/mutation/controlplanevirtualip/providers/templates/configure-for-kube-vip.sh
```diff
@@ -0,0 +1,67 @@
...
+  fi
+}
+
+function set_host_aliases() {
+  echo "127.0.0.1   kubernetes" >>/etc/hosts
+}
+
+function print_usage {
+  cat >&2 <<EOF
+  Usage: ${SCRIPT_NAME} [use-super-admin.conf|use-admin.conf|set-host-aliases]
```
dkoshkin, line 31: Looking for suggestions on how we want to structure this script.
Initially I had it as `prekubeadm|postkubeadm` but thats not very descriptive of what it is actually doing.
On the flip side, having individual commands has the potential of adding more pre/postKubeadmCommands that are just doing simple single line commands.

jimmidyson: I actually think this is a clean structure already! I think that you can also run `./configure-for-kube-vip.sh set-host-aliases use-super-admin.conf` btw.

jimmidyson (approved):

thunderboltsid (approved):
/lgtm

charts/cluster-api-runtime-extensions-nutanix/defaultclusterclasses/nutanix-cluster-class.yaml
```diff
@@ -153,7 +153,6 @@ spec:
         - hostnamectl set-hostname "{{ ds.meta_data.hostname }}"
         - echo "::1         ipv6-localhost ipv6-loopback" >/etc/hosts
         - echo "127.0.0.1   localhost" >>/etc/hosts
-        - echo "127.0.0.1   kubernetes" >>/etc/hosts
```
faiq, line 156: can we remove the debug echos from above

dkoshkin: Yes, I will do that in a separate PR

jimmidyson (approved):

PR #709, opened 2024-06-10
fix: credential provider response secret ownership

dkoshkin:
Saw that the owner was not getting correctly set on image credential Secret.

Fixes #

### DIFF
--- a/pkg/handlers/generic/mutation/imageregistries/credentials/inject.go
+++ b/pkg/handlers/generic/mutation/imageregistries/credentials/inject.go
@@ -368,15 +368,14 @@ func createSecretIfNeeded(
 		)
 	}
 	if credentialsSecret != nil {
-		if err := client.ServerSideApply(ctx, c, credentialsSecret, client.ForceOwnership); err != nil {
-			return fmt.Errorf("failed to apply Image Registry Credentials Secret: %w", err)
-		}
-
 		if err = controllerutil.SetOwnerReference(cluster, credentialsSecret, c.Scheme()); err != nil {
 			return fmt.Errorf(
 				"failed to set owner reference on Image Registry Credentials Secret: %w",
 				err,
 			)
+		}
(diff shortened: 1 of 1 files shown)


jimmidyson:
Can we add an e2e test that includes creating this kind of secret? This should automatically check ownership

dkoshkin:
> Can we add an e2e test that includes creating this kind of secret? This should automatically check ownership

Good idea, any hints how I can do that without adding some dummy registry into the examples?

supershal (approved):
Thanks for fixing this. can we create a unit test for this? it can be separate PR / GH issue/Jira.

jimmidyson:
@dkoshkin 

> any hints how I can do that without adding some dummy registry into the examples?

Ooo good question! Maybe this is easier with a envtest based test instead of our e2es?

dkoshkin:
Adding a unit test is turning out way more complicated than it needs to be.
Using the fakeclient it fails creating the Secret because of https://github.com/kubernetes/kubernetes/issues/115598. I thought that I can pre-create an empty Secret as a workaround but its still giving me this error:
```
Received unexpected error: failed to apply Image Registry Credentials Secret: server-side apply failed: apply patches are not supported in the fake client. Follow https://github.com/kubernetes/kubernetes/issues/115598 for the current status
```

I tried writing a Ginkgo test and using envtest but thats also failing because of strange error
```
failed to apply Image Registry Credentials Secret: server-side apply failed: the server could not find the requested resource
```

Here is the commit https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/commit/1a8a1dff8430c0e1ae8f9403121e6952144a5034

If its ok, I would like to come back to the unit tests in a separate PR. I'm trying to clean up the owner refs for all these Secrets and there are still a few other ones.

jimmidyson:
@dkoshkin 

> If its ok, I would like to come back to the unit tests in a separate PR. I'm trying to clean up the owner refs for all these Secrets and there are still a few other ones.

Sure, let's do that!

jimmidyson (approved):

PR #710, opened 2024-06-10
refactor: rename credential provider response secret

dkoshkin:
Wanted to avoid "creds" abbreviation.
Also, the Konvoy secret names are `<cluster-name>-image-registry-credentials` and `<cluster-name>-image-registry-mirror-credentials` so it was pretty confusing also having a `<cluster-name>-registry-creds`. Wanted to make it a little more descriptive. (Suggestions welcome!)
Here is what that Secret looks like:
```
{
  "kind":"CredentialProviderResponse",
  "apiVersion":"credentialprovider.kubelet.k8s.io/v1",
  "cacheKeyType":"Image",
  "cacheDuration":"0s",
  "auth":{
    "docker.io": {"username": "<>", "password": "<>"}
  }
}
```

Fixes #

### DIFF
--- a/pkg/handlers/generic/mutation/imageregistries/credentials/credentials_secret.go
+++ b/pkg/handlers/generic/mutation/imageregistries/credentials/credentials_secret.go
@@ -114,5 +114,5 @@ func kubeletStaticCredentialProviderSecretContents(configs []providerConfig) (st
 }
 
 func credentialSecretName(clusterName string) string {
-	return fmt.Sprintf("%s-registry-creds", clusterName)
+	return fmt.Sprintf("%s-credential-provider-response", clusterName)
 }
--- a/pkg/handlers/generic/mutation/imageregistries/credentials/inject_test.go
+++ b/pkg/handlers/generic/mutation/imageregistries/credentials/inject_test.go
@@ -27,8 +27,7 @@ import (
 )
 
 const (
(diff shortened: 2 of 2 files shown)


supershal (approved):
Descriptive name is good.

pkg/handlers/generic/mutation/imageregistries/credentials/credentials_secret.go
```diff
@@ -114,5 +114,5 @@ func kubeletStaticCredentialProviderSecretContents(configs []providerConfig) (st
 }
 
 func credentialSecretName(clusterName string) string {
-	return fmt.Sprintf("%s-registry-creds", clusterName)
+	return fmt.Sprintf("%s-credential-provider-response", clusterName)
```
jimmidyson, line 117: I was a little concerned about the length of this (even before the change tbh) but I looked into it and all is good here. Secret name is maximum 253 characters ([DNS subdomain name](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/validation/validation.go#L282)) and Cluster name is maximum 63 characters ([label name](https://github.com/kubernetes-sigs/cluster-api/blob/main/internal/webhooks/cluster.go#L167)).

jimmidyson (approved):

PR #714, opened 2024-06-11
feat: cluster-api v1.7.3

jimmidyson:

Fixes #


supershal (approved):

dkoshkin (approved):
This is an interesting change https://github.com/kubernetes-sigs/cluster-api/pull/10701/files, do want to use this new field and set a controlPlane variable instead of having it inside the `clusterConfig`?

thunderboltsid (approved):

jimmidyson:
@dkoshkin 

> This is an interesting change [kubernetes-sigs/cluster-api#10701 (files)](https://github.com/kubernetes-sigs/cluster-api/pull/10701/files), do want to use this new field and set a controlPlane variable instead of having it inside the `clusterConfig`?

Hmm maybe. I'm not quite sure yet how useful this is because there is only one control plane configuration, compared to multiple nodepools which could each have different configurations where overrides makes more sense. Maybe I'm missing the use cases for this 🤔

PR #715, opened 2024-06-11
feat: Controller that copies ClusterClasses to namespaces

dlipovetsky:
A Cluster must be in the same namespace as the ClusterClass it references (see https://github.com/kubernetes-sigs/cluster-api/issues/5673).

This PR adds a controller that copies any ClusterClass (and the Templates it references) found in a source namespace (`default` by default) to every "target" namespace, i.e. a namespace that has a well-defined label key (`caren.nutanix.com/namespace-sync`). The controller does not update or delete the resources in a target namespace.

Fixes #

This PR adds envtest-based integration tests that verify that the controller:
1. Copies ClusterClasses to a newly-created target namespace.
2. Copies a newly-created ClusterClass to target namespaces.
3. Does not update resources in a target namespace.

This PR includes a modified copy of the Cluster API internal test framework.
The namespacesync controller reads and writes Templates of various types. The
upstream test package creates "generic" Template types and CRDs. This allows us
to test the controller using envtest, with real types and CRDs.

## TODO
- [ ] Expose "source" namespace name in the helm Chart values 
- [ ] Decide the label key to use for target namespaces (right now, it's `caren.nutanix.com/namespace-sync`)


dlipovetsky:
Black duck check fails due to issue fixed in #716.

internal/test/envtest/external.go
```diff
@@ -0,0 +1,53 @@
+package envtest
+
+import (
+	"fmt"
+	"os/exec"
+	"path/filepath"
+	"strings"
+)
+
+func ExternalCRDDirectoryPaths() []string {
```
jimmidyson, line 10: We already have something similar in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/main/test/helpers/envtest.go#L162 - can we reuse this instead?

jimmidyson:
I'm wondering if it would be simpler to just use one (or all) of the provider CRDs that are already present as dependencies?

dlipovetsky:
I did try to use the existing code in `test/helpers`, but ran into various issues. For example, that code will not configure a controller.

Reusing the existing code is still my plan, but I want to do that in a future PR, because this is this PR introduces a new use case (a controller), and I want some signal before I make changes to the existing code.

I also wanted to use the "generic" CRDs, because I don't want to tie the tests to a specific infra provider.

I'm going to ask that the upstream package to create the "generic" CRDs be moved out of internal, because it's useful to a project, like CAREN, that implements controllers that work with Templates.

internal/test/README.md
```diff
@@ -0,0 +1,17 @@
+<!--
+ Copyright 2024 Nutanix. All rights reserved.
+ SPDX-License-Identifier: Apache-2.0
+ -->
+
+# Test Framework
```
dkoshkin, line 6: Thanks for this README! How did you copy these, maybe a Makefile target to update these with the current CAPI version would make it easier to maitain?

pkg/controllers/namespacesync/references.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	obj, err := external.Get(ctx, cli, ref, ref.Namespace)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get %s %s/%s", ref.Kind, ref.Name, ref.Namespace)
+	}
+	return obj, nil
+}
+
+func walkReferences(
```
dkoshkin, line 53: Clean!

dkoshkin (approved):
Its a shame we need to import all the `internal/test` stuff, just left a nit comment about having some tooling to sync that.
Everything else looks great!

internal/test/README.md
dlipovetsky: The upstream framework initializes webhooks. I had to remove this imports and initialization code.

I don't think it makes sense to automatically update.

In the future, I'd like upstream  to publish `internal/builder`, so we can remove our copy, and I'll combine `internal/test` and the similar code we already use in `test/helpers`.

jimmidyson (approved):

dlipovetsky:
I will address the two TODOs in a separate PR:
1.Expose "source" namespace name in the helm Chart values
2. Decide the label key to use for target namespaces (right now, it's `caren.nutanix.com/namespace-sync`).

PR #717, opened 2024-06-12
fix: static credentials Secret generation

dkoshkin:
While working in this package I was confused why `createSecretIfNeeded` was called that since it always generated a Secret.
Turns out there was a bug in these files that it never checked if the passed in `configs []providerConfig` even have static credentials. Added some failing unit tests and fixed the logic to skip generating the Secret and the corresponding file if not needed.
While writing the tests, I also noticed that the templating was not generating valid JSON for multiple registries. To keep that separator logic simple I moved the `docker.io` alias into Go code.

Fixes #

New unit tests and brought up a Nutanix cluster with `docker.io` creds.

### DIFF
--- a/pkg/handlers/generic/mutation/imageregistries/credentials/credentials_secret.go
+++ b/pkg/handlers/generic/mutation/imageregistries/credentials/credentials_secret.go
@@ -156,5 +156,5 @@ func configsRequireStaticCredentials(configs []providerConfig) bool {
 }
 
 func credentialSecretName(clusterName string) string {
-	return fmt.Sprintf("%s-credential-provider-response", clusterName)
+	return fmt.Sprintf("%s-static-credential-provider-response", clusterName)
 }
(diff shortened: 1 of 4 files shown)


pkg/handlers/generic/mutation/imageregistries/credentials/credentials_secret.go
```diff
@@ -88,22 +86,53 @@ func kubeletStaticCredentialProviderSecretContents(configs []providerConfig) (st
 		RegistryHost string
 		Username     string
 		Password     string
+		Separator    string
 	}
 
-	inputs := make([]templateInput, 0, len(configs))
+	inputs := make([]templateInput, 0)
```
jimmidyson, line 92: nit: No need to pre-allocate now we don't know the necessary capacity
```suggestion
	var inputs []templateInput
```

jimmidyson (approved):

jimmidyson:
@dkoshkin Thanks for fixing this! I updated the templating a little to use range index for separator addition rather than do it in the code, all works (no changes to test for that). Hope you're ok with that change.

faiq (approved):

PR #722, opened 2024-06-17
fix: Allow Node Feature Discovery garbage collector to run on control-plane nodes

dlipovetsky:
With this change, all NFD components can run on a single node cluster.

Fixes #https://jira.nutanix.com/browse/NCN-100706

### DIFF
--- a/charts/cluster-api-runtime-extensions-nutanix/templates/nfd/manifests/helm-addon-installation.yaml
+++ b/charts/cluster-api-runtime-extensions-nutanix/templates/nfd/manifests/helm-addon-installation.yaml
@@ -27,4 +27,11 @@ data:
         - effect: NoSchedule
           key: node-role.kubernetes.io/control-plane
     ### <NFD-WORKER-CONF-END-DO-NOT-REMOVE>
+
+    gc:
+      tolerations:
+        - effect: NoSchedule
+          key: node-role.kubernetes.io/master
+        - effect: NoSchedule
+          key: node-role.kubernetes.io/control-plane
 {{- end -}}
--- a/charts/cluster-api-runtime-extensions-nutanix/templates/nfd/manifests/node-feature-discovery-configmap.yaml
+++ b/charts/cluster-api-runtime-extensions-nutanix/templates/nfd/manifests/node-feature-discovery-configmap.yaml
@@ -967,6 +967,11 @@ data:
           dnsPolicy: ClusterFirstWithHostNet
           securityContext: {}
           serviceAccountName: node-feature-discovery-gc
+          tolerations:
+          - effect: NoSchedule
+            key: node-role.kubernetes.io/master
+          - effect: NoSchedule
+            key: node-role.kubernetes.io/control-plane
 kind: ConfigMap
 metadata:
   creationTimestamp: null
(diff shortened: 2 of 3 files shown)


charts/cluster-api-runtime-extensions-nutanix/templates/nfd/manifests/node-feature-discovery-configmap.yaml
```diff
@@ -967,6 +967,11 @@ data:
           dnsPolicy: ClusterFirstWithHostNet
           securityContext: {}
           serviceAccountName: node-feature-discovery-gc
+          tolerations:
```
faiq, line 970: this is only going to happen when its deployed using cluster resource sets. we would also need to add this bit in the chart values with the helm addon provider.

faiq: copy these values to this file 

`charts/cluster-api-runtime-extensions-nutanix/templates/nfd/manifests/helm-addon-installation.yaml`

dlipovetsky: Got it. Thanks!

dlipovetsky: Done in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/722/commits/ad60625c904da207da7900862ac3c613bd7493a6

charts/cluster-api-runtime-extensions-nutanix/templates/nfd/manifests/helm-addon-installation.yaml
```diff
@@ -27,4 +27,11 @@ data:
         - effect: NoSchedule
           key: node-role.kubernetes.io/control-plane
     ### <NFD-WORKER-CONF-END-DO-NOT-REMOVE>
+
+    gc:
+      tolerations:
+        - effect: NoSchedule
+          key: node-role.kubernetes.io/master
```
supershal, line 34: `node-role.kubernetes.io/master` is deprecated. we should not expect it on the CAREN created cluster.

jimmidyson:
From https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/723#issuecomment-2174406269:

> I wonder if NFD should also tolerate the CriticalAddonsOnly taint, as many other Addons do?

From what I can see, NFD is deployed with an empty `priorityClassName` so this wouldn't have any effect right now as the `CriticalAddonsOnly` taint is only applied temporarily when pre-empting lower priority pods.

So I guess the next question is should we set the `priorityClassName` on the NFD addon deployments and then tolerate `CriticalAddonsOnly`? 😅 I don't see any harm in it, but is it really critical?

charts/cluster-api-runtime-extensions-nutanix/templates/nfd/manifests/helm-addon-installation.yaml
dlipovetsky: While that is true, we have 38 uses of that in CAREN. Let's replace it in a separate PR, please.

dlipovetsky:
> But we should also see if we need to add `priorityClassName: system-cluster-critical` to the NFD pods?

I created a separate JIRA to review this for all Addons.

jimmidyson (approved):

mhrabovcin (approved):

PR #723, opened 2024-06-17
fix: Allow Nutanix CSI snapshot controller & webhook to run on CP nodes

dlipovetsky:
With this change, all Nutanix CSI components can run on a single node cluster.

Fixes #https://jira.nutanix.com/browse/NCN-100706


dlipovetsky:
I wonder if NFD should also tolerate the `CriticalAddonsOnly` taint, as many other Addons do?

jimmidyson (approved):

jimmidyson:
> I wonder if NFD should also tolerate the `CriticalAddonsOnly` taint, as many other Addons do?

Let's discuss in #722.

mhrabovcin (approved):

PR #725, opened 2024-06-18
fix: Namespace Sync controller should list no resources when source namespace is empty string

dlipovetsky:
Previously, the controller listed resources in all namespaces when the source namespace was an empty string. Now, it lists no resources.

The PR adds a test verifying the above. It also adds a test that verifies that an the controller will not match any namespaces when the target namespace label key is an empty string. 

Fixes #

### DIFF
--- a/pkg/controllers/namespacesync/controller.go
+++ b/pkg/controllers/namespacesync/controller.go
@@ -150,6 +150,12 @@ func (r *Reconciler) listSourceClusterClasses(
 	[]clusterv1.ClusterClass,
 	error,
 ) {
+	// Handle the empty string explicitly, because listing resources with an empty
+	// string namespace returns resources in all namespaces.
+	if r.SourceClusterClassNamespace == "" {
+		return nil, nil
+	}
+
 	ccl := &clusterv1.ClusterClassList{}
 	err := r.Client.List(ctx, ccl, client.InNamespace(r.SourceClusterClassNamespace))
 	if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/controllers/namespacesync/controller.go
```diff
@@ -157,6 +150,12 @@ func (r *Reconciler) listSourceClusterClasses(
 	[]clusterv1.ClusterClass,
 	error,
 ) {
+	// Handle the empty string explicitly, because listing resources with an empty
+	// string namespace returns resources in all namespaces.
+	if r.SourceClusterClassNamespace == "" {
+		return []clusterv1.ClusterClass{}, nil
```
jimmidyson, line 156: nit
```suggestion
		return nil, nil
```

jimmidyson (commented):
Generally I wonder if we can exit early if the sync controller is enabled (perhaps add an enabled property in the helm chart to enable/disable this) and no namespace is specified (again, can error out in the helm chart if enabled but not specified too).

dlipovetsky:
> Generally I wonder if we can exit early if the sync controller is enabled (perhaps add an enabled property in the helm chart to enable/disable this) and no namespace is specified (again, can error out in the helm chart if enabled but not specified too).

Sure. I'll do this in a separate PR.

pkg/controllers/namespacesync/controller.go
dlipovetsky: Thanks

dlipovetsky:
e2e tests failing, will rebase on main once #736 merges

dlipovetsky:
These e2e passed locally. Not sure what the issue is in CI.

faiq (approved):
thanks for writing the test to verify this quickly

jimmidyson (approved):

PR #731, opened 2024-06-19
feat: Support CRS for local-path provisioner and add CSI e2e

jimmidyson:
Further strategy support for other providers coming in follow up PR,
plus more e2e tests for those other providers.

Blocked by #736.


faiq:
why do we need crs support for this?

jimmidyson:
> why do we need crs support for this?

This is more about adding consistency across addons/providers - what i'd really like to do is remove all CRS installations.

dkoshkin (approved):

dkoshkin (approved):
Tested locally and checked the statuses of ClusterResourceSets and HelmChartProxies and the Deployment on the cluster.

PR #735, opened 2024-06-20
refactor: Reusable HelmAddon strategy

jimmidyson:
This commit refactors some of the existing HelmAddon code used in various
handlers to a common applier, thus reducing code maintenance.

There are a number of other places where this could be applied, but they
currently have specific behaviour that is harder to generalise and as
such I have not refactored those handlers yet.

Depends on #734.


pkg/handlers/generic/lifecycle/appliers/interfaces.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"context"
+
+	"github.com/go-logr/logr"
+	"github.com/spf13/pflag"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+)
+
+type Applier interface {
```
faiq, line 14: consider a different name - this is redundant

supershal (approved):
Thanks for the improvement.

mhrabovcin (approved):

faiq (approved):

PR #740, opened 2024-06-21
feat: Add dynamic addon handler application via controller

jimmidyson:
This commit enables out of upgrade flow application of lifecycle hooks. This
enables dynamic updates of cluster variables that will result in application of
the lifecycle hook without requiring an upgrade to be in progress.


dkoshkin:
The design doc has diverged from this implementation, can we close this PR (but keep the branch around)?

cc @supershal @dlipovetsky

PR #741, opened 2024-06-21
fix: adds a docker buildx step before release-snapshot

faiq:

This should fix the local development flow for `make release-snapshot`.
Tested on amd64 and arm64.

We also remove unnecessary use of third-party github actions for docker

We now have a make target that creates a dockerx builder container
with multiarch support. Both the release and release-snapshot make
targets depend on this make target.

The two github actions are no longer necessary, and are removed.

Fixes #

### DIFF
--- a/make/goreleaser.mk
+++ b/make/goreleaser.mk
@@ -9,7 +9,7 @@ export GORELEASER_CURRENT_TAG=$(GIT_TAG)
 endif
 
 .PHONY: docker-buildx
-docker-buildx:
+docker-buildx: ## Creates buildx builder container that supports multiple platforms.
 docker-buildx:
 	 docker buildx create --use --name=caren --platform=linux/arm64,linux/amd64 || true
 
--- a/make/goreleaser.mk
+++ b/make/goreleaser.mk
@@ -8,6 +8,11 @@ ifndef GORELEASER_CURRENT_TAG
 export GORELEASER_CURRENT_TAG=$(GIT_TAG)
 endif
 
+.PHONY: docker-buildx
+docker-buildx:
+docker-buildx:
+	 docker buildx create --use --name=caren --platform=linux/arm64,linux/amd64 || true
+
 .PHONY: build-snapshot
 build-snapshot: ## Builds a snapshot with goreleaser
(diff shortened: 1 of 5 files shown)


make/goreleaser.mk
```diff
@@ -8,6 +8,11 @@ ifndef GORELEASER_CURRENT_TAG
 export GORELEASER_CURRENT_TAG=$(GIT_TAG)
 endif
 
+.PHONY: docker-buildx
+docker-buildx:
```
dlipovetsky, line 12:
```suggestion
docker-buildx: ## Creates buildx builder container that supports multiple platforms.
```

dlipovetsky (approved):

dlipovetsky:
Because it also builds multi-arch images, I think we want the `release` target to also depend on `docker-buildx`.

supershal (approved):
I tested on mac + colima(x86) and mac + colima(arm64) successfully.

jimmidyson:
I think an alternative way to fix this without having to create a new buildx builder is to change the Dockerfile to be multistage, the first stage to run the mindthegap create bundle with the native build host arch, the second stage just to copy the created bundle to the arch specific mindthegap image. This wouldn't require running the mindthegap binary with a different arch to the host.

make/goreleaser.mk
```diff
@@ -8,6 +8,11 @@ ifndef GORELEASER_CURRENT_TAG
 export GORELEASER_CURRENT_TAG=$(GIT_TAG)
 endif
 
+.PHONY: docker-buildx
+docker-buildx: ## Creates buildx builder container that supports multiple platforms.
+docker-buildx:
+	 docker buildx create --use --name=caren --platform=linux/arm64,linux/amd64 || true
```
jimmidyson, line 14: Won't the use flag here cause all my other (non CAREN) image builds to also use this builder? If we go this way, then perhaps it's better not to add the use here and instead specify the builder to use in goreleaser config (https://goreleaser.com/customization/docker/#use-a-specific-builder-with-docker-buildx)?

faiq: that's a good idea, i'll add that to the config.

dlipovetsky (approved):

PR #745, opened 2024-06-24
build: Tidy up goreleaser config

jimmidyson:
This commit tidies up the goreleaser config that contained some
configuration that is no longer required, e.g. deleting the `v`

prefix handling code; templating the arch specific Docker
image builds (now handled by loading the arch specific
image into KinD and e2e environment); and removing the
caren buildx build container (changing the helm registry
Dockerfile to be a multistage build to only run the
compatible architecture of mindthegap).

### DIFF
--- a/hack/addons/mindthegap-helm-registry/Dockerfile
+++ b/hack/addons/mindthegap-helm-registry/Dockerfile
@@ -1,5 +1,16 @@
-FROM ghcr.io/mesosphere/mindthegap:v1.14.3
+ARG MINDTHEGAP_VERSION=v1.14.3
+
+FROM --platform=${BUILDPLATFORM} ghcr.io/mesosphere/mindthegap:${MINDTHEGAP_VERSION} as bundle_builder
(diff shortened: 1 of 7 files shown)


hack/addons/mindthegap-helm-registry/Dockerfile
```diff
@@ -1,5 +1,16 @@
-FROM ghcr.io/mesosphere/mindthegap:v1.14.3
+ARG MINDTHEGAP_VERSION=v1.14.3
+
+FROM --platform=${BUILDPLATFORM} ghcr.io/mesosphere/mindthegap:${MINDTHEGAP_VERSION} as bundle_builder
```
jimmidyson, line 3: Making this into a multiplatform build saves a small amount of time by benefitting from layer caching (will only be run once for both amd64 and arm64 builds as the layers will be shared). Also doesn't require a new buildx builder instance.

faiq (approved):
ran the changes and looks good.

dkoshkin (approved):

PR #752, opened 2024-06-26
fix: skip kubeadm CA file when Secret doesn't have a CA 

dkoshkin:
Found a bug when testing with a mirror registry where it doesn't have the CA key set in the Secret.
The handler incorrectly generated a patch to add `/etc/certs/...` file with a Secret source and key `ca.crt`, however its possible that the Secret does not have that key set.
Example of an incorrect file:
```
  - contentFrom:
      secret:
        key: ca.crt
        name: e2e-aws-ag-cc-test-1719374309-image-registry-mirror-credentials
    path: /etc/certs/999867407951.dkr.ecr.us-west-2.amazonaws.com-e2e-aws-ag-cc-test-1719374309.pem
    permissions: "0600"
```
This led to CAPI failing because `ca.crt` did not exist in the Secret.

Fixes #

The first commit contains a failing unit tests to cover this scenario.

### DIFF
--- a/pkg/handlers/generic/mutation/mirrors/containerd_files.go
+++ b/pkg/handlers/generic/mutation/mirrors/containerd_files.go
@@ -19,7 +19,7 @@ import (
 
 const (
 	containerdHostsConfigurationOnRemote = "/etc/containerd/certs.d/_default/hosts.toml"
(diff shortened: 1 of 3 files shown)


jimmidyson (approved):
Thanks for fixing!

faiq (approved):

PR #753, opened 2024-06-27
ci: Add Nutanix e2e workflow using self-hosted github runner

thunderboltsid:
This is necessary as we want some e2e jobs to run on hosted runners and some to run on self-hosted runners. The actions runner controller does not support using multiple labels to target the runners See: https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/using-actions-runner-controller-runners-in-a-workflow#using-runner-scale-set-names


.github/workflows/checks.yml
```diff
@@ -72,14 +71,32 @@ jobs:
...
     strategy:
       matrix:
         provider:
           - Docker
           # Uncomment below once we have the ability to run e2e tests on other providers from GHA.
           # - AWS
-          # - Nutanix
       fail-fast: false
     uses: ./.github/workflows/e2e.yml
     with:
```
jimmidyson, line 102: I wonder if we can use a new input to the e2e test to change the `runs-on` and use a conditional to check the provider to set the runner label in the inputs here?

jimmidyson: Pushed #755 that does this.

thunderboltsid:
Closed in favor of https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/755

PR #762, opened 2024-06-28
feat: Add waiter for object

dlipovetsky:
Implements a wait for a check to pass against a typed object.

We'll use this in some lifecycle handers, e.g., in a future change for deploying ServiceLoadBalancer configuration to the remote cluster.

Fixes #

### DIFF
--- a/pkg/wait/wait.go
+++ b/pkg/wait/wait.go
@@ -1,0 +1,74 @@
+// Copyright 2024 Nutanix. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+package wait
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"k8s.io/apimachinery/pkg/util/wait"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// CheckFailedError is used to determine whether the wait failed because wraps an error returned by a failed check.
+type CheckFailedError struct {
+	cause error
+}
+
+func (e *CheckFailedError) Error() string {
+	return fmt.Sprintf("check failed: %s", e.cause)
+}
+
+func (e *CheckFailedError) Is(target error) bool {
+	_, ok := target.(*CheckFailedError)
+	return ok
+}
+
+func (e *CheckFailedError) Unwrap() error {
+	return e.cause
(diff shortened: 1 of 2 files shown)


pkg/wait/wait_test.go
```diff
@@ -0,0 +1,137 @@
...
+						ObjectMeta: v1.ObjectMeta{
+							Name: "example",
+						},
+					},
+				},
+			},
+			errCheck: func(err error) bool {
+				return errors.Is(err, &CheckFailedError{}) &&
+					!wait.Interrupted(err)
+			},
```
dkoshkin, line 123: Nice!

dkoshkin (approved):

pkg/wait/wait.go
```diff
@@ -0,0 +1,74 @@
...
+
+	var getErr error
+	waitErr := wait.PollUntilContextTimeout(
+		ctx,
+		input.Interval,
+		input.Timeout,
+		true,
+		func(checkCtx context.Context) (bool, error) {
+			if getErr = input.Reader.Get(checkCtx, key, input.Target); getErr != nil {
+				// Retry if get fails.
```
jimmidyson, line 54: Maybe only try on not found?

pkg/wait/wait.go
```diff
@@ -0,0 +1,74 @@
...
+				return false, &CheckFailedError{cause: err}
+			} else {
+				// Retry if check fails.
+				return ok, nil
+			}
+		})
+
+	if wait.Interrupted(waitErr) {
+		if getErr != nil {
+			return fmt.Errorf("%w; last get error: %w", waitErr, getErr)
```
jimmidyson, line 68: Can you use multiple wrapped errors like this? Do both get included in error chain?

jimmidyson: From https://pkg.go.dev/fmt#Errorf

> If there is more than one %w verb, the returned error will implement an Unwrap method returning a []error containing all the %w operands in the order they appear in the arguments.

So all good here! TIL :)

jimmidyson (approved):

faiq (approved):

PR #771, opened 2024-07-02
feat: Use HelmAddon as default addon strategy

jimmidyson:
This is properly defaulted by CAPI as can be seen in the Cluster resource when it gets applied (these do not appear in the example cluster manifests):

```
      variables:
      - name: clusterConfig
        value:
          addons:
            clusterAutoscaler:
              strategy: HelmAddon
            cni:
              provider: Cilium
              strategy: HelmAddon
            csi:
              defaultStorage:
                provider: local-path
                storageClassConfig: default
              providers:
                local-path:
                  storageClassConfigs:
                    default:
                      allowExpansion: false
                      reclaimPolicy: Delete
                      volumeBindingMode: WaitForFirstConsumer
                  strategy: HelmAddon
              snapshotController:
                strategy: HelmAddon
            nfd:
              strategy: HelmAddon
```


dkoshkin (approved):
Nice!

thunderboltsid (approved):

PR #774, opened 2024-07-02
fix: use minimal image when deploying nfd chart

faiq:
When deploying with the helm addon strategy i noticed that we did not deploy the minimal image. 

I'm going to follow this up with a PR that automates syncing the values in the templates and `hack/addons/kustomize/**/helm-values.yaml` 

Fixes #

### DIFF
--- a/charts/cluster-api-runtime-extensions-nutanix/templates/nfd/manifests/helm-addon-installation.yaml
+++ b/charts/cluster-api-runtime-extensions-nutanix/templates/nfd/manifests/helm-addon-installation.yaml
@@ -8,6 +8,8 @@ metadata:
   name: '{{ .Values.hooks.nfd.helmAddonStrategy.defaultValueTemplateConfigMap.name }}'
 data:
   values.yaml: |-
+    image:
+      tag: v0.15.2-minimal
     master:
       extraLabelNs:
         - nvidia.com


charts/cluster-api-runtime-extensions-nutanix/templates/nfd/manifests/helm-addon-installation.yaml
```diff
@@ -8,6 +8,8 @@ metadata:
   name: '{{ .Values.hooks.nfd.helmAddonStrategy.defaultValueTemplateConfigMap.name }}'
 data:
   values.yaml: |-
+    image:
+      tag: v0.15.2-minimal
```
dkoshkin, line 12: Do we get a variable with the chart version? Or can automatically set this when generating these values?
We will forget to change this.

faiq: i'll be following up with a tool to sync these files with. this was just one i noticed that was broken right now 

hack/addons/kustomize/**/helm-values.yaml

dkoshkin (approved):

jimmidyson:
Fixing the `lint-test-helm` check in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/779.

charts/cluster-api-runtime-extensions-nutanix/templates/nfd/manifests/helm-addon-installation.yaml
jimmidyson: Merge this and follow up in a different PR, i guess that's what you're saying @faiq ?

jimmidyson (approved):

PR #775, opened 2024-07-02
fix: Namespacesync controller should reconcile an updated namespace

dlipovetsky:
The namespacesync controller did not reconcile a namespace when it was updated. We anticipated namespaces to meet the criteria for "target namespace" when they were created. We also wanted to avoid reconciling namespaces unnecessarily.

This PR makes the controller reconcile an updated namespace, if the answer to question "is this a target namespace," has changed. It also adds an integration test (which fails without the subsequent fix commit).

Fixes #

### DIFF
--- a/pkg/controllers/namespacesync/controller.go
+++ b/pkg/controllers/namespacesync/controller.go
@@ -66,9 +66,8 @@ func (r *Reconciler) SetupWithManager(
 							return false
 						}
 						// Only reconcile the namespace if the answer to the question "Is this a
-						// target namespace?" has changed. Other changes are not relevant to
-						// this controller.
-						return r.IsTargetNamespace(nsOld) != r.IsTargetNamespace(nsNew)
+						// target namespace?" changed from no to yes.
+						return !r.IsTargetNamespace(nsOld) && r.IsTargetNamespace(nsNew)
 					},
 					DeleteFunc: func(e event.DeleteEvent) bool {
 						// Ignore deletes.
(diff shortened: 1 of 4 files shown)


pkg/controllers/namespacesync/controller.go
```diff
@@ -52,12 +52,23 @@ func (r *Reconciler) SetupWithManager(
...
+							return false
+						}
+						nsNew, ok := e.ObjectNew.(*corev1.Namespace)
+						if !ok {
+							return false
+						}
+						// Only reconcile the namespace if the answer to the question "Is this a
+						// target namespace?" has changed. Other changes are not relevant to
+						// this controller.
+						return r.IsTargetNamespace(nsOld) != r.IsTargetNamespace(nsNew)
```
dkoshkin, line 71: Do we also want to skip the case where the user removed the label from the namespace?
```suggestion
						return r.IsTargetNamespace(nsNew) && r.IsTargetNamespace(nsOld) != r.IsTargetNamespace(nsNew)
```

dkoshkin (approved):

pkg/controllers/namespacesync/controller.go
dlipovetsky: Good observation! We can skip every case except the one where the label is added. Done in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/775/commits/6553aff0210d8db0da37ca50437dda09de5f4399

pkg/controllers/namespacesync/controller.go
```diff
@@ -52,12 +52,22 @@ func (r *Reconciler) SetupWithManager(
...
-						return false
+						nsOld, ok := e.ObjectOld.(*corev1.Namespace)
+						if !ok {
+							return false
+						}
+						nsNew, ok := e.ObjectNew.(*corev1.Namespace)
+						if !ok {
+							return false
+						}
+						// Only reconcile the namespace if the answer to the question "Is this a
```
supershal, line 68: very well simple explanation. thanks.

supershal (approved):

PR #763, opened 2024-06-28
feat: Define ServiceLoadBalancer Configuration API

dlipovetsky:
**What problem does this PR solve?**:

**Which issue(s) this PR fixes**:
Fixes #

**How Has This Been Tested?**:

**Special notes for your reviewer**:

PR #755, opened 2024-06-27
ci: Run Nutanix provider e2e tests on self-hosted runner

jimmidyson:
Modify e2e workflow to Run Nutanix jobs on self-hosted-github runner

This is necessary as we want some e2e jobs to run on hosted runners
and some to run on self-hosted runners. The actions runner controller
does not support using multiple labels to target the runners
See: https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/using-actions-runner-controller-runners-in-a-workflow#using-runner-scale-set-names

Depends on #754.

PR #778, opened 2024-07-03
feat: Define ServiceLoadBalancer Configuration API

dlipovetsky:
This adds an optional configuration for the ServiceLoadBalancer Addon, consisting of one or more IPv4 address ranges, e.g.,

```yaml
apiVersion: cluster.x-k8s.io/v1beta1
kind: Cluster
metadata:
  name: <NAME>
spec:
  topology:
    variables:
      - name: clusterConfig
        value:
          addons:
            serviceLoadBalancer:
              provider: MetalLB
              configuration:
                addressRanges:
                - start: 10.100.1.1
                  end: 10.100.1.20
                - start: 10.100.1.51
                  end: 10.100.1.70
```

(This is a copy of https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/763. I had to close that after https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/755 added required checks that can't be run from PRs from public forks. )

Fixes #


dkoshkin (approved):
Please also update the [doc](https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/main/docs/content/addons/serviceloadbalancer.md)

api/v1alpha1/addon_types.go
```diff
@@ -257,4 +257,27 @@ type ServiceLoadBalancer struct {
...
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinItems=1
+	AddressRanges []AddressRange `json:"addressRanges"`
+}
+
+// AddressRange defines an IPv4 range.
+type AddressRange struct {
+	// Source: https://github.com/metal3-io/ip-address-manager/blob/b4f7c2a/api/v1alpha1/common_types.go#L19
+
+	// +kubebuilder:validation:Pattern="^((([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5]))$"
```
jimmidyson, line 278:
```suggestion
	// +kubebuilder:validation:Format=ipv4
```

api/v1alpha1/addon_types.go
```diff
@@ -257,4 +257,27 @@ type ServiceLoadBalancer struct {
...
+	Start string `json:"start"`
+
+	// +kubebuilder:validation:Pattern="^((([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5]))$"
```
jimmidyson, line 281:
```suggestion
	// +kubebuilder:validation:Format=ipv4
```

jimmidyson (changes requested):
Let's use openapi format instead of regex

api/v1alpha1/addon_types.go
dlipovetsky: Ok. Can you please link to where this is documented?

dlipovetsky: I appreciate the effort, but I'm looking for something authoritative.

The swagger docs says:

> However, format is an open value, so you can use any formats, even not those defined by the OpenAPI Specification

And the kubebuilder docs don't mention ipv4.

And I don't think the apiextensions code is relevant here.

It looks like the API server imports https://github.com/kubernetes/kube-openapi, but that doesn't have any documentation...

dlipovetsky: I want to an authoritative source, so that future me (and others) can discover what formats are available.

jimmidyson: And please add a test for this to prove the validation works. We have examples of this in various places.

jimmidyson: I documented about using formats in our [API conventions](https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/main/docs/content/contributing/api_conventions.md?plain=1#L55-L61) btw, including highlighting that those linked formats work.

jimmidyson: We already use `uri` in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/main/api/v1alpha1/clusterconfig_types.go#L244 and have tests for this in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/main/pkg/handlers/generic/mutation/mirrors/variables_test.go

dlipovetsky: Upstream issue: https://github.com/kubernetes-sigs/kubebuilder/issues/3479

jimmidyson (approved):

PR #782, opened 2024-07-03
ci: Move govulncheck to nightly and push to main triggers

jimmidyson:
This removes potentially confusing checks on PRs while still giving
early feedback on CVEs that may have been introduced.


dlipovetsky (approved):
The nightly trigger will alert us to issues introduced by code or dependencies already in main, while the push trigger will help us identify whether some PR introduced the issue.

PR #787, opened 2024-07-04
test(e2e): Use parallel tests for providers other than Docker

jimmidyson:
Keep Docker using serial tests to not exhaust resources of the machine
that the Docker tests are running on. Other providers do not require so
many resources as they use cloud or external infrastructure for clusters.


dlipovetsky (approved):

.github/workflows/e2e.yml
```diff
@@ -77,25 +69,8 @@ jobs:
           echo "After removing files:"
           df -h
 
-      - name: Get Control Plane endpoint IP
-        id: get-control-plane-endpoint-ip
-        if: inputs.provider == 'Nutanix'
-        run: |
-          export CONTROL_PLANE_ENDPOINT_RANGE_START="${{ vars.NUTANIX_CONTROL_PLANE_ENDPOINT_RANGE_START }}"
-          export CONTROL_PLANE_ENDPOINT_RANGE_END="${{ vars.NUTANIX_CONTROL_PLANE_ENDPOINT_RANGE_END }}"
-          control_plane_endpoint_ip="$(devbox run -- make nutanix-cp-endpoint-ip)"
```
supershal, line 86: we should remove `nutanix-cp-endpoint-ip:` from makefile

test/e2e/framework/nutanix/networking.go
```diff
@@ -0,0 +1,230 @@
...
+	default:
+		return uuid.UUID{}, fmt.Errorf("unknown response: %+v", subnets)
+	}
+}
+
+type reservedIPs struct {
+	ReservedIPs []string `json:"reserved_ips"`
+}
+
+func ReserveIP(
```
supershal, line 87: I could not find how this function used?

.github/workflows/e2e.yml
jimmidyson: Good catch!

supershal (approved):
LGTM.

PR #790, opened 2024-07-05
build: Prepend header to generated clusterclass and cluster templates

dlipovetsky:
It's not clear that the files in `examples/capi-quick-start` and `defaultclusterclasses` are generated, and should not be edited. This adds headers that make that clear.

Fixes #


hack/common.sh
```diff
@@ -9,3 +9,22 @@ trap_add() {
...
+
+  generated_by_header_file="${tmp_dir}/header.txt"
+  printf "#=================================================================\n\
+#                 DO NOT EDIT THIS FILE\n\
+#  IT HAS BEEN GENERATED BY %s\n\
+#=================================================================\n\
+" "$generated_by" >"$generated_by_header_file"
+
+  tmp_dst_file="${tmp_dir}/dst.txt"
+  cat "$generated_by_header_file" "$dst_file" >"$tmp_dst_file"
```
deepakm-ntnx, line 28: there are some simple ways to do it using sed -i. pls check if this reduces the code https://superuser.com/questions/246837/how-do-i-add-text-to-the-beginning-of-a-file-in-bash

dlipovetsky: I actually did this at first (and ended up reading that same link!) :smile: 

IMO, that method more difficult to understand, and easier to break.

I think our goal is code that is easier to understand, and to maintain, even at the expense of length, or performance (e.g. this PR is relatively "slow," because it writes temp files, instead of doing in-memory replacement).

examples/capi-quick-start/aws-cluster-calico-crs.yaml
```diff
@@ -1,3 +1,7 @@
+#=================================================================
```
dkoshkin, line 1: These files get published as part of the release where users of it can modify these if needed.

dlipovetsky: Thanks I did not realize that!

How about I remove the warning as part of the release process?

dlipovetsky: I wonder if goreleaser will complain if the files are modified in the repository... Do I need to copy them to an external directory?

dlipovetsky: I think it might... https://goreleaser.com/errors/dirty/

PR #795, opened 2024-07-08
fix: add omitempty to addon strategy

dkoshkin:
Saw a failure when using the API to generate a spec and not setting the `strategy` in Go code:
```
is invalid: [spec.topology.variables.addons.cni.strategy: Unsupported value: "": supported values: "ClusterResourceSet", "HelmAddon"...
```

Adding `+kubebuilder:validation:Optional` didn't change the generated CRDs, but I think we still want them?

Fixes #


dlipovetsky:
> Optional fields have the following properties:
> They are a pointer type in the Go definition (e.g. AwesomeFlag *SomeFlag) or have a built-in nil value (e.g. maps and slices).
> -- https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

Given https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/9039258cf489e4d9e6c6d6c46fedc0b56c9b9f4a/api/v1alpha1/addon_types.go#L101

Do we want to make this a pointer to a string?

faiq (approved):

jimmidyson:
> Do we want to make uses of AddonStrategy pointers?

Yes please!

jimmidyson:
@dkoshkin 

> Adding +kubebuilder:validation:Optional didn't change the generated CRDs

That's because the `+kubebuilder:validation:Required` annotation is on the [`AddonType` type alias](https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/main/api/v1alpha1/addon_types.go#L99). I've updated that now and will push in a minute.

jimmidyson:
Switched to pointer field, fixed up handlers and tests, and pushed.

jimmidyson (approved):

mhrabovcin (approved):

PR #796, opened 2024-07-08
ci: Run e2e jobs only if unit-test, lint-*, and pre-commit jobs pass

dlipovetsky:
It's wasteful to run the e2e jobs before the other jobs pass, since making the other jobs pass almost certainly requires a change to the PR, which then requires the e2e jobs to run again, discarding the previous result.

Before:
image

After:
image

Fixes #

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -55,6 +55,12 @@ jobs:
           test-results: test.json
 
   e2e-quick-start:
+    needs:
+      - "lint-gha"
+      - "lint-go"
+      - "lint-test-helm"
+      - "pre-commit"
+      - "unit-test"
     strategy:
       matrix:
         provider:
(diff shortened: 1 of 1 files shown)


dkoshkin (approved):
The downside is that e2e tests will now take about an extra 4-5 for the e2e tests to start, but now that we have Nutanix and soon will have AWS tests, the tradeoff is worth it.

supershal (approved):
Thanks. This is useful when tests are triggered for every push to the PR.

dlipovetsky:
> The downside is that e2e tests will now take about an extra 4-5 for the e2e tests to start, but now that we have Nutanix and soon will have AWS tests, the tradeoff is worth it.

Good observation. Since we run the "pre-requisite" jobs in parallel, we only need to wait for the longest job to complete.

Right now ,that's `unit-test`, and it takes about 6m; 4.5m running the tests, and 1.5m setting up devbox and go cache.

image

I think the time also depends on the CPU (and possibly memory) allocation of the runner. In CI, the unit tests for the 3 modules run in 212, 32, and 9 seconds. For me, locally, the same tests run in 58, 4, and 3 seconds.

PR #807, opened 2024-07-16
fix: Clean up MetalLB pod security standards labels

jimmidyson:
This commit removes the `pod-security.kubernetes.io/audit` and
`pod-security.kubernetes.io/warn` labels as they are redundant when
specifying the `pod-security.kubernetes.io/enforce` label to the same
level.

Also add the `pod-security.kubernetes.io/enforce-version` label to
always enforce the latest pod security policy version, even on upgrade.
This is fine because we are specifying the most privileged pod security
standard, `privileged`, and as such should be safe to always enforce the
latest policy version.


faiq (approved):

dlipovetsky:
Thanks. Really wish these could be managed together with the Chart.

dlipovetsky (approved):

PR #808, opened 2024-07-16
feat: Secure ciphers, min TLS v1.2, and disable auto TLS for etcd

### LINKED ISSUE #806: Improve default ETCD configuration
there is few change we can apply to the default etcd config to:
- fix CVE issue rise by Nesus scanner
- improve Kubernetes STIG posture

goal is to inject inside kubeadmcontrolplane object in spec.kubeadmConfigSpec. clusterConfiguration.etcd

the following entry

```
  etcd:
    local:
      extraArgs:
        auto-tls: "false"
        peer-auto-tls: "false"
        cipher-suites: "TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384"
```

jimmidyson:
This increases ootb security and provides STIG compliance for
this area at least.

Fixes #806.

### DIFF
--- a/pkg/handlers/generic/mutation/etcd/inject.go
+++ b/pkg/handlers/generic/mutation/etcd/inject.go
@@ -5,6 +5,8 @@ package etcd
 
 import (
 	"context"
+	"crypto/tls"
+	"strings"
 
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -45,6 +47,21 @@ func newEtcdPatchHandler(
 	}
 }
 
+var defaultEtcdExtraArgs = map[string]string{
+	"auto-tls":      "false",
+	"peer-auto-tls": "false",
+	"cipher-suites": strings.Join(
+		[]string{
+			tls.CipherSuiteName(tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256),
+			tls.CipherSuiteName(tls.TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256),
+			tls.CipherSuiteName(tls.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384),
+			tls.CipherSuiteName(tls.TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384),
+		},
+		",",
+	),
+	"tls-min-version": "TLS1.2",
+}
+
 func (h *etcdPatchHandler) Mutate(
 	ctx context.Context,
 	obj *unstructured.Unstructured,
@@ -62,11 +79,7 @@ func (h *etcdPatchHandler) Mutate(
 		h.variableName,
 		h.variableFieldPath...,
 	)
-	if err != nil {
-		if variables.IsNotFoundError(err) {
-			log.V(5).Info("etcd variable not defined")
-			return nil
-		}
+	if err != nil && !variables.IsNotFoundError(err) {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/handlers/generic/mutation/etcd/inject.go
```diff
@@ -95,10 +97,25 @@ func (h *etcdPatchHandler) Mutate(
 			}
 
 			localEtcd := obj.Spec.Template.Spec.KubeadmConfigSpec.ClusterConfiguration.Etcd.Local
-			if etcd.Image != nil && etcd.Image.Tag != "" {
+
+			if localEtcd.ExtraArgs == nil {
+				localEtcd.ExtraArgs = make(map[string]string, 3)
```
dkoshkin, line 102: nit: 
```suggestion
				localEtcd.ExtraArgs = make(map[string]string, len(defaultEtcdExtraArgs))
```

dkoshkin (approved):

dlipovetsky:
Does this affect FIPS compliance in any way?

jimmidyson:
@dlipovetsky 

> Does this affect FIPS compliance in any way?

That is a good call out! And yes it does 😞 While go did upgrade BoringCrypto that supports TLS 1.3 in [go 1.21.6](https://github.com/golang/go/issues/64717), it was then [reverted in subsequent versions](https://github.com/golang/go/issues/65321) including >go1.21.6 and go1.22 until the new version of BoringCrypto is approved by NIST.

jimmidyson:
The motivation for this change was to mitigate against sweet32 due to vulnerable 3DES ciphers. In go1.23 3DES ciphers are [disabled by default](https://go-review.googlesource.com/c/go/+/587295) so this change will be unnecessary when etcd is built with go1.23.

In the meantime, I will adopt the recommended ciphers in #806 by @tuxtof instead. I have checked they are all FIPS-compliant so this should work for both FIPS and non-FIPS deployments until TLS1.3 is supported in go FIPS.

pkg/handlers/generic/mutation/etcd/inject.go
```diff
@@ -45,6 +47,21 @@ func newEtcdPatchHandler(
...
+			tls.CipherSuiteName(tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256),
+			tls.CipherSuiteName(tls.TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256),
+			tls.CipherSuiteName(tls.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384),
+			tls.CipherSuiteName(tls.TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384),
+		},
+		",",
+	),
+	"tls-min-version": "TLS1.2",
+}
+
```
dlipovetsky, line 64: Please add a brief comment explaining the goal of this configuration (similar to my comment about the tests). :pray:

pkg/handlers/generic/mutation/etcd/inject_test.go
```diff
@@ -56,6 +56,12 @@ var _ = Describe("Generate etcd patches", func() {
 							"local": map[string]interface{}{
 								"imageRepository": "my-registry.io/my-org/my-repo",
 								"imageTag":        "v3.5.99_custom.0",
+								"extraArgs": map[string]interface{}{
+									"auto-tls":        "false",
+									"peer-auto-tls":   "false",
+									"cipher-suites":   "TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384", //nolint:lll // Long list of ciphers ok in test.
+									"tls-min-version": "TLS1.2",
+								},
```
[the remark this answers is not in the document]
jimmidyson: Added.

pkg/handlers/generic/mutation/etcd/inject.go
jimmidyson: Added.

faiq (approved):

PR #798, opened 2024-07-09
ci: adds tooling to sync values file 

faiq:
**What problem does this PR solve?**:

This tooling syncs the helm values configmaps in charts/templates from hack/addons/kustomize

part of the changes here also include some file organization fixes so that we have similar structures in hack and in addons.

**Which issue(s) this PR fixes**:
Fixes #

**How Has This Been Tested?**:

**Special notes for your reviewer**:

PR #769, opened 2024-07-02
build: Add images list make target

jimmidyson:
Integrate into release build so the list of images is attaches
to each release.

PR #809, opened 2024-07-17
feat: Extract CAAPH values templates to files

jimmidyson:
This simplifies the Helm templating directives by not requiring
inception-style escaping of templating braces, e.g. `{{ "{{" }}``
which are very hard to read and can introduce bugs.


charts/cluster-api-runtime-extensions-nutanix/templates/ccm/aws/manifests/helm-addon-installation.yaml
```diff
@@ -7,28 +7,15 @@ kind: ConfigMap
...
-    args:
-      - --v=2
-      - --cloud-provider=aws
-      - --configure-cloud-routes=false
-
     {{ "{{" }} $k8sMinorVersionToCCMVersion := dict
     {{ range $k8sVersion, $ccmVersion := .Values.hooks.ccm.aws.k8sMinorVersionToCCMVersion -}}
       "{{ $k8sVersion }}" "{{ $ccmVersion }}"
     {{ end -}}
     {{ "}}" }}
```
faiq, line 18: should this map be gone now?

jimmidyson: No that is still required as it is configured by Helm template from Helm values. If you render the whole template you'll see why it's needed.

faiq: Can it be placed in the file as well?

jimmidyson: No because it is Helm templated, not an file to be included as-is.

faiq:
I think we should consolidate the work in here and #798. This change would make it so we can have a simpler syncing process between the values in `hack/addons/kustomize` and `charts`.


Once it's all consolidated, we can finally address the image listing problem in #769

faiq:
closing this because the work is included in #819

PR #810, opened 2024-07-18
fix: Fix ownership of ClusterAutoscaler resources

jimmidyson:
The Cluster Autoscaler addon (either HelmChartProxy or ClusterResourceSet
) has to be deployed in the management cluster namespace and as such cannot
be owned by the workload cluster, which commonly resides in a different
namespace.

This commit fixes that and introduces a BeforeClusterDelete hook to delete
the HelmChartProxy or ClusterResourceSet rather than relying no Kubernetes
GC as was previously expected.

Tested locally. Created workload cluster in different namespace to management cluster. HCP does not have ownership set:

```
apiVersion: addons.cluster.x-k8s.io/v1alpha1
kind: HelmChartProxy
metadata:
  creationTimestamp: "2024-07-18T13:02:30Z"
  finalizers:
  - helmchartproxy.addons.cluster.x-k8s.io
  generation: 1
  name: cluster-autoscaler-my-docker-cluster
  namespace: test
  resourceVersion: "6238"
  uid: 0a5ebb11-5318-4114-bedd-9bcc9459f85e
spec:
``` 

And is properly cleaned up on deletion:

```
$ k get hcp -A
No resources found
```

### DIFF
--- a/pkg/handlers/generic/lifecycle/clusterautoscaler/strategy_crs.go
+++ b/pkg/handlers/generic/lifecycle/clusterautoscaler/strategy_crs.go
@@ -13,6 +13,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	"sigs.k8s.io/cluster-api/controllers/remote"
+	crsv1 "sigs.k8s.io/cluster-api/exp/addons/api/v1beta1"
 	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/common/pkg/k8s/client"
@@ -74,7 +75,7 @@ func (s crsStrategy) apply(
 		},
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: cluster.Namespace,
-			Name:      defaultCM.Name + "-" + cluster.Name,
+			Name:      s.crsNameForCluster(cluster),
 		},
 		Data: data,
 	}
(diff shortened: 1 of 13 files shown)


pkg/handlers/generic/lifecycle/clusterautoscaler/strategy_crs.go
```diff
@@ -119,3 +130,37 @@ func (s crsStrategy) apply(
...
+	}
+
+	crs := &crsv1.ClusterResourceSet{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: crsv1.GroupVersion.String(),
+			Kind:       "ClusterResourceSet",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: targetCluster.Namespace,
+			Name:      s.config.defaultClusterAutoscalerConfigMap + "-" + cluster.Name,
```
dkoshkin, line 154: nit: it was kind of difficult to track down why the name is based on `defaultClusterAutoscalerConfigMap`, maybe some helper function to get the name would be useful?

https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/eabc6896e1126086474a11f3153cb682cf427eb4/pkg/handlers/generic/lifecycle/clusterautoscaler/strategy_crs.go#L55

pkg/handlers/utils/utils.go
```diff
@@ -21,11 +21,23 @@ import (
 	"github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/common/pkg/k8s/client"
 )
 
+type EnsureCRSForClusterFromObjectsOptions struct {
+	// SetClusterOwnership will set the ownership of the CRS to the cluster when set to true
+	SetClusterOwnership bool
+}
+
+func DefaultEnsureCRSForClusterFromObjectsOptions() EnsureCRSForClusterFromObjectsOptions {
```
dkoshkin, line 29: Nice pattern

test/e2e/ownerreference_helpers.go
```diff
@@ -224,3 +226,30 @@ func clusterResourceSetBindingIsOnlyOwnedByClusterResourceSets(
...
+	clusterName string,
+) clusterctlcluster.GetOwnerGraphFilterFunction {
+	return func(u unstructured.Unstructured) bool {
+		if !clusterctlcluster.FilterClusterObjectsWithNameFilter(
+			clusterName,
+		)(
+			u,
+		) {
+			return false
+		}
```
dkoshkin, line 242: Am I reading this incorrectly? I don't understand the syntax. 
```
		if !clusterctlcluster.FilterClusterObjectsWithNameFilter(clusterName)(u) {
			return false
		}
```

dkoshkin (approved):
Thanks for fixing both CAAPH and CRS!

test/e2e/ownerreference_helpers.go
jimmidyson: Yes, sorry that's just bad formatting by golines 🤦 Fixed that now. Basically filter out anything without the cluster name in the resource name - this is the standard filtering that we adopted before this change.

deepakm-ntnx (approved):

PR #819, opened 2024-07-23
ci: Adds sync tool for hack and charts

faiq:

- Splits the templates helm templates to read from files instead of embedding in the configmap. This makes it easier to read and modify the templates #809
- Syncs `hack/` and `charts/` helm values to avoid drift. An explanation of the tool can be found in the README file in `hack/tools/sync-helm-values`

As a part of the file changes to support the sync `helm-values.yaml` had to be changed to use chart names as the config map field. 

For example `local-path-provisioner-csi` local-path-provisioner` 

Some changes in `pkg/handlers/generic/lifecycle/` were changed to use the new names. 

Fixes #


charts/cluster-api-runtime-extensions-nutanix/addons/ccm/aws/map-template.yaml
```diff
@@ -0,0 +1,7 @@
+# These values are a combination of a Helm template to create the dict for the k8s version to
+# CCM version map, and then including the non-templated values from the values-template.yaml.
+# This simplifies the templating required in the values-template.yaml.
+$k8sMinorVersionToCCMVersion := dict
+{{ range $k8sVersion, $ccmVersion := .Values.hooks.ccm.aws.k8sMinorVersionToCCMVersion -}}
+  "{{ $k8sVersion }}" "{{ $ccmVersion }}"
```
supershal, line 6: not too familiar with the dict format. 
should the KV pair assigned to the `k8sMinorVersionToCCMVersion` ?
ie.
```
set $k8sMinorVersionToCCMVersion "{{ $k8sVersion }}" "{{ $ccmVersion }}"
```

faiq: no it should be a dict of all the values of k8s minor to image version

```yaml 
    {{ $k8sMinorVersionToCCMVersion := dict
    "1.27" "v1.27.7"
    "1.28" "v1.28.6"
    "1.29" "v1.29.3"
    "1.30" "v1.30.1"
    }}
```

https://helm.sh/docs/chart_template_guide/function_list/#dict it should be a list of pairs

hack/tools/sync-helm-values/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+For example the file in `hack/addons/kustomize/nfd/manifests/helm-addon-installation.yaml` gets synced to
+`charts/cluster-api-runtime-extensions-nutanix/templates/nfd/manifests/helm-addon-installation.yaml` and the
+`hack/addons/kustomize/nfd/manifests/values-template.yaml` get synced to
+`charts/cluster-api-runtime-extensions-nutanix/addons/nfd/values-template.yaml`.
+
+The tool sets the files located in `hack/addons/kustomize` to serve as a single source of truth for
+`ClusterResourceSets` that are generated as well as the helm charts deployed.
+
+## How it works
```
supershal, line 23: Thanks for the doc.

hack/tools/sync-helm-values/sync-values.go
```diff
@@ -0,0 +1,175 @@
...
+	return fullPath, nil
+}
+
+func SyncHelmValues(sourceDirectory, destDirectory, licenseFile string) error {
+	sourceFS := os.DirFS(sourceDirectory)
+	err := fs.WalkDir(sourceFS, ".", func(filepath string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if strings.Contains(filepath, "tigera") {
```
supershal, line 107: can you please add comment why the `tigera` related files are skipped.

charts/cluster-api-runtime-extensions-nutanix/addons/ccm/aws/map-template.yaml
supershal: I see. thanks for the reference. I was looking at go template syntax for dict.

supershal (approved):
Tested locally by modifying files in `/hack/addons/kustomize` and syncing the values.

hack/tools/sync-helm-values/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+This tool works by traversing the file tree located at `hack/addons/kustomize` and looking for files named
+`helm-addon-installation.yaml` and `*-template.yaml` and moves them to
```
dlipovetsky, line 26:
```suggestion
`helm-addon-installation.yaml` and `*-template.yaml` and copies them to
```

hack/tools/sync-helm-values/README.md
```diff
@@ -0,0 +1,41 @@
+<!--
+ Copyright 2023 Nutanix. All rights reserved.
+ SPDX-License-Identifier: Apache-2.0
+ -->
+
+# Sync Helm Values
+
+## Motivation
+
+The purpose of this tool is to synchronize the helm values files located `hack/addons/kustomize` with the files
```
dlipovetsky, line 10:
```suggestion
The purpose of this tool is to synchronize the helm values files located in `hack/addons/kustomize` with the files
```

dlipovetsky:
Just checked out the branch and ran the command, but got an error.

```
> go run sync-values.go \
                                                                                           -kustomize-directory=../../addons/kustomize/ \
                                                                                           -helm-chart-directory=../../../charts/cluster-api-runtime-extensions-nutanix/
failed to sync err: failed to write license to file /home/dlipovetsky/nutanix/cluster-api-runtime-extensions-nutanix/charts/cluster-api-runtime-extensions-nutanix/templates/ccm/aws/manifests/helm-addon-installation.yaml write /home/dlipovetsky/nutanix/cluster-api-runtime-extensions-nutanix/charts/cluster-api-runtime-extensions-nutanix/templates/ccm/aws/manifests/helm-addon-installation.yaml: copy_file_range: is a directory
```

faiq:
@dlipovetsky can you run the make target?

`make sync-helm-values`

.pre-commit-config.yaml
```diff
@@ -70,7 +76,7 @@ repos:
   - id: check-yaml
     args: ["-m", "--unsafe"]
     stages: [commit]
-    exclude: ^charts/.+/templates/
+    exclude: ^charts/[^/]+/addons/.|charts/[^/]+/templates/.|hack/addons/kustomize/.*$
```
jimmidyson, line 79:
```suggestion
    exclude: ^(charts/.+/(templates|addons)/|hack/addons/kustomize/)
```

.pre-commit-config.yaml
```diff
@@ -140,7 +146,7 @@ repos:
     name: License headers - YAML and Makefiles
     stages: [commit]
     files: (^Makefile|\.(ya?ml|mk))$
-    exclude: ^(internal/test|pkg/handlers/.+/embedded|examples|charts/cluster-api-runtime-extensions-nutanix/defaultclusterclasses)/.+\.ya?ml|docs/static/helm/index\.yaml|charts/cluster-api-runtime-extensions-nutanix/templates/helm-config.yaml$
+    exclude: ^(internal/test|pkg/handlers/.+/embedded|examples|charts/cluster-api-runtime-extensions-nutanix/(defaultclusterclasses|addons))/.+\.ya?ml|docs/static/helm/index\.yaml|charts/cluster-api-runtime-extensions-nutanix/templates/helm-config.yaml|hack/addons/kustomize/.+/(map|values)-template.yaml$
```
jimmidyson, line 149:
```suggestion
    exclude: ^(internal/test|pkg/handlers/.+/embedded|examples|charts/cluster-api-runtime-extensions-nutanix/((defaultclusterclasses|addons)/.+\.ya?ml|templates/helm-config.yaml)|docs/static/helm/index\.yaml|hack/addons/kustomize/.+/(map|values)-template.yaml)$
```

charts/cluster-api-runtime-extensions-nutanix/addons/cluster-autoscaler/values-template.yaml
```diff
@@ -0,0 +1,36 @@
+fullnameOverride: "cluster-autoscaler-{{ .Cluster.Name }}"
```
jimmidyson, line 1: This isn't correct any more, see https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/v0.13.7/charts/cluster-api-runtime-extensions-nutanix/templates/cluster-autoscaler/manifests/helm-addon-installation.yaml#L12

```suggestion
fullnameOverride: "cluster-autoscaler-{{ `{{ index .Cluster.Annotations "caren.nutanix.com/cluster-uuid" }}` }}"
```

hack/addons/kustomize/cluster-autoscaler/manifests/helm-values.yaml
```diff
@@ -1,11 +1,12 @@
 # Copyright 2023 Nutanix. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+# This is used to generate CRS
 ---
 # This is a hack, but because a single cluster-autoscaler deployment can only monitor a single Kubernetes cluster
 # we expect 'tmpl-clustername-tmpl' and 'tmpl-clusternamespace-tmpl'
 # to be replaced with the Cluster's name and namespace.
-fullnameOverride: cluster-autoscaler-tmpl-clusteruuid-tmpl
+fullnameOverride: cluster-autoscaler-tmpl-clustername-tmpl
```
jimmidyson, line 9: Please revert to clusteruuid

jimmidyson (changes requested):

jimmidyson:
I think I'm missing something, but why do we need to sync the values from hack to chart dir rather than just work with them directly in chart dir? This seems like it's duplicating files.

faiq:
> I think I'm missing something, but why do we need to sync the values from hack to chart dir rather than just work with them directly in chart dir? This seems like it's duplicating files.

Right now I think this would be helpful when updating values for addons with this new directory structure. Rather than going out and modifying them in `charts/addons` and `charts/templates` we can make all the changes for values when handling an addon in one place.  

I was also able to consolidate some of the files by sharing value files for some of the addons as well.

charts/cluster-api-runtime-extensions-nutanix/templates/cluster-autoscaler/manifests/cluster-autoscaler-configmap.yaml
```diff
@@ -169,7 +169,7 @@ data:
             - ./cluster-autoscaler
             - --cloud-provider=clusterapi
             - --namespace='{{ `{{ .Cluster.Namespace }}` }}'
-            - --node-group-auto-discovery=clusterapi:clusterName='{{ `{{ .Cluster.Name }}` }}',namespace='{{ `{{ .Cluster.Namespace }}` }}'
+            - --node-group-auto-discovery=clusterapi:clusterName='{{ `{{ index .Cluster.Annotations "caren.nutanix.com/cluster-uuid" }}` }}',namespace='{{ `{{ .Cluster.Namespace }}` }}'
```
jimmidyson, line 172: This isn't right.

charts/cluster-api-runtime-extensions-nutanix/templates/cluster-autoscaler/manifests/cluster-autoscaler-configmap.yaml
```diff
@@ -211,7 +211,7 @@ data:
               items:
               - key: value
                 path: kubeconfig
-              secretName: '{{ `{{ .Cluster.Name }}` }}-kubeconfig'
+              secretName: '{{ `{{ index .Cluster.Annotations "caren.nutanix.com/cluster-uuid" }}` }}-kubeconfig'
```
jimmidyson, line 214: Still should be `.Cluster.Name` here

jimmidyson (changes requested):
Found a few issues with the CA UUID annotation usage - please can you double check the changes related there.

hack/addons/kustomize/cluster-autoscaler/manifests/helm-values.yaml
```diff
@@ -1,9 +1,10 @@
 # Copyright 2023 Nutanix. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+# This is used to generate CRS
 ---
 # This is a hack, but because a single cluster-autoscaler deployment can only monitor a single Kubernetes cluster
-# we expect 'tmpl-clustername-tmpl' and 'tmpl-clusternamespace-tmpl'
+# we expect 'tmpl-clusteruuid-tmpl' and 'tmpl-clusteruuidspace-tmpl'
```
jimmidyson, line 7: Looks like a copy-paste issue here. 
```suggestion
# we expect 'tmp-clustername-tmpl', 'tmpl-clusternamespace-tmpl', and 'tmpl-clusteruuid-tmpl'
```

hack/addons/kustomize/cluster-autoscaler/manifests/helm-values.yaml
```diff
@@ -20,7 +21,7 @@ tolerations:
 
 # Limit a single cluster-autoscaler Deployment to a single Cluster.
 autoDiscovery:
-  clusterName: tmpl-clustername-tmpl
+  clusterName: tmpl-clusteruuid-tmpl
```
jimmidyson, line 24: This shouldn't be changing.
```suggestion
  clusterName: tmpl-clustername-tmpl
```

hack/addons/kustomize/cluster-autoscaler/manifests/helm-values.yaml
```diff
@@ -32,7 +33,7 @@ clusterAPIMode: kubeconfig-incluster
 clusterAPIWorkloadKubeconfigPath: /cluster/kubeconfig
 extraVolumeSecrets:
   kubeconfig:
-    name: tmpl-clustername-tmpl-kubeconfig
+    name: tmpl-clusteruuid-tmpl-kubeconfig
```
jimmidyson, line 36: This shouldn't be changing.
```suggestion
    name: tmpl-clustername-tmpl-kubeconfig
```

hack/addons/update-cluster-autoscaler.sh
```diff
@@ -19,12 +19,15 @@ trap_add "rm -rf ${ASSETS_DIR}" EXIT
...
-readonly KUSTOMIZE_BASE_DIR="${SCRIPT_DIR}/kustomize/cluster-autoscaler/"
+readonly KUSTOMIZE_BASE_DIR="${SCRIPT_DIR}/kustomize/cluster-autoscaler/manifests"
 envsubst -no-unset <"${KUSTOMIZE_BASE_DIR}/kustomization.yaml.tmpl" >"${ASSETS_DIR}/kustomization.yaml"
 cp "${KUSTOMIZE_BASE_DIR}"/*.yaml "${ASSETS_DIR}"
 
 kustomize build --enable-helm "${ASSETS_DIR}" >"${ASSETS_DIR}/${FILE_NAME}"
 
+# this replaces the custom templating used for kustomize to generate the configmap to using the same templating as we do
+# with helm chart proxy
+# transforming tmpl-clustername-tmpl to -- {{ .Cluster.Name }}
```
jimmidyson, line 30:
```suggestion
# transforming tmpl-clustername-tmpl to {{ .Cluster.Name }}
```

dlipovetsky:
Could you please rename the various `$NAME-template.yaml` files to `$NAME.yaml.tmpl`, to help signal that the files are intended to be processed by go's text/template?

jimmidyson:
As I've said to you before @faiq, my big problem with this approach is editing files that are literally copied to the chart directory in a different directory, even though both files are committed. I now have duplication for no reason in the codebase.

As an example, [charts/cluster-api-runtime-extensions-nutanix/templates/ccm/nutanix/manifests/helm-addon-installation.yaml](https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/819/files#diff-3d9c469be36f2fd6aca622895e1cb1dbfd3120c58059a2cd64df9533f3093d8b) is identical to [charts/cluster-api-runtime-extensions-nutanix/templates/ccm/nutanix/manifests/helm-addon-installation.yaml](https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/819/files#diff-70e53dd1807c7bf58c8ce259fc32094a2aec183ee41086782caca9b860d515a6) and you have to edit the file outside the chart to be synced to the chart directory.

I would much prefer the chart files to be edited in the charts directory - this is the charts source directory and editing the source somewhere else only to copy it back to source directory feels very odd to me, like editing go source in a different directory only to copy it to somewhere else to compile it, and committing both files to git.

I think this approach is really only here because we generate ClusterResourceSet configmap manifests for the addons. With the near term hope of removing ClusterResourceSet support and only relying on CAAPH for addon deployment I feel we will just have to undo this method and move the chart source back to the charts directory. This seems a big change for a short term problem.

I do like the change to pull the embedded go template files used for CAAPH values out to files (as I originally did in #809). This a big win as it removes the very confusing and error prone double embedded go template stuff.

jimmidyson:
I wonder if we can get the same benefits by putting the values templates only in the charts dir (to make me happy!) and just reference them there from the kustomizations via relative path? Would have to use the same more lax load restricter none we use when syncing examples, but I think that would remove them need for this values syncing completely while still gaining the other benefits of keeping values in sync between helm addon and crs (because they would reference the same values file from charts dir).

PR #823, opened 2024-07-26
build: fix amd64 binary used in arm64 image

dkoshkin:
The published CAREN images have an amd64 binary in the arm64 image. This fixes that by using `ko` directly instead of using GoReleaser's `kos`. 
Both the dev and prod `ko` builds are moved to `before.hooks`.

Fixes #

Published to a private repo and verified that the binary inside the arm64 image is an arm64 binary.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -52,6 +52,34 @@ before:
         ".releaseSeries |= (. + [{contract: \"v1beta1\", major: {{ .Major }}, minor: {{ .Minor }}}] | unique)" \
         metadata.yaml >release-metadata.yaml'
     - make template-helm-repository
+    # dev image
+    - |
+      sh -ec 'if [ {{ .IsSnapshot }} == true ] ; then
+        GOOS=linux \
+        SOURCE_DATE_EPOCH=$(date +%s) \
+        KO_DOCKER_REPO=ko.local/{{ .ProjectName }} \
+          ko build \
+          --bare \
+          --tags v{{ trimprefix .Version "v" }} \
+          ./cmd
+      fi'
+    # released image
+    - |
+      sh -ec 'if [ {{ .IsSnapshot }} == false ] ; then
+        unset GOOS GOARCH && \
+        SOURCE_DATE_EPOCH=$(date +%s) \
+        KO_DOCKER_REPO="dkoshkin/{{ .ProjectName }}" \
+        ko build \
+          --bare \
+          --platform linux/amd64,linux/arm64 \
+          --tags v{{ trimprefix .Version "v" }} \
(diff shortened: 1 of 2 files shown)


.ko.yaml
```diff
@@ -0,0 +1,10 @@
+defaultLdflagdss:
```
dkoshkin, line 1: Not possible to set multiple ldflags in the command line https://ko.build/advanced/faq/#how-can-i-set-ldflags

.goreleaser.yml
```diff
@@ -52,6 +52,34 @@ before:
...
+        SOURCE_DATE_EPOCH=$(date +%s) \
+        KO_DOCKER_REPO=ko.local/{{ .ProjectName }} \
+          ko build \
+          --bare \
+          --tags v{{ trimprefix .Version "v" }} \
+          ./cmd
+      fi'
+    # released image
+    - |
+      sh -ec 'if [ {{ .IsSnapshot }} == false ] ; then
```
dkoshkin, line 68: Moving it here means that the images are pushed before the artifacts are uploaded to Github. I think thats ok though since its already possible for one of these steps to fail and have a half broken release until its rerun.

.goreleaser.yml
```diff
@@ -52,6 +52,34 @@ before:
...
+        unset GOOS GOARCH && \
```
dkoshkin, line 69: ko complained about these envs being set together with the `--platform` flag.

.goreleaser.yml
```diff
@@ -52,6 +52,34 @@ before:
         ".releaseSeries |= (. + [{contract: \"v1beta1\", major: {{ .Major }}, minor: {{ .Minor }}}] | unique)" \
         metadata.yaml >release-metadata.yaml'
     - make template-helm-repository
+    # dev image
+    - |
+      sh -ec 'if [ {{ .IsSnapshot }} == true ] ; then
```
dkoshkin, line 57: Its probably possible to combine these 2 `ko` steps and based on the `.IsSnapshot` condition build either a single platform or multiple, but wanted to keep the scope of these changes small.

.goreleaser.yml
jimmidyson: I think this is the ultimate problem btw but I couldn't figure out why or how it made a difference 🤔

jimmidyson: https://github.com/ko-build/ko/issues/274

jimmidyson (changes requested):
Alternative approach in #824 which I think is a bit cleaner.

dkoshkin:
> Alternative approach in #824 which I think is a bit cleaner.

Thanks, I like your PR more, closing this one.

PR #824, opened 2024-07-26
build: Ensure clean go build environment for goreleaser and ko

jimmidyson:
Previous builds were overriding the `GOOS` and `GOARCH` enviroment
variables, which is fine to do when using `ko` directly, but appears to be
broken when building via ko using `goreleaser`.

This commit ensures that `GOOS` and `GOARCH` are always unset when
running builds by undefining them via `make` `undefine`. The `override`
keyword ensures that the env vars are always undefined regardless of the
source (e.g. enviroment, make variables, etc) to ensure a clean build
environment.

Instead of using `GOARCH` env var for building for the local build
architecure, this commit uses `goreleaser`'s `{{ .Runtime.Goarch }}`
variable available via templating which contains the architecture of the
build machine.

This commit also removes the `ldflags` config from the `ko`
configuration in `goreleaser` config as this is automatically inherited
from the `build` config referenced in the `ko` config.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -41,8 +41,8 @@ before:
         name: caren-system
       $(helm template {{ .ProjectName }} ./charts/{{ .ProjectName }} \
         --namespace caren-system \
-        --set-string image.tag=v{{ trimprefix .Version "v" }} \
-        --set-string helmRepositoryImage.tag=v{{ trimprefix .Version "v" }}{{ if .IsSnapshot }}-{{ .Env.GOARCH }} \
+        --set-string image.tag=v{{ trimprefix .Version "v" }}{{ if .IsSnapshot }}-{{ .Runtime.Goarch }}{{ end }} \
+        --set-string helmRepositoryImage.tag=v{{ trimprefix .Version "v" }}{{ if .IsSnapshot }}-{{ .Runtime.Goarch }} \
         --set-string image.repository=ko.local/{{ .ProjectName }}{{ end }} \
       )
       EOF'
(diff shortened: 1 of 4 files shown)


jimmidyson:
The error on Nutanix is to do with reserving the IPs for the CP endpoint:

```
failed to wait for task completion: failed to wait for task ZXJnb24=:2c4527f5-30b4-4562-98d7-c99345cb9410 to complete: context deadline exceeded
```

I think we may have a leak but this is blocking PR merges so we need to fix this ASAP.

.goreleaser.yml
```diff
@@ -80,12 +80,13 @@ builds:
...
-            env GOOS=linux GOARCH={{ .Arch }} \
-                SOURCE_DATE_EPOCH=$(date +%s) \
+            env SOURCE_DATE_EPOCH=$(date +%s) \
+                KO_DATA_DATE_EPOCH=$(date +%s) \
                 KO_DOCKER_REPO=ko.local/{{ .ProjectName }} \
                 ko build \
                   --bare \
-                  -t v{{ trimprefix .Version "v" }} \
+                  --platform linux/{{ .Arch }} \
+                  -t v{{ trimprefix .Version "v" }}-{{ .Arch }} \
```
dkoshkin, line 89: Thanks of adding the arch and making it consistent for the 2 images

dkoshkin (approved):
Tested locally on my arm Mac. The e2e tests are running on an amd64 Linux.

mhrabovcin (approved):

PR #836, opened 2024-08-01
build(deps): Upgrade CCM Version to v0.4.0

thunderboltsid:
https://github.com/nutanix-cloud-native/cloud-provider-nutanix/releases/tag/v0.4.0


dkoshkin:
How confident are we with this version? I'm a bit concerned about possible regressions.

thunderboltsid:
> How confident are we with this version? I'm a bit concerned about possible regressions.

I'm more confident with this version than without it. This contains changes for the long lived client cache w/ session auth. Without this we might end up in the same issue we did with CAPX DOS-ing PC IAM APIs.

dkoshkin (approved):

.github/workflows/checks.yml
```diff
@@ -244,7 +244,7 @@ jobs:
           devbox run -- \
             ct install \
               --config charts/ct-config.yaml \
-              --helm-extra-set-args "--set-string image.repository=ko.local/cluster-api-runtime-extensions-nutanix --set-string image.tag=$(devbox run -- gojq -r .version dist/metadata.json) --set-string helmRepositoryImage.tag=$(devbox run -- gojq -r .version dist/metadata.json)-$(devbox run -- go env GOARCH)"
+              --helm-extra-set-args "--set-string image.repository=ko.local/cluster-api-runtime-extensions-nutanix --set-string image.tag=$(devbox run -- gojq -r .version dist/metadata.json)-$(devbox run -- go env GOARCH) --set-string helmRepositoryImage.tag=$(devbox run -- gojq -r .version dist/metadata.json)-$(devbox run -- go env GOARCH)"
```
dkoshkin, line 247: This was missed in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/824 but didn't show up until now because this step only runs when there are changes to the helm chart.

dlipovetsky (dismissed):

deepakm-ntnx (commented):
I am a little skeptical to introduce this at this late stage. We should not change at this time if its not broken till we can observe it. 
Given CCM only runs on single cluster, and till we have prometheus counters to prove the count of APIs are high, I agree that we should atleast fix the part of using token instead of basic auth in general.
this fix also adds cache changes which technically will have only one cluster in it.

Also we need to identify the tests we need to do in order for ccm to get tested after every upgrade thru caren. 
- mainly providerID and machine instance recreation related cases (https://github.com/nutanix-cloud-native/cloud-provider-nutanix/blob/main/pkg/provider/instances_v2.go)

thunderboltsid:
> I am a little skeptical to introduce this at this late stage. We should not change at this time if its not broken till we can observe it. Given CCM only runs on single cluster, and till we have prometheus counters to prove the count of APIs are high, I agree that we should atleast fix the part of using token instead of basic auth in general. this fix also adds cache changes which technically will have only one cluster in it.
> 
> Also we need to identify the tests we need to do in order for ccm to get tested after every upgrade thru caren.
> 
> * mainly providerID and machine instance recreation related cases (https://github.com/nutanix-cloud-native/cloud-provider-nutanix/blob/main/pkg/provider/instances_v2.go)

I agree that it's coming in later than it should have. With that said, I'd like to make a case for why it is needed. The current helm chart version (v0.3.4) points to a released image that is over 1y old. The main functional change this bump brings about is introducing session auth changes to make sure we don't DOS PC with Basic auth requests. While we can be pedantic about correlating with instrumented metrics, in their absence, the best effort testing we did a couple of months with session auth changes on CAPX still showed meaningful difference in PC request metrics. Now, as for testing: 
- Any change that happens on CCM goes through a E2E test which ensures basic cluster CRUD works with that CCM version
- CAPX is already using v0.4.0 CCM in the E2Es and all CAPX E2Es that we run on PRs plus E2Es from periodic conformance tests are passing.
- We have a E2E as part of CAREN PRs which also ensures basic cluster CRUD works

deepakm-ntnx (approved):

PR #846, opened 2024-08-08
fix: Ensure MetalLB lifecycle handler is independent of other handlers

dlipovetsky:
Previously, the MetalLB handler created custom resources on the workload cluster. These required a webhook to be ready. The webhook required one or more Nodes to be Ready. That required CNI to be deployed.

When CAPI called the MetalLB handler before the CNI handler, this condition would never be satisified. No handlers after the MetalLB would be called.

With this PR, we delegate the creation of MetalLB configuration on the workload cluster to the ClusterResourceSet controller.

Fixes #

Deployed a Nutanix cluster.

### DIFF
--- a/test/e2e/serviceloadbalancer_helpers.go
+++ b/test/e2e/serviceloadbalancer_helpers.go
@@ -128,7 +128,7 @@ func waitForMetalLBServiceLoadBalancerToBeReadyInWorkloadCluster(
 
 	resources := make([]client.Object, len(cos))
 	for i := range cos {
-		resources[i] = cos[i]
+		resources[i] = cos[i].DeepCopy()
 	}
 
 	WaitForResources(ctx, WaitForResourcesInput{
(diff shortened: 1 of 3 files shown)


test/e2e/serviceloadbalancer_helpers.go
```diff
@@ -128,7 +128,7 @@ func waitForMetalLBServiceLoadBalancerToBeReadyInWorkloadCluster(
 
 	resources := make([]client.Object, len(cos))
 	for i := range cos {
-		resources[i] = cos[i]
+		resources[i] = cos[i].DeepCopy()
```
dlipovetsky, line 131: Note: Consequence of passing around a slice of unstructured.Unstructured, instead of a slice of pointers to unstructured.Unstructured.

dlipovetsky:
Closing in favor of #847

PR #847, opened 2024-08-08
fix: Deterministic ordering of lifecycle hooks

jimmidyson:
This commit introduces wrapper handlers to allow ordering
of lifecyce hooks to be called via a single registered hook.
This ensures that hooks are called in a defined order, fixing
issues that have arisen that require CNI hook to be applied
before metallb hook.

### DIFF
--- a/pkg/handlers/generic/lifecycle/handlers.go
+++ b/pkg/handlers/generic/lifecycle/handlers.go
@@ -4,11 +4,13 @@
 package lifecycle
 
 import (
+	"github.com/samber/lo"
 	"github.com/spf13/pflag"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 
 	"github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/api/v1alpha1"
 	"github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/common/pkg/capi/clustertopology/handlers"
+	"github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/common/pkg/capi/clustertopology/handlers/lifecycle"
 	"github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pkg/handlers/generic/lifecycle/ccm"
 	awsccm "github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pkg/handlers/generic/lifecycle/ccm/aws"
 	nutanixccm "github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pkg/handlers/generic/lifecycle/ccm/nutanix"
@@ -102,7 +104,7 @@ func (h *Handlers) AllHandlers(mgr manager.Manager) []handlers.Named {
 			helmChartInfoGetter,
 		),
 	}
-	return []handlers.Named{
+	allHandlers := []handlers.Named{
 		calico.New(mgr.GetClient(), h.calicoCNIConfig, helmChartInfoGetter),
 		cilium.New(mgr.GetClient(), h.ciliumCNIConfig, helmChartInfoGetter),
 		nfd.New(mgr.GetClient(), h.nfdConfig, helmChartInfoGetter),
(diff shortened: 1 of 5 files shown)


dlipovetsky:
With one handler for a specific hook type running "sub-handlers" sequentially, we'll need to be careful to make sure the handler responds within the [CAPI deadline of 30s](https://cluster-api.sigs.k8s.io/tasks/experimental-features/runtime-sdk/implement-extensions#timeouts).

We talked about running "sub-handlers" in parallel, and we can do that in follow-up PR.

dkoshkin:
On `main` I see that timeout is 10seconds, am I looking at the correct place?

```
apiVersion: runtime.cluster.x-k8s.io/v1alpha1
kind: ExtensionConfig
metadata:
  annotations:
    runtime.cluster.x-k8s.io/inject-ca-from-secret: caren-system/cluster-api-runtime-extensions-nutanix-runtimehooks-tls
  creationTimestamp: "2024-08-08T22:21:32Z"
  generation: 2
  labels:
    cluster.x-k8s.io/provider: runtime-extension-caren
    clusterctl.cluster.x-k8s.io: ""
  name: cluster-api-runtime-extensions-nutanix
  resourceVersion: "4431"
  uid: 28e66cd5-b136-4f4b-b818-4c0037e6930f
status:
  - failurePolicy: Fail
    name: serviceloadbalancerhandler-acpi.cluster-api-runtime-extensions-nutanix
    requestHook:
      apiVersion: hooks.runtime.cluster.x-k8s.io/v1alpha1
      hook: AfterControlPlaneInitialized
    timeoutSeconds: 10
...
```

dlipovetsky:
> On `main` I see that timeout is 10seconds, am I looking at the correct place?

Good catch. Those 10s timeouts make sense for individual handlers, but we should increase it, perhaps to 30s, for the one aggregate handler.

dlipovetsky:
On second thought, if the sub-handler fails, no other sub-handlers will run, so shortening the timeout isn't useful. But it will be necessary if/when we execute sub-handlers in parallel.

dlipovetsky:
I'll reduce the MetalLB timeout to 10s, since the handler is already limited to 10s by CAPI's default. We'll want to raise the handler timeout to 30s, but in a way that fits with the library, and that'll need some thought.

dlipovetsky (approved):

dkoshkin (approved):
Verified in the logs that the handlers are called in the expected order and the HCPs are also created in the same order:
(Notice the metallb is last) 
```
$ kubectl get hcp -w 
NAME                                             READY   REASON
cilium-cni-installation-dkoshkin-hook-ordering           
nutanix-ccm-dkoshkin-hook-ordering                       
node-feature-discovery-dkoshkin-hook-ordering                              
cluster-autoscaler-dkoshkin-hook-ordering                
nutanix-csi-dkoshkin-hook-ordering                       
snapshot-controller-dkoshkin-hook-ordering               
metallb-dkoshkin-hook-ordering                           
```

pkg/handlers/generic/lifecycle/handlers.go
```diff
@@ -102,17 +104,101 @@ func (h *Handlers) AllHandlers(mgr manager.Manager) []handlers.Named {
...
-		ccm.New(mgr.GetClient(), ccmHandlers),
-		serviceloadbalancer.New(mgr.GetClient(), serviceLoadBalancerHandlers),
 		snapshotcontroller.New(mgr.GetClient(), h.snapshotControllerConfig, helmChartInfoGetter),
+		servicelbgc.New(mgr.GetClient()),
+		// The order of the handlers in the list is important and are called consecutively.
+		// The MetalLB provider may be configured to create a IPAddressPool on the remote cluster.
+		// However, the MetalLB provider also has a webhook that validates IPAddressPool requests.
+		// Because this webhook relies on CNI and CCM to already be installed on the remote cluster,
+		// we are placing this handler after the CNI and CCM handlers.
+		serviceloadbalancer.New(mgr.GetClient(), serviceLoadBalancerHandlers),
```
dkoshkin, line 121: I pushed a commit to reorder the handlers with the `serviceloadbalancer` going last, and moved CCM higher up since the Nodes won't become ready until CCM runs.

If we don't want this change here I'm happy to remove this commit

jimmidyson: This is a good change! Once we change this to run all handlers and aggregate even even failed responses then this won't be a problem.

dlipovetsky: I think it's a good change, but since this PR didn't re-order handlers to begin with, I think it would be better in a separate PR. I appreciate the explanatory comment, and I can add more of them.

dkoshkin (approved):

PR #845, opened 2024-08-07
fix: Handle long cluster names

jimmidyson:
This commit starts fixing a bug that means
addons fail to be fully deployed if the cluster name
is longer than 44 characters. This is caused by the name
of the HCP being over 63 characters. This name is then
used in HRP labels which have a maximum length of 63
characters, so the HRPs are rejected by the API server
when CAAPH applies them.

The fix is to add a UUID annotation to all Cluster resources,
enforce this as immutable, and use that UUID in HCP names.

PR #848, opened 2024-08-09
fix: reorder lifecycle handlers with serviceloadbalancer being last

dkoshkin:
The lifecycle handlers are called in the order defined (after https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/845). The MetalLB serviceloadbalancer has a dependency on CNI and CCM to run first and start its deployment process.
This PR is a small optimization that moves CCM higher in the list and the serviceloadbalancer to be last with a comment on why its last.

Fixes #


dlipovetsky (approved):
Thank you!

jimmidyson (approved):

PR #855, opened 2024-08-13
feat: Run hooks in parallel with aggregated responses

jimmidyson:
This commit runs lifecycle hooks in parallel and aggregates the results. All hooks are run and responses are aggregated. This removes the ordering of hooks, as any single failure will result in all hooks being called again by CAPI cluster topology controller.


api/go.mod
```diff
@@ -3,7 +3,7 @@
 
 module github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/api
 
-go 1.21
+go 1.22
```
dkoshkin, line 6: Curious, is the Go bump needed here?

common/pkg/capi/clustertopology/handlers/lifecycle/parallel.go
```diff
@@ -0,0 +1,292 @@
...
+	for _, f := range hookFuncs {
+		hookResponse := reflect.New(responsePtrType).Interface().(U)
+		go func() {
+			f(ctx, request, hookResponse)
+			responseChan <- hookResponse
+			wg.Done()
+		}()
+	}
+
+	// Important final goroutine to wait until WaitGroup is done
```
dkoshkin, line 36: Just checking, each hook runs independently and allows to either succeed, error or timeout?

common/pkg/capi/clustertopology/handlers/lifecycle/parallel.go
```diff
@@ -0,0 +1,292 @@
...
+	"reflect"
+	"strings"
+	"sync"
+
+	"github.com/samber/lo"
+	runtimehooksv1 "sigs.k8s.io/cluster-api/exp/runtime/hooks/api/v1alpha1"
+	"sigs.k8s.io/cluster-api/util"
+)
+
+func runHooksInParallel[T runtimehooksv1.RequestObject, U runtimehooksv1.ResponseObject](
```
dkoshkin, line 17: One downside of this is that the logs are interleaved from different hooks. Any other potential downside to this?

common/pkg/capi/clustertopology/handlers/lifecycle/parallel.go
jimmidyson: Yes, I've added comments throughout this function now.

common/pkg/capi/clustertopology/handlers/lifecycle/parallel.go
jimmidyson: I can't think of any.

We can improve the logging by adding hook names to the log output to make the interleaved logs more readable, or rather to be able to trace through the individual hooks by filtering the logs. Generally the lifecycle hooks have few logs so it shouldn't cause any issue for now though.

api/go.mod
jimmidyson: Yes, this was needed to ensure go 1.22 is used across the project. Having it set to 1.21 caused inconsistent language versions to be used, for example for loop variables scoping was still in 1.21 mode in the test i wrote here, which led to issues that I couldn't figure out for a while.

Thinking about this again, i think I could not bump these here and instead use `//go:build` directives in the necessary test files, but I think that is more opaque so would prefer to keep it consistent at go 1.22 in go mods.

Let me know if you can see any problems with that.

common/pkg/capi/clustertopology/handlers/lifecycle/parallel.go
```diff
@@ -0,0 +1,325 @@
...
+	wg.Add(len(hookFuncs))
+
+	// Run each hook in a new goroutine, send the response to the response channel and decrement the
+	// wait group counter when the hook finishes.
+	for _, f := range hookFuncs {
+		// Create a new instance of the response object type for each hook. Due to the use of generics,
+		// we need to use reflection to create a new instance of the response object type.
+		hookResponse := reflect.New(responsePtrType).Interface().(U)
+		go func() {
+			// Ru the hook function.
```
mhrabovcin, line 39:
```suggestion
			// Run the hook function.
```

jimmidyson: Done.

mhrabovcin (approved):

dkoshkin (approved):

dkoshkin (approved):

PR #856, opened 2024-08-14
fix: Cilium-Istio compatibility fixes

dkoshkin:

Setting  `cni.exclusive: false` and `socketLB.hostNamespaceOnly: true`.

See Cilium's docs https://docs.cilium.io/en/latest/network/servicemesh/istio/. Without these value Cilium can interfere with Istio functionality, by always cleaning up cni config directory on the host, thus preventing Istio Pods from coming up, and interfere with Istio load-balancing once they do come up.

It's safe to always set these value, because it is not Cilium's responsibility to prevent other applications from acting like a network plugin.

Fixes #


msdolbey (approved):
👍 
We're not seeing each other exclusively, I thought

jimmidyson (changes requested):
As per https://docs.cilium.io/en/latest/network/servicemesh/istio/#cilium-configuration let's also set `socketLB.hostNamespaceOnly: true`

jimmidyson (approved):

mhrabovcin (approved):

PR #857, opened 2024-08-14
build: Latest devbox update

jimmidyson:

Fixes #

### DIFF
--- a/charts/cluster-api-runtime-extensions-nutanix/templates/role.yaml
+++ b/charts/cluster-api-runtime-extensions-nutanix/templates/role.yaml
@@ -22,13 +22,6 @@ rules:
       - ""
     resources:
       - namespaces
-    verbs:
-      - get
-      - list
-      - watch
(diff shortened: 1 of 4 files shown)


mhrabovcin (approved):

charts/cluster-api-runtime-extensions-nutanix/templates/role.yaml
```diff
@@ -49,17 +42,6 @@ rules:
       - addons.cluster.x-k8s.io
     resources:
       - clusterresourcesets
-    verbs:
```
faiq, line 52: why is this changed?

faiq (approved):

charts/cluster-api-runtime-extensions-nutanix/templates/role.yaml
jimmidyson: Changes in controller-gen I guess - this is a generated file.

PR #860, opened 2024-08-15
fix: skip UUID annotation webhook for clusters with nil topology

dkoshkin:
We only want this webhook to run on Clusters that are created with cluster topology. This also fixes an issue when an existing Cluster that was created before deploying CAREN and this webhook is moved to the bootstrap cluster.
Currently it fails the move with 
```
  2024-08-14 11:14:55 ERR     err="unable to pivot to the to-cluster: error setting Cluster.Spec.Paused=true: action failed after 10 attempts: error patching Cluster default/e2e-preprovisioned-upgrade-573225644: admission webhook "cluster-validator.caren.nutanix.com" denied the request: missing cluster UUID annotation caren.nutanix.com/cluster-uuid"
```

Fixes #

I'm also now wondering if we want the validate to fail when the annotation is missing. Its possible that an existing Cluster that was created using another ClusterClass that does not use CAREN and then when CAREN gets installed, it prevents those existing Clusters from being updated because of a missing annotation that won't even be used.
**Should this webhook be smarter and only validate and default Clusters that are referencing a CC that is using CAREN?**

### DIFF
--- a/internal/test/envtest/environment.go
+++ b/internal/test/envtest/environment.go
@@ -83,7 +83,7 @@ func init() {
 // RunInput is the input for Run.
 type RunInput struct {
 	M                     *testing.M
-	ManagerUncachedObjs   []client.Object
+	EnvironmentOpts       []EnvironmentOpt
 	SetupReconcilers      func(ctx context.Context, mgr ctrl.Manager)
 	SetupEnv              func(e *Environment)
 	WebhookInstallOptions envtest.WebhookInstallOptions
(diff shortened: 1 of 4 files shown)


jimmidyson:
> I'm also now wondering if we want the validate to fail when the annotation is missing. Its possible that an existing Cluster that was created using another ClusterClass that does not use CAREN and then when CAREN gets installed, it prevents those existing Clusters from being updated because of a missing annotation that won't even be used.

I've updated the webhook to handle this use case by adding the annotation on update to clusters that use topology but do not have a UUID annotation already set. This wil handle clusters that are migrated to CAREN as well.

> Should this webhook be smarter and only validate and default Clusters that are referencing a CC that is using CAREN?

This is pretty hard to figure out tbh so I don't think so, for now at least. Adding the annotation doesn't harm anything.

Ultimately I feel that this stable UUID should be part of CAPI core as an official annotation and will open an issue there to see if we can get traction.

jimmidyson (approved):

mhrabovcin (approved):

PR #863, opened 2024-08-15
build: Use new clusterctl-aws name for clusterawsadm

jimmidyson:
This was renamed as part of CAPA v2.6.0 which releases both
clusterctl-aws and clusterawsadm. However clusterawsadm
will be removed in the near future so make the change here
now so we are prepared for it.


dkoshkin (approved):

hack/flakes/flake.nix
```diff
@@ -57,8 +57,8 @@
             ldflags = [ "-s" "-w" ];
           };
 
-          clusterawsadm = buildGo122Module rec {
-            name = "clusterawsadm";
+          clusterctl-aws = buildGo122Module rec {
```
faiq, line 60: why do we always have to build these?

jimmidyson: They're not available from the nixpkgs in binary form. I did have a way to download these via nix from GH releases for packages but I hit problems so dropped back to building them.

faiq: The thing is with building all these packages is that they take up a lot of resources on my CPU on first build. Not a blocker or anything I was just wondering.

dlipovetsky (approved):
Thanks

hack/flakes/flake.nix
jimmidyson: I'm thinking of different ways to do this (try to figure out nix language and provide an easy way to package github releases so no need to build; or use a nix remote cache) but in the meantime can we merge this and I can look at that separately?

jimmidyson: Gonna merge and look at this again later.

PR #864, opened 2024-08-15
fix: Ensure ClusterAutoscaler can write status CM in workload cluster

jimmidyson:
Previously workload ClusterAutoscaler deployments were configured to
write the status configmap to the same namesapce as the workload cluster
is in on the management cluster. However, ClusterAutoscaler actually
writes this configmap and acquires a lock in the specified namespace in
the workload cluster. If that namespace does not exist then CA will fail
to run correctly, repeatedly throwing the error `error initially creating leader
election record: namespaces "my-new-ns" not found`.

This commit updates the target namespace to use the default
`kube-system` which always exists on all clusters. This is the CA
default and as such is completely safe to use.


deepakm-ntnx:
hi @jimmidyson, just for my knowledge, will this scenario get tested with any of the tests thru the PR pre-commit tests?

jimmidyson:
@deepakm-ntnx Sadly not right now, we need to add an e2e that creates a self-managed management cluster (we have that already) and then creates a workload cluster in a new namespace on that management cluster. An extension of the existing self-managed test suite... but no we don't have that yet.

I and @dkoshkin have manually tested it though.

dkoshkin (approved):
Tested on a worklaod cluster and observed that the CM is getting created in the `kube-system` namespace.

msdolbey (approved):
👍 
It's always the system, isn't it

deepakm-ntnx (approved):

thunderboltsid (approved):

PR #867, opened 2024-08-15
fix: CRS generated CA Deployment has extra quotes

dkoshkin:
Fixes an issue with CRS genertated CA Deployment not working because of extra quotes.
```
I0815 20:12:58.376871       1 reflector.go:332] Listing and watching cluster.x-k8s.io/v1beta1, Resource=machines from k8s.io/client-go/dynamic/dynamicinformer/informer.go:108
W0815 20:12:58.379140       1 reflector.go:547] k8s.io/client-go/dynamic/dynamicinformer/informer.go:108: failed to list cluster.x-k8s.io/v1beta1, Resource=machines: machines.cluster.x-k8s.io is forbidden: User "system:serviceaccount:default:cluster-autoscaler-0191579a-2104-7ace-a5a2-ceae4590d7fe" cannot list resource "machines" in API group "cluster.x-k8s.io" in the namespace "'default'"
E0815 20:12:58.379170       1 reflector.go:150] k8s.io/client-go/dynamic/dynamicinformer/informer.go:108: Failed to watch cluster.x-k8s.io/v1beta1, Resource=machines: failed to list cluster.x-k8s.io/v1beta1, Resource=machines: machines.cluster.x-k8s.io is forbidden: User "system:serviceaccount:default:cluster-autoscaler-0191579a-2104-7ace-a5a2-ceae4590d7fe" cannot list resource "machines" in API group "cluster.x-k8s.io" in the namespace "'default'"
```

The extra quotes are only an issue in `--node-group-auto-discovery=`, but the [same script](https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/f632224257cb159b04abc2c6c6eb6874c503bb1c/hack/addons/update-cluster-autoscaler.sh#L28) replaces all occurrences. It should be safe to drop the single quotes everywhere though because the name and namespace will be strings and won't be interpreted as numbers by yaml.

Fixes #

~I think we can improve the e2e tests by checking that all Deployments are Ready on the self-managed clusters, but I did not do that as part of this PR.~ The tests already do that https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/f632224257cb159b04abc2c6c6eb6874c503bb1c/test/e2e/clusterautoscaler_helpers.go#L115

Instead we can also wait for the status ConfigMap to be `Running`
This is what the data will contain for a working Deployment:
```
data:
  status: |+
    time: 2024-05-22 19:33:34.074058252 +0000 UTC
    autoscalerStatus: Running
```
And for non working:
```
data:
  status: |+
    time: 2024-08-15 20:07:37.204175116 +0000 UTC
    autoscalerStatus: Initializing
```

### DIFF
--- a/hack/addons/update-cluster-autoscaler.sh
+++ b/hack/addons/update-cluster-autoscaler.sh
@@ -25,7 +25,7 @@ cp "${KUSTOMIZE_BASE_DIR}"/*.yaml "${ASSETS_DIR}"
 
 kustomize build --enable-helm "${ASSETS_DIR}" >"${ASSETS_DIR}/${FILE_NAME}"
 
-sed -i -e "s/\([a-z-]*\)tmpl-clustername\([^-]*\)-tmpl\([a-z-]*\)/'\1{{ \`{{ .Cluster.Name\2 }}\` }}\3'/g" \
+sed -i -e "s/\([a-z-]*\)tmpl-clustername\([^-]*\)-tmpl\([a-z-]*\)/\1{{ \`{{ .Cluster.Name\2 }}\` }}\3/g" \
   -e "s/\([a-z-]*\)tmpl-clusteruuid-tmpl\([a-z-]*\)/'\1{{ \`{{ index .Cluster.Annotations \"caren.nutanix.com\/cluster-uuid\" }}\` }}\2'/g" \
   "${ASSETS_DIR}/${FILE_NAME}"
 
--- a/hack/addons/update-cluster-autoscaler.sh
+++ b/hack/addons/update-cluster-autoscaler.sh
@@ -26,7 +26,7 @@ cp "${KUSTOMIZE_BASE_DIR}"/*.yaml "${ASSETS_DIR}"
 kustomize build --enable-helm "${ASSETS_DIR}" >"${ASSETS_DIR}/${FILE_NAME}"
 
 sed -i -e "s/\([a-z-]*\)tmpl-clustername\([^-]*\)-tmpl\([a-z-]*\)/\1{{ \`{{ .Cluster.Name\2 }}\` }}\3/g" \
-  -e "s/\([a-z-]*\)tmpl-clusteruuid-tmpl\([a-z-]*\)/'\1{{ \`{{ index .Cluster.Annotations \"caren.nutanix.com\/cluster-uuid\" }}\` }}\2'/g" \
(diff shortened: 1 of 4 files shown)


hack/addons/update-cluster-autoscaler.sh
```diff
@@ -25,7 +25,7 @@ cp "${KUSTOMIZE_BASE_DIR}"/*.yaml "${ASSETS_DIR}"
 
 kustomize build --enable-helm "${ASSETS_DIR}" >"${ASSETS_DIR}/${FILE_NAME}"
 
-sed -i -e "s/\([a-z-]*\)tmpl-clustername\([^-]*\)-tmpl\([a-z-]*\)/'\1{{ \`{{ .Cluster.Name\2 }}\` }}\3'/g" \
+sed -i -e "s/\([a-z-]*\)tmpl-clustername\([^-]*\)-tmpl\([a-z-]*\)/\1{{ \`{{ .Cluster.Name\2 }}\` }}\3/g" \
   -e "s/\([a-z-]*\)tmpl-clusteruuid-tmpl\([a-z-]*\)/'\1{{ \`{{ index .Cluster.Annotations \"caren.nutanix.com\/cluster-uuid\" }}\` }}\2'/g" \
```
jimmidyson, line 29: Should we remove the single quotes here too?

jimmidyson:
@dkoshkin So with this error the deployment shows as ready but it's not working properly so we should check the status configmap too?

dkoshkin:
> @dkoshkin So with this error the deployment shows as ready but it's not working properly so we should check the status configmap too?

Yep! Attempt on the check here https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/867/commits/b447ba4bea5211afc649ea0de48fa96d8ce4df79.

dkoshkin:
The new ConfigMap test failed as expected without this fix https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/868#issuecomment-2292675203

hack/addons/update-cluster-autoscaler.sh
dkoshkin: Good idea, lets try it. The new e2e check should tell us if its still working.

jimmidyson (approved):
Thank you for fixing it and improving the tests!

PR #874, opened 2024-08-21
fix: Properly set additionalTrustBundle for Nutanix CCM

jimmidyson:
The helm values template contained an error meaning that the
additionalTrustBundle setting would have been templated as part of the
Helm install, rather than the CAAPH addon deployment, meaning that
specifying an additional trust bundle would not work for Nutanix CCM
deployments.

This commit fixes that and adds tests that use the same template as the
CAAPH addon would do from the Helm chart to ensure that the template is
now correct.


charts/cluster-api-runtime-extensions-nutanix/templates/ccm/nutanix/manifests/helm-addon-installation.yaml
```diff
@@ -12,9 +12,9 @@ data:
     prismCentralEndPoint: {{ `{{ .PrismCentralHost }}` }}
     prismCentralPort: {{ `{{ .PrismCentralPort }}` }}
     prismCentralInsecure: {{ `{{ .PrismCentralInsecure }}` }}
-    {{- with .PrismCentralAdditionalTrustBundle }}
-    prismCentralAdditionalTrustBundle: {{ `{{ . }}` }}
-    {{- end }}
+    {{ `{{- with .PrismCentralAdditionalTrustBundle }}` }}
+    prismCentralAdditionalTrustBundle: {{ `{{ printf "%q" . }}` }}
+    {{ `{{- end }}` }}
```
faiq, line 17: this double inception stuff is so difficult to read.

pkg/handlers/generic/lifecycle/ccm/nutanix/handler_test.go
```diff
@@ -39,24 +37,61 @@ prismCentralAdditionalTrustBundle: "LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSUVj
...
+		t,
+		template.Must(
+			template.New(
+				"helm-addon-installation.yaml").ParseFiles(templateFile),
+		).Execute(&templatedBytes, templateData),
+	)
+	cm := &corev1.ConfigMap{}
+	require.NoError(t, yaml.UnmarshalStrict(templatedBytes.Bytes(), cm))
+
+	valuesTemplate := cm.Data["values.yaml"]
```
faiq, line 93: i honestly am very confused by whats going on here.

jimmidyson: Reading the configmap from the charts dir to get the actual template as would be rendered by Helm.

faiq: oh ok i get it now. 

can you add more comments making this clear. alternatively you dont have to use the unstructured to set the helm bits and can extract the template some other way. 

definetly the changes in #819 would make things easier here :smile:

mhrabovcin (approved):

PR #876, opened 2024-08-21
revert: Revert upgrades for CAPI 1.8

jimmidyson:
This needs to remain at 1.7 for compatibility with other projects
and will be upgraded again in a future minor release of CAREN.


dlipovetsky (approved):
Thank you

faiq (approved):

PR #888, opened 2024-08-26
build: Update patch only for local release branch dev builds

jimmidyson:
Prevents confusion over what version is being built locally.


dkoshkin (approved):

dkoshkin:
Should we make the same change on main too?

jimmidyson:
> Should we make the same change on main too?

No, we want main to do proper bumps for conventional commits, e.g. minor bump for feat commits.

mhrabovcin (approved):

PR #889, opened 2024-08-26
feat: Build with go 1.23

jimmidyson:


mhrabovcin (approved):

api/v1alpha1/nutanix_clusterconfig_types.go
```diff
@@ -55,11 +55,11 @@ type NutanixPrismCentralEndpointCredentials struct {
 }
 
 //nolint:gocritic // No need for named return values
-func (s NutanixPrismCentralEndpointSpec) ParseURL() (string, int32, error) {
+func (s NutanixPrismCentralEndpointSpec) ParseURL() (string, uint16, error) {
```
dkoshkin, line 58: Why is this func changing as part of this PR?

jimmidyson: Updated golangci-lint using new gosec has a very aggressive check for inteher overflows. This change made sense considering it's supposed to be a port number which is a uint16. Do you have concerns about it? I see now that it is a public API breaking change...

jimmidyson: Do you see this as a problem or just out of interest?

dkoshkin: > I see now that it is a public API breaking change...

I think thats ok for an alpha API since it's only breaking when using it a Go library, and not with yaml.

PR #895, opened 2024-08-29
fix: Remove deprecated toleration for node-role.kubernetes.io/master

jimmidyson:
This was removed in Kubernetes v1.24 so is no longer needed and actually
throws a warning now if it is kept when applying manifests:

```
[KubeAPIWarningLogger] spec.template.spec.affinity.nodeAffinity.preferredDuringSchedulingIgnoredDuringExecution[1].preference.matchExpressions[0].key: node-role.kubernetes.io/master is use "node-role.kubernetes.io/control-plane" instead
```

Removal note in https://kubernetes.io/blog/2022/04/07/upcoming-changes-in-kubernetes-1-24/#api-removals-deprecations-and-other-changes-for-kubernetes-1-24.

Blocked by #893.


dlipovetsky (approved):
Thanks for cleaning this up.

mhrabovcin (approved):

PR #896, opened 2024-08-30
feat: Extract CAAPH values templates to files

jimmidyson:
This simplifies the Helm templating directives by not requiring
inception-style escaping of templating braces, e.g. `{{ "{{" }}``
which are very hard to read and can introduce bugs.

This PR also removes the duplicate helm values files currently being used
to generate the CRS configmaps, and instead references the helm values
that are in the charts directory, which ends up with a reduction in LOC in
the project to maintain.

I feel this is a simpler way to achieve the same goals as #819 but without
duplicating files and keeping all chart source files in the charts directory.

Blocked by #895.


charts/cluster-api-runtime-extensions-nutanix/addons/ccm/aws/map-template.yaml
```diff
@@ -0,0 +1,8 @@
+# These values are a combination of a Helm template to create the dict for the k8s version to
+# CCM version map, and then including the non-templated values from the values-template.yaml.
+# This simplifies the templating required in the values-template.yaml.
+$k8sMinorVersionToCCMVersion := dict
+{{ range $k8sVersion, $ccmVersion := .Values.hooks.ccm.aws.k8sMinorVersionToCCMVersion -}}
+  "{{ $k8sVersion }}" "{{ $ccmVersion }}"
+supershal marked this conversation as resolved.
```
faiq, line 7: ?

jimmidyson: Oops! Copy paste error 🤣

pkg/handlers/generic/lifecycle/ccm/nutanix/handler_test.go
```diff
@@ -147,38 +143,12 @@ func Test_templateValues(t *testing.T) {
 // and renders it with dummy values, finally extracting the embedded template that will be used by
 // CAAPH when installing the Nutanix CCM addon.
```
faiq, line 144: this comment is no longer true

jimmidyson: Fixed up.

hack/addons/update-cluster-autoscaler.sh
```diff
@@ -21,7 +21,16 @@ readonly FILE_NAME="cluster-autoscaler.yaml"
...
+cat <<EOF >"${ASSETS_DIR}/gomplate-context.yaml"
+Cluster:
+  Name: tmpl-clustername-tmpl
+  Namespace: tmpl-clusternamespace-tmpl
+  Annotations:
+    caren.nutanix.com/cluster-uuid: tmpl-clusteruuid-tmpl
+EOF
+gomplate -f "${GIT_REPO_ROOT}/charts/cluster-api-runtime-extensions-nutanix/addons/cluster-autoscaler/values-template.yaml" \
+  --context .="${ASSETS_DIR}/gomplate-context.yaml" \
+  >"${ASSETS_DIR}/helm-values.yaml"
```
faiq, line 33: its not clear that `helm-values.yaml` is generated. 

maybe we should include it in `.gitignore` so that people don't modify it

jimmidyson: In this case it's written to a temp directory which is deleted when this script exits. It is never in the repo or working copy.

hack/addons/kustomize/cilium/kustomization.yaml.tmpl
```diff
@@ -15,10 +15,13 @@ helmCharts:
   repo: https://helm.cilium.io/
   releaseName: cilium
   version: ${CILIUM_VERSION}
-  valuesFile: helm-values.yaml
+  valuesFile: ../../../../charts/cluster-api-runtime-extensions-nutanix/addons/cni/cilium/values-template.yaml
   includeCRDs: true
   skipTests: true
   namespace: kube-system
   kubeVersion: ${E2E_KUBERNETES_VERSION}
+  valuesInline:
```
dkoshkin, line 23: Please add a comment why this is needed in CRS.

dkoshkin: Maybe we should also put this next to `valuesFile`?

pkg/handlers/generic/lifecycle/ccm/nutanix/handler_test.go
```diff
@@ -143,42 +139,14 @@ func Test_templateValues(t *testing.T) {
 	}
 }
 
-// readCCMTemplateFromProjectHelmChart gets the CCM template from the Helm chart in the project
-// and renders it with dummy values, finally extracting the embedded template that will be used by
-// CAAPH when installing the Nutanix CCM addon.
+// readCCMValuesTemplateFromProjectHelmChart reads the requested Helm values file from the Helm chart in the project.
 // This is important to do this way to ensure that the hard-to-read double templating works as expected.
-func readCCMTemplateFromProjectHelmChart(t *testing.T) string {
+func readCCMValuesTemplateFromProjectHelmChart(t *testing.T) string {
```
dkoshkin, line 144: Nice simplification

hack/addons/update-cluster-autoscaler.sh
```diff
@@ -21,7 +21,16 @@ readonly FILE_NAME="cluster-autoscaler.yaml"
 
 readonly KUSTOMIZE_BASE_DIR="${SCRIPT_DIR}/kustomize/cluster-autoscaler/"
 envsubst -no-unset <"${KUSTOMIZE_BASE_DIR}/kustomization.yaml.tmpl" >"${ASSETS_DIR}/kustomization.yaml"
-cp "${KUSTOMIZE_BASE_DIR}"/*.yaml "${ASSETS_DIR}"
+cat <<EOF >"${ASSETS_DIR}/gomplate-context.yaml"
```
dkoshkin, line 24: I don't think I understand what this is doing.
Can you please add a comment to leave some hints for future us why this is needed and why this is different from the other scripts.

dkoshkin (approved):
Looking great, just some minor nits.
Thanks @jimmidyson and @faiq for the team effort here.

hack/addons/kustomize/cilium/kustomization.yaml.tmpl
jimmidyson: Added a comment and moved values file and values inline next to each other.

hack/addons/update-cluster-autoscaler.sh
jimmidyson: Added comment.

hack/addons/kustomize/tigera-operator/kustomization.yaml.tmpl
```diff
@@ -18,7 +18,9 @@ helmCharts:
   repo: https://docs.tigera.io/calico/charts
   releaseName: tigera-operator
   version: ${CALICO_VERSION}
-  valuesFile: helm-values.yaml
+  valuesInline:
+    installation:
+      enabled: false
```
dlipovetsky, line 23: I don't understand where this was expressed previous to the PR.

jimmidyson: This was in the helm values file that was referenced in the deleted line above, referencing deleted file https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/896/files/6e0f6c29799a81a8db9f1a11e9e0365ded496b49#diff-20b1d5135d69f7710be48ad71740bcc354f53c9b8302e5d19b0540e407cc4c0e (expand to see previous contents which include this value). Moving this inline here makes the scripts simpler.

dlipovetsky: Thanks! Annoying of GitHub to hide that.

faiq (approved):

PR #822, opened 2024-07-25
ci: images tool 

faiq:
**What problem does this PR solve?**:
Adds a tool to fetch all images CAREN uses

**Which issue(s) this PR fixes**:
Fixes #

**How Has This Been Tested?**:

I ran the following command from `hack/tools/fetch-images`

```
$ go run main.go -chart-directory=../../../charts/cluster-api-runtime-extensions-nutanix/ --helm-chart-configmap=../../../charts/cluster-api-runtime-extensions-nutanix/templates/helm-config.yaml -caren-version=v0.13.1
ghcr.io/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix:v0.13.1
ghcr.io/nutanix-cloud-native/caren-helm-reg:v0.13.1
registry.k8s.io/sig-storage/snapshot-controller:v8.0.1
registry.k8s.io/provider-aws/cloud-controller-manager:v1.27.7
registry.k8s.io/provider-aws/cloud-controller-manager:v1.28.6
registry.k8s.io/provider-aws/cloud-controller-manager:v1.29.3
registry.k8s.io/provider-aws/cloud-controller-manager:v1.30.1
public.ecr.aws/ebs-csi-driver/aws-ebs-csi-driver:v1.33.0
public.ecr.aws/eks-distro/kubernetes-csi/node-driver-registrar:v2.11.0-eks-1-30-10
public.ecr.aws/eks-distro/kubernetes-csi/livenessprobe:v2.13.0-eks-1-30-10
public.ecr.aws/ebs-csi-driver/aws-ebs-csi-driver:v1.33.0
public.ecr.aws/eks-distro/kubernetes-csi/external-provisioner:v5.0.1-eks-1-30-10
public.ecr.aws/eks-distro/kubernetes-csi/external-attacher:v4.6.1-eks-1-30-10
public.ecr.aws/eks-distro/kubernetes-csi/external-resizer:v1.11.1-eks-1-30-10
public.ecr.aws/eks-distro/kubernetes-csi/livenessprobe:v2.13.0-eks-1-30-10
registry.k8s.io/autoscaling/cluster-autoscaler:v1.30.0
quay.io/metallb/speaker:v0.14.8
quay.io/frrouting/frr:9.1.0
quay.io/metallb/speaker:v0.14.8
quay.io/metallb/controller:v0.14.8
registry.k8s.io/nfd/node-feature-discovery:v0.16.1-minimal
[...]

PR #900, opened 2024-09-05
build: Include Calico images in image list

jimmidyson:
Calico images are deployed by the tigera-operator and as such are not
discoverable by the helm list images plugin. Using the extra images file
functionality of the helm list images plugin allows for adding these
images as templated strings that will adapt with the chart version as it
is updated in future.

This commit also fixes up some other things:

- Use `filepath` instead of `path` for portabiity
- Output the image list to stdout from make to allow user to decide
  where to redirect list to
- Omit changing dir from image list when running as goreleaser hook
- Slightly more defensive when checking errors

Final output is:

```shell
$ make list-images

docker.io/nutanix/ntnx-csi-precheck:3.0.0
docker.io/nutanix/ntnx-csi:3.0.0
docker.io/rancher/local-path-provisioner:v0.0.28
ghcr.io/nutanix-cloud-native/caren-helm-reg:v0.0.0-dev
ghcr.io/nutanix-cloud-native/cloud-provider-nutanix/controller:v0.4.1
ghcr.io/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix:v0.0.0-dev
public.ecr.aws/ebs-csi-driver/aws-ebs-csi-driver:v1.34.0
public.ecr.aws/eks-distro/kubernetes-csi/external-attacher:v4.6.1-eks-1-30-10
public.ecr.aws/eks-distro/kubernetes-csi/external-provisioner:v5.0.1-eks-1-30-10
public.ecr.aws/eks-distro/kubernetes-csi/external-resizer:v1.11.1-eks-1-30-10
public.ecr.aws/eks-distro/kubernetes-csi/livenessprobe:v2.13.0-eks-1-30-10
public.ecr.aws/eks-distro/kubernetes-csi/node-driver-registrar:v2.11.0-eks-1-30-10
quay.io/calico/apiserver:v3.28.1
quay.io/calico/cni:v3.28.1
quay.io/calico/csi:v3.28.1
quay.io/calico/ctl:v3.28.1
quay.io/calico/kube-controllers:v3.28.1
quay.io/calico/node-driver-registrar:v3.28.1
quay.io/calico/node:v3.28.1
quay.io/calico/pod2daemon-flexvol:v3.28.1
quay.io/calico/typha:v3.28.1
quay.io/cilium/certgen:v0.2.0
quay.io/cilium/cilium-envoy@sha256:bd5ff8c66716080028f414ec1cb4f7dc66f40d2fb5a009fff187f4a9b90b566b
quay.io/cilium/cilium:v1.16.1
quay.io/cilium/operator-generic:v1.16.1
quay.io/frrouting/frr:9.1.0
quay.io/metallb/controller:v0.14.8
quay.io/metallb/speaker:v0.14.8
quay.io/tigera/operator:v1.34.3
registry.k8s.io/autoscaling/cluster-autoscaler:v1.30.0
registry.k8s.io/nfd/node-feature-discovery:v0.16.4
registry.k8s.io/provider-aws/cloud-controller-manager:v1.27.9
registry.k8s.io/provider-aws/cloud-controller-manager:v1.28.9
registry.k8s.io/provider-aws/cloud-controller-manager:v1.29.6
registry.k8s.io/provider-aws/cloud-controller-manager:v1.30.2
registry.k8s.io/provider-aws/cloud-controller-manager:v1.31.0
registry.k8s.io/sig-storage/csi-attacher:v4.4.3
registry.k8s.io/sig-storage/csi-external-health-monitor-controller:v0.10.0
registry.k8s.io/sig-storage/csi-node-driver-registrar:v2.9.3
registry.k8s.io/sig-storage/csi-provisioner:v3.6.3
registry.k8s.io/sig-storage/csi-resizer:v1.9.3
registry.k8s.io/sig-storage/csi-snapshotter:v3.0.3
registry.k8s.io/sig-storage/livenessprobe:v2.11.0
registry.k8s.io/sig-storage/snapshot-controller:v8.1.0
```

Depends on #822.

### DIFF
--- a/make/addons.mk
+++ b/make/addons.mk
@@ -87,8 +87,7 @@ template-helm-repository: generate-mindthegap-repofile ## this is used by goreal
 
 .PHONY: list-images
 list-images:
-	cd hack/tools/fetch-images && go build
-	./hack/tools/fetch-images/fetch-images \
-		-chart-directory=./charts/cluster-api-runtime-extensions-nutanix/ \
-		-helm-chart-configmap=./charts/cluster-api-runtime-extensions-nutanix/templates/helm-config.yaml \
-		-caren-version=$(CAREN_VERSION) >> caren-images.txt
+	cd hack/tools/fetch-images && go run . \
+	  -chart-directory=$(PWD)/charts/cluster-api-runtime-extensions-nutanix/ \
+	  -helm-chart-configmap=$(PWD)/charts/cluster-api-runtime-extensions-nutanix/templates/helm-config.yaml \
+	  -caren-version=$(CAREN_VERSION)
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -56,7 +56,7 @@ before:
     - make template-helm-repository
     - |
       sh -ec 'if [ {{ .IsSnapshot }} == false ] ; then
-        env CAREN_VERSION=v{{ trimprefix .Version "v" }} make list-images
+        make --no-print-directory CAREN_VERSION=v{{ trimprefix .Version "v" }} list-images >caren-images.txt
       fi'
 
 builds:
--- a/hack/tools/fetch-images/main.go
+++ b/hack/tools/fetch-images/main.go
@@ -9,7 +9,7 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"path"
+	"path/filepath"
 	"slices"
 	"strings"
 	"text/template"
@@ -20,6 +20,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/yaml"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 )
@@ -30,10 +31,11 @@ const (
 )
 
 type ChartInfo struct {
-	repo         string
-	name         string
-	valuesFile   string
-	stringValues []string
+	repo             string
+	name             string
+	valuesFile       string
+	stringValues     []string
+	extraImagesFiles []string
 }
 
 func main() {
@@ -47,9 +49,9 @@ func main() {
 	flagSet.StringVar(&chartDirectory, "chart-directory", "",
 		"path to chart directory for CAREN")
 	flagSet.StringVar(&helmChartConfigMap, "helm-chart-configmap", "",
-		"path to chart directory for CAREN")
+		"path to helm chart configmap for CAREN")
 	flagSet.StringVar(&carenVersion, "caren-version", "",
-		"caren version for images override")
+		"CAREN version for images override")
 	err := flagSet.Parse(args[1:])
 	if err != nil {
 		fmt.Println("failed to parse args", err.Error())
@@ -66,7 +68,7 @@ func main() {
 		os.Exit(1)
 	}
 	if chartDirectory == "" || helmChartConfigMap == "" {
-		fmt.Println("chart-directory helm-chart-configmap must be set")
+		fmt.Println("chart-directory and helm-chart-configmap must be set")
 		os.Exit(1)
 	}
 	i := &ChartInfo{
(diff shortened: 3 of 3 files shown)


dkoshkin (approved):

make/addons.mk
```diff
@@ -87,8 +87,7 @@ template-helm-repository: generate-mindthegap-repofile ## this is used by goreal
 
 .PHONY: list-images
 list-images:
```
faiq, line 89: Why not leave the redirection? 

Users are already able to run the tool as is to get the output

faiq (changes requested):
we dont need to print to stdout

make/addons.mk
```diff
@@ -87,8 +87,7 @@ template-helm-repository: generate-mindthegap-repofile ## this is used by goreal
...
-	cd hack/tools/fetch-images && go build
-	./hack/tools/fetch-images/fetch-images \
-		-chart-directory=./charts/cluster-api-runtime-extensions-nutanix/ \
-		-helm-chart-configmap=./charts/cluster-api-runtime-extensions-nutanix/templates/helm-config.yaml \
-		-caren-version=$(CAREN_VERSION) >> caren-images.txt
+	cd hack/tools/fetch-images && go run . \
```
supershal, line 90: This will also print log messages about downloading packages in clean environment. i.e. the release job in GHA may print logs similar to: 
```
❯ make list-images
go: downloading sigs.k8s.io/cluster-api v1.8.1
go: downloading k8s.io/cli-runtime v0.30.3
go: downloading k8s.io/client-go v0.30.3
go: downloading k8s.io/apiextensions-apiserver v0.30.3
go: downloading go4.org v0.0.0-20201209231011-d4a079459e60
go: downloading k8s.io/kubectl v0.30.3
go: downloading k8s.io/apiserver v0.30.3
go: downloading google.golang.org/grpc v1.62.2
go: downloading go.opentelemetry.io/otel v1.24.0
go: downloading go.opentelemetry.io/otel/trace v1.24.0
go: downloading go.opentelemetry.io/otel/metric v1.24.0
go: downloading k8s.io/component-base v0.30.3
docker.io/nutanix/ntnx-csi-precheck:3.0.0
```
should we redirect stderr to `/dev/null` when running `go run`

make/addons.mk
jimmidyson: Using the make recipe builds and calls the tool with the correct flags which is just a bit simpler for the user.

I found this much more convenient to use when I was testing the changes I made and it feels more natural for the make target to list to stdout and allow the caller to redirect where they want. I think I saw @dkoshkin recommend the same thing on the original PR, which triggered me changing it to this - I prefer this approach as well.

make/addons.mk
jimmidyson: I'll change back to `go build` and calling the built executable directly so as not to hide build issues.

jimmidyson: Actually even using `go build` will print out the same log messages but because they're going to stderr I think this is fine. In GHA that won't matter because it will only be used when releasing, in which case the stdout is redirected to a `caren-images.txt`.

supershal (approved):

make/addons.mk
dkoshkin: I made the original suggestion because it my mind it makes the make target more portable.
The caller can decide what to do with this output, write it out to file, sort it or filter it, etc.

Even when reviewing this PR, it made it slightly simpler calling the target and seeing the output in my terminal and not having to look up a file name.

faiq: I think it’s a helpful thing when reviewing a pr like this but we need to add more steps to the goreleaser file. 

If you want to list the images you can create another target or run the tool directly.

jimmidyson: Why do you feel we need more steps in the goreleaser config? It seems to work fine in my tests as is

jimmidyson: I'm missing something - the change here is the only change from your pr in goreleaser config which calls the same target. No extra logic to your pr, just very minor tweaks in how make is called. Adding another target is unnecessary with how it works in this pr and doesn't require any extra logic in goreleaser config from what I can see.

dkoshkin: @faiq I understand your concern with more logic in the goreleaser file but it already needs to know the name of the file [here](https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/900/files#diff-42e26dc67aed8aa3edb2472b4403288c1699fb6dc47419b9a475f0f224fe4689L34), shouldn't it also own writing to this file instead of having to keep track of it in 2 different files?

faiq: I just don't think this target needs to print to stdout by default when there isn't a use case for this that will normally come up. I've simplified this in my PR as well 

https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/822/commits/892c6e4bb57d2b11ef4da79fd7ca10b96afea28a

jimmidyson: Can we agree to disagree? A couple of us would like to go one way, one would like to go another. It feels like we're taking an unnecessary amount of time for a trivial change where there is no right or wrong answer, just a difference of opinion.

faiq: ok ill die on another hill

faiq (approved):

PR #899, opened 2024-09-03
feat: Enable Hubble Relay in Cilium deployment via CAAPH

dlipovetsky:
**What problem does this PR solve?**:
Relay is required for network visibility of a cluster as a whole. See https://docs.cilium.io/en/stable/gettingstarted/hubble_intro/.

**Which issue(s) this PR fixes**:
Fixes #

**How Has This Been Tested?**:

**Special notes for your reviewer**:

PR #901, opened 2024-09-05
fix: Do not use digests for Cilium images

jimmidyson:
This would break for air-gapped installations as the repo digest
changes when we build the air-gapped bundle.

Depends on #900.
Depends on #899.


dkoshkin (approved):
Looks good, changed from:
```
"quay.io/cilium/cilium-envoy:v1.29.7-39a2a56bbd5b3a591f69dbca51d3e30ef97e0e51@sha256:bd5ff8c66716080028f414ec1cb4f7dc66f40d2fb5a009fff187f4a9b90b566b"
```

```
"quay.io/cilium/cilium-envoy:v1.29.7-39a2a56bbd5b3a591f69dbca51d3e30ef97e0e51"
```

PR #903, opened 2024-09-09
build: Include Calico images in image list

jimmidyson:
Calico images are deployed by the tigera-operator and as such are not
discoverable by the helm list images plugin. Using the extra images file
functionality of the helm list images plugin allows for adding these
images as templated strings that will adapt with the chart version as it
is updated in future.

This commit also fixes up some other things:

- Use `filepath` instead of `path` for portabiity
- Output the image list to stdout from make to allow user to decide
  where to redirect list to
- Omit changing dir from image list when running as goreleaser hook
- Slightly more defensive when checking errors

Depends on #822.

### DIFF
--- a/hack/tools/fetch-images/main.go
+++ b/hack/tools/fetch-images/main.go
@@ -9,7 +9,7 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"path"
+	"path/filepath"
 	"slices"
 	"strings"
 	"text/template"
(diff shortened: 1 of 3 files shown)


mhrabovcin (approved):

hack/tools/fetch-images/main.go
```diff
@@ -195,21 +221,50 @@ func getHelmValues(carenChartDirectory string) (map[string]interface{}, error) {
...
+	case "tigera-operator":
+		f := filepath.Join(carenChartDirectory, "addons", "cni", "calico", "aws", defaultHelmAddonFilename)
+		tempFile, err := os.CreateTemp("", "")
+		if err != nil {
+			return "", fmt.Errorf("failed to create temp file: %w", err)
+		}
+
+		// CAAPH uses unstructured internally, so we need to create an unstructured copy of a cluster
+		// to pass to the CAAPH values template.
+		c, _ := runtime.DefaultUnstructuredConverter.ToUnstructured(&clusterv1.Cluster{})
```
thunderboltsid, line 245: we seem to be ignoring an error here

thunderboltsid (approved):

hack/tools/fetch-images/main.go
jimmidyson: Added error check.

thunderboltsid (approved):

mhrabovcin (approved):

PR #909, opened 2024-09-17
feat: Support node taints per nodepool and control plane

jimmidyson:

Fixes https://jira.nutanix.com/browse/NCN-102652


api/v1alpha1/nodeconfig_types.go
```diff
@@ -100,8 +106,54 @@ func (s NutanixNodeConfig) VariableSchema() clusterv1.VariableSchema { //nolint:
...
+	// +kubebuilder:validation:Optional
+	Value string `json:"value,omitempty"`
+
+	// The effect of the taint on pods that do not tolerate the taint.
+	// Valid effects are NoSchedule, PreferNoSchedule and NoExecute.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:default=NoSchedule
+	// +kubebuilder:validation:Enum:=NoSchedule;PreferNoSchedule;NoExecute
+	Effect TaintEffect `json:"effect"`
 }
```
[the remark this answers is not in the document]
jimmidyson: No need for `TimeAdded` field for one thing.

But more importantly need to add `kubebuilder` annotations to for enums etc to allow for OpenAPI validation of variable values.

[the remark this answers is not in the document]
manoj-nutanix: yeah, enums, etc. validation won't be achievable without explicit kubebuilder annotation. Thanks.

pkg/handlers/generic/mutation/taints/inject_controlplane.go
```diff
@@ -0,0 +1,107 @@
...
+			).Info("adding taints to worker node kubeadm config template")
+			if obj.Spec.Template.Spec.KubeadmConfigSpec.InitConfiguration != nil {
+				obj.Spec.Template.Spec.KubeadmConfigSpec.InitConfiguration = &bootstrapv1.InitConfiguration{}
+			}
+			if obj.Spec.Template.Spec.KubeadmConfigSpec.JoinConfiguration != nil {
+				obj.Spec.Template.Spec.KubeadmConfigSpec.JoinConfiguration = &bootstrapv1.JoinConfiguration{}
+			}
+			coreTaints := toCoreTaints(taintsVar)
+
+			obj.Spec.Template.Spec.KubeadmConfigSpec.InitConfiguration.NodeRegistration.Taints = append(
```
dkoshkin, line 96: >taints specifies the taints the Node API object should be registered with. If this field is unset, i.e. nil, it will be defaulted with a control-plane taint for control-plane nodes. 

From the docs https://kubernetes.io/docs/reference/config-api/kubeadm-config.v1beta4/#kubeadm-k8s-io-v1beta4-NodeRegistrationOptions 

If we do set a taint here, do the default control-plane taints still get added?
Also do we want to support the usecase of setting an empty taints and letting users remove the defaults?

dkoshkin (approved):
The changes look great, just a great a question on the default taints for control-planes.

api/v1alpha1/nodeconfig_types.go
```diff
@@ -100,8 +106,54 @@ func (s NutanixNodeConfig) VariableSchema() clusterv1.VariableSchema { //nolint:
...
+	// The taint value corresponding to the taint key.
+	// +kubebuilder:validation:Optional
+	Value string `json:"value,omitempty"`
+
+	// The effect of the taint on pods that do not tolerate the taint.
+	// Valid effects are NoSchedule, PreferNoSchedule and NoExecute.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:default=NoSchedule
+	// +kubebuilder:validation:Enum:=NoSchedule;PreferNoSchedule;NoExecute
+	Effect TaintEffect `json:"effect"`
```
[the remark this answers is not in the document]
jimmidyson: Answered in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/909#discussion_r1764698769

pkg/handlers/generic/mutation/taints/inject_controlplane.go
jimmidyson: You're suspicion is correct: default CP taints do not get added...

```
<SNIP>
labels:
      node-role.kubernetes.io/control-plane: ""
<SNIP>
spec:
    taints:
    - effect: NoSchedule
      key: some-key
      value: some-value
```

Is that what we want here? I suggest tainting the CP nodes is less common than for worker nodes so I think this is OK, and indeed doing what you suggest if the taints are set but empty slice would allow removing the default taints, allowing scheduling in constrained resource environments.

dkoshkin: >To support this would require a larger change that I would not be happy doing in this PR at least.

That makes sense, users can still do this by setting it to an empty slice in the KCP Template.

jimmidyson: > > To support this would require a larger change that I would not be happy doing in this PR at least.
> 
> That makes sense, users can still do this by setting it to an empty slice in the KCP Template.

Or by adding an inline JSON patch to the ClusterClass.

jimmidyson: Added docs.

jimmidyson: FIgured it out! We now support not setting default taints! Please take another look @dkoshkin, also added docs about it.

mhrabovcin (approved):

thunderboltsid (approved):

PR #912, opened 2024-09-17
ci(main): enable creating release-please PR from release branches

supershal:
- enable creating release-please PR from release branch
- set caren provider to use v0.15.x in CAREN e2e tests.

### DIFF
--- a/make/dev.mk
+++ b/make/dev.mk
@@ -45,7 +45,8 @@ dev.update-bootstrap-credentials-aws:
 
 .PHONY: release-please
 release-please:
-ifneq ($(shell [[ "$(GIT_CURRENT_BRANCH)" = "main" ]] || [[ "$(GIT_CURRENT_BRANCH)" == release/v* ]]; echo $$?),0)
+# filter Returns all whitespace-separated words in text that do match any of the pattern words.
+ifeq ($(filter main release/v%,$(GIT_CURRENT_BRANCH)),)
 	$(error "release-please should only be run on the main or release branch")
 else
 	release-please release-pr --repo-url $(GITHUB_ORG)/$(GITHUB_REPOSITORY) --target-branch $(GIT_CURRENT_BRANCH) --token "$$(gh auth token)"
--- a/make/dev.mk
+++ b/make/dev.mk
@@ -45,11 +45,10 @@ dev.update-bootstrap-credentials-aws:
 
 .PHONY: release-please
 release-please:
-ifneq ($(GIT_CURRENT_BRANCH),main)
-	$(error "release-please should only be run on the main branch")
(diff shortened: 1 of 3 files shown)


test/e2e/config/caren.yaml
```diff
@@ -134,8 +134,8 @@ providers:
 - name: caren
   type: RuntimeExtensionProvider
   versions:
-  - name: "{go://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix@v0.14}"
-    value: "https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/releases/download/{go://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix@v0.14}/runtime-extension-components.yaml"
+  - name: "{go://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix@v0.15}"
```
supershal, line 137: @jimmidyson can you please verify if the main branch should be using current minor release from `main` ?

jimmidyson: Until we get the upgrade test in place this is actually unused so I'm happy to merge this as it is and we can revisit this when we get the upgrade test in place. I think we might want to have both previous patch and previous minor in here and use both in upgrade tests but like I said let's discuss that when we come to it.

make/dev.mk
```diff
@@ -45,11 +45,10 @@ dev.update-bootstrap-credentials-aws:
 
 .PHONY: release-please
 release-please:
-ifneq ($(GIT_CURRENT_BRANCH),main)
-	$(error "release-please should only be run on the main branch")
+ifneq ($(shell [[ "$(GIT_CURRENT_BRANCH)" = "main" ]] || [[ "$(GIT_CURRENT_BRANCH)" == release/v* ]]; echo $$?),0)
```
jimmidyson, line 48: Instead of using `shell` here, let's use make's `filter` instead to not need to shell out:

```suggestion
ifeq ($(filter main release/v%,$(GIT_CURRENT_BRANCH)),)
```

jimmidyson (changes requested):
Please update release docs with details for running for release branch too.

test/e2e/config/caren.yaml
supershal: thank you for the details..

make/dev.mk
supershal: fixed https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/912/commits/15279b83dc0cd138df520e5df8fdeffca6563da7

jimmidyson (approved):

dkoshkin (approved):

dlipovetsky (approved):

PR #918, opened 2024-09-26
build: Enable building binary only on macos

jimmidyson:
Current config only allows building local binaries on Linux machines so
adding config to allow building binaries on macos without changing the
release config (and hence still only releasing linux binaries as required).


jimmidyson:
Yes it does always run in a container, but I found it useful to have the binary locally to check flags when I was adding feature gates. 

At the same time, if you run `make build-snapshot` on macos without this pr then literally nothing builds, not the binary, not the image, nada. This pr fixes that as well so a native darwin binary will be built plus the Linux container with relevant arch.

thunderboltsid (approved):

mhrabovcin (approved):

PR #921, opened 2024-09-26
fix: autogenerated file to satisfy linter

dkoshkin:
This caused a release error https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/actions/runs/11041694582/job/30713405814

Fixes #

I just ran `make dev.run-on-kind` to generate these changes.


supershal (approved):

thunderboltsid:
Do we have two separate linters in place for two separate workflows? I remember encountering this where pre-commit target would actually mutate this back.

jimmidyson:
> Do we have two separate linters in place for two separate workflows? I remember encountering this where pre-commit target would actually mutate this back.

Yeah I think that's right. I think we should exclude it from the golines call somehow as part of this PR so it doesn't get reverted.

jimmidyson:
I know it's not great but I fixed it up in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/922/files#diff-fe1727af26e0329c2bb94f9d274ed74fa926a0a7af1392c8db761788949a0bc3 I think so it will be ignored by golines for formatting so we shouldn't see this issue again once we merge #922.

jimmidyson:
Actually maybe better to pull those makefile changes out to this PR.

jimmidyson:
Realised that the change I proposed to fix this going forward is not going to work so I pushed #923 as an alternative that should fix this permanently.

PR #924, opened 2024-09-27
feat: Auto renewal of control plane certificates patch

jimmidyson:
This commit enables configuration of automatic certificate renewal via
the Kubeadm Control Plane provider as detailed in
https://cluster-api.sigs.k8s.io/tasks/certs/auto-rotate-certificates-in-kcp.html.

Although there is a reasonably sized refactoring (including renames) the
API types in the api module, the variable schema as opposed to CAPI
clusters via the external patches hook contains only the changes for
auto-renewal of certificates. I feel this is OK as the main clients of
the API are the Cluster authors rather than other projects using this
via code at this point at least.


api/v1alpha1/clusterconfig_types.go
```diff
@@ -151,7 +151,7 @@ type NutanixClusterConfigSpec struct {
 	Addons *NutanixAddons `json:"addons,omitempty"`
 
 	// +kubebuilder:validation:Optional
-	ControlPlane *NutanixNodeConfigSpec `json:"controlPlane,omitempty"`
+	ControlPlane *NutanixControlPlaneSpec `json:"controlPlane,omitempty"`
```
deepakm-ntnx, line 154: how do we handle existing cluster's upgrade to this schema? should we upgrade api version?

jimmidyson: As I explained in the pr description, this does not affect the cluster class variable schema so will only affect downstream projects using this in go code, of which I can only think of one potential which is konvoy. I do not think we need to bump an API version for this, even more so that it's still alpha versioned.

api/v1alpha1/controlplane_types.go
```diff
@@ -0,0 +1,49 @@
...
+	// certificates of the control plane.
+	// +kubebuilder:validation:Optional
+	AutoRenewCertificates *AutoRenewCertificatesSpec `json:"autoRenewCertificates,omitempty"`
+}
+
+type AutoRenewCertificatesSpec struct {
+	// DaysBeforeExpiry indicates a rollout needs to be performed if the
+	// certificates of the control plane will expire within the specified days.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Minimum=7
```
dkoshkin, line 17: ~Is thins minimum coming from CAPI?~ Yes https://cluster-api.sigs.k8s.io/tasks/certs/auto-rotate-certificates-in-kcp

docs/content/customization/generic/auto-renewal-of-control-plane-certs.md
```diff
@@ -0,0 +1,41 @@
...
+
+Applying this configuration will result in the following configuration being applied:
+
+- `KubeadmControlPlaneTemplate`:
+
+  - ```yaml
+    spec:
+      kubeadmConfigSpec:
+        rolloutBefore:
+          certificatesExpiryDays: 30
```
dkoshkin, line 38: I think its like this for the `KubeadmControlPlaneTemplate`? There is an example in https://cluster-api.sigs.k8s.io/tasks/certs/auto-rotate-certificates-in-kcp, but its for the `KubeadmControlPlane` and not the template
```suggestion
  - ```yaml
    spec:
      template:
        spec:
          rolloutBefore:
            certificatesExpiryDays: 30
```

dkoshkin (approved):
Thanks for keeping the external API the same.
Not a big deal to update this in Konvoy.

docs/content/customization/generic/auto-renewal-of-control-plane-certs.md
jimmidyson: Good catch, thanks.

prajnutanix (approved):

mhrabovcin (approved):

PR #930, opened 2024-10-01
fix: Shorten readiness probe period to try to prevent races

jimmidyson:
Default readiness probe period is 10s so this should reduce timeouts
caused by overzealous probes.

This will need to backported to 0.14 release branch.

### DIFF
--- a/charts/cluster-api-runtime-extensions-nutanix/templates/deployment.yaml
+++ b/charts/cluster-api-runtime-extensions-nutanix/templates/deployment.yaml
@@ -73,7 +73,7 @@ spec:
             fieldRef:
               fieldPath: metadata.namespace
         resources:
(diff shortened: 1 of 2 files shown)


thunderboltsid (approved):

dkoshkin (approved):

dkoshkin:
:( This is the wrong timeout on the helm registry

jimmidyson:
> :( This is the wrong timeout on the helm registry

What do you mean?

jimmidyson:
I pushed a different implementation that I prefer to #934 that introduces a startup probe to the CAREN pod instead of changing the readiness probe config so that it doesn't cause the kubelet to run readiness checks more often than necessary after a successful startup.

PR #939, opened 2024-10-08
feat(api): Add kubernetes version to coredns version mapping

thunderboltsid:
Added script and makefile target to update coredns version mapping.
`make coredns.sync` will fetch the list of releases from upstream
release branches and create a mapping of k8s versions to coredns versions.


jimmidyson:
Can you explain how this is used? The CoreDNS version mapping aren't referenced anywhere in this PR.

thunderboltsid:
> Can you explain how this is used? The CoreDNS version mapping aren't referenced anywhere in this PR.

Indeed, this PR just adds the mapping. We will need more changes to actually use that mapping (subsequent PRs). For cluster-class based clusters, we need a post cluster upgrade hook in CAREN that will set the CoreDNS version to the new the new version. For classic clusters, CAREN isn't in the picture, so we will need to also adopt that logic in the classic upgrade path.

jimmidyson:
Ah OK this is for the CoreDNS upgrade issue

api/versions/coredns.go
```diff
@@ -0,0 +1,60 @@
...
+	Kubernetes_V1_27: CoreDNS_V1_10_1,
+	Kubernetes_V1_28: CoreDNS_V1_10_1,
+	Kubernetes_V1_29: CoreDNS_V1_11_1,
+	Kubernetes_V1_30: CoreDNS_V1_11_3,
+	Kubernetes_V1_31: CoreDNS_V1_11_3,
+}
+
+// GetCoreDNSVersion returns the CoreDNS version for a given Kubernetes version.
+// If the Kubernetes version is not found, it returns an empty string and false.
+func GetCoreDNSVersion(kubernetesVersion string) (string, bool) {
```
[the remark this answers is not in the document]
thunderboltsid: That's a good idea, it'll also capture the patch and build info being different.

hack/tools/go.mod
```diff
@@ -7,6 +7,7 @@ go 1.22.5
 
 require (
 	github.com/d2iq-labs/helm-list-images v0.11.0
+	golang.org/x/mod v0.17.0
```
dkoshkin, line 10: nit: The rest of the code uses `"github.com/blang/semver/v4` would that package work here?

dkoshkin (approved):

hack/tools/go.mod
dlipovetsky: +1, if possible

thunderboltsid: I don't see this being pulled directly in the go.mod for tools: https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/939/files#diff-f389ca9ba468c8313b9d514e887e4b8899886bb2d6790352386aaf01da3aa588R8-R16

for the api, I chose golang.org/x/mod/semver to avoid adding more third-party dependencies for consumers as golang.org/x/mod/semver is technically still upstream go.

dlipovetsky:
Question: How do we ensure that the mapping stays up-to-date with the supported Kubernetes versions? I'm not sure that we define "supported Kubernetes versions" anywhere in the repo, actually. We also have a mapping https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/c644526d967b6c2778d34a7a75908a1042110d20/charts/cluster-api-runtime-extensions-nutanix/values.yaml#L74-L79 of versions for a different purpose.

I think we need a "source of truth" for supported Kubernetes versions, but we don't have to add it in this PR.

dlipovetsky (approved):

hack/tools/go.mod
thunderboltsid: okay; done. Although, I liked the APIs on the older package better.

dkoshkin (approved):

dkoshkin:
Thanks for updating the semver package.

thunderboltsid:
> Question: How do we ensure that the mapping stays up-to-date with the supported Kubernetes versions? I'm not sure that we define "supported Kubernetes versions" anywhere in the repo, actually. We also have a mapping
> 
> https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/c644526d967b6c2778d34a7a75908a1042110d20/charts/cluster-api-runtime-extensions-nutanix/values.yaml#L74-L79
> 
> of versions for a different purpose.
> I think we need a "source of truth" for supported Kubernetes versions, but we don't have to add it in this PR.

The script in this PR looks for every version greater than a minimum version (default v1.22: 10 versions old). The source of truth is basically lookups to k/k every time this script is run (pre-commit).

dlipovetsky (approved):

PR #941, opened 2024-10-11
fix: use minimal image for NFD

dkoshkin:
This images was reverted (by accident?) in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/commit/73b1648265a6bc9f26dccc76c19916c37a2fecf5#diff-b4f5af9dc43df286d19ee851401e0c36f5e5a1db0e902f8eec24718d93dfce72R1176

Setting the tag in the Helm values to use the minimal image.

Fixes #


jimmidyson:
The ninjas suffix is no longer required as it is just a different tag for backwards compatibility. The default image (without the monkey suffix) is the distroless minimal base image. Details at https://kubernetes-sigs.github.io/node-feature-discovery/v0.13/deployment/image-variants.html

dkoshkin:
Ahh great, thanks for the link!

PR #946, opened 2024-10-16
feat: adds previous versions to self hosted helm registry

faiq:
https://jira.nutanix.com/browse/NCN-103149

Fixes #

### DIFF
--- a/make/addons.mk
+++ b/make/addons.mk
@@ -30,7 +30,6 @@ export PREVIOUS_CAREN_CHARTS_VERSION := v0.14.6
 .PHONY: addons.sync
 addons.sync: $(addprefix update-addon.,calico cilium nfd cluster-autoscaler snapshot-controller local-path-provisioner-csi aws-ebs-csi kube-vip)
 addons.sync: $(addprefix update-addon.aws-ccm.,127 128 129 130 131)
-addons.sync: template-helm-repository
 
 .PHONY: update-addon.calico
 update-addon.calico: ; $(info $(M) updating calico manifests)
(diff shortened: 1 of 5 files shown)


make/addons.mk
```diff
@@ -25,6 +25,8 @@ export KUBE_VIP_VERSION := v0.8.3
 
 export METALLB_CHART_VERSION := 0.14.8
 
+export PREVIOUS_CAREN_CHARTS_VERSION := v0.14.6
```
dkoshkin, line 28: How do we decide the previous version? I think instead we would want all previous versions.

For example, assume we have these versions:
```
v0.17.0
v0.17.1
v0.18.0
v0.18.1
```

So:
- `v0.18.1` would contain all Charts from itself, `v0.18.0` and `v0.17.1`
- `v0.18.0` would contain all Charts from itself and `v0.17.1`
- `v0.17.1` would contain all Charts from itself and `v0.17.0`
- `v0.17.0` would contain all Charts from itself plus some manually generated list from all previous versions (down to some version we decide on)

Then once `v0.19.0` the process would just work because `v0.18.1` has all the Charts.

Periodically we would probably need to drop old Charts.

faiq: I think we should do this based on the versions that we ship, otherwise this container would get far too bloated with useless things.

faiq: I think the way to solve this is by passing two versions in as flags n-1 and n-2 this way we have all the relevant information we need.

jimmidyson: CAREN doesn't know what versions are shipped in any product so I don't know how we actually restrict the set of images here.

I think I would prefer the registry image to be built downstream using the image lists from whatever releases are shipped by that downstream product.

faiq: i'll see if creating it somewhere else isn't too cumbersome and hacky

PR #947, opened 2024-10-21
feat: copy charts to pvc

faiq:
Copies charts to a common pvc. 

Still missing:
- Special container that holds previous versions that would populate the initial PVC for previous versions. 

Fixes #

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -108,8 +108,7 @@ dockers:
       - hack/addons/mindthegap-helm-registry/repos.yaml
     build_flag_templates:
       - "--platform=linux/amd64"
-      - "--pull"
-      - "--builder=caren"
+      - '--build-arg=VERSION=v{{ trimprefix .Version "v" }}'
       - "--label=org.opencontainers.image.created={{.CommitDate}}"
       - "--label=org.opencontainers.image.title=caren-helm-reg"
       - "--label=org.opencontainers.image.revision={{.FullCommit}}"
@@ -124,8 +123,7 @@ dockers:
       - hack/addons/mindthegap-helm-registry/repos.yaml
     build_flag_templates:
       - "--platform=linux/arm64"
-      - "--pull"
-      - "--builder=caren"
+      - '--build-arg=VERSION=v{{ trimprefix .Version "v" }}'
       - "--label=org.opencontainers.image.created={{.CommitDate}}"
       - "--label=org.opencontainers.image.title=caren-helm-reg"
       - "--label=org.opencontainers.image.revision={{.FullCommit}}"
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -108,7 +108,7 @@ dockers:
       - hack/addons/mindthegap-helm-registry/repos.yaml
     build_flag_templates:
       - "--platform=linux/amd64"
-      - '--build-arg=version=v{{ trimprefix .Version "v" }}'
+      - '--build-arg=VERSION=v{{ trimprefix .Version "v" }}'
       - "--pull"
       - "--label=org.opencontainers.image.created={{.CommitDate}}"
       - "--label=org.opencontainers.image.title=caren-helm-reg"
@@ -125,7 +125,7 @@ dockers:
     build_flag_templates:
       - "--platform=linux/arm64"
       - "--pull"
-      - '--build-arg=version=v{{ trimprefix .Version "v" }}'
+      - '--build-arg=VERSION=v{{ trimprefix .Version "v" }}'
       - "--label=org.opencontainers.image.created={{.CommitDate}}"
       - "--label=org.opencontainers.image.title=caren-helm-reg"
       - "--label=org.opencontainers.image.revision={{.FullCommit}}"
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -44,7 +44,7 @@ before:
       $(helm template {{ .ProjectName }} ./charts/{{ .ProjectName }} \
         --namespace caren-system \
         --set-string image.tag=v{{ trimprefix .Version "v" }}{{ if .IsSnapshot }}-{{ .Runtime.Goarch }}{{ end }} \
-        --set-string helmChartsImage.tag=v{{ trimprefix .Version "v" }}{{ if .IsSnapshot }}-{{ .Runtime.Goarch }} \
+        --set-string helmRepositoryImage.tag=v{{ trimprefix .Version "v" }}{{ if .IsSnapshot }}-{{ .Runtime.Goarch }} \
         --set-string image.repository=ko.local/{{ .ProjectName }}{{ end }} \
       )
       EOF'
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -108,8 +108,8 @@ dockers:
       - hack/addons/mindthegap-helm-registry/repos.yaml
     build_flag_templates:
       - "--platform=linux/amd64"
-      - '--build-arg=VERSION=v{{ trimprefix .Version "v" }}'
       - "--pull"
+      - "--builder=caren"
       - "--label=org.opencontainers.image.created={{.CommitDate}}"
       - "--label=org.opencontainers.image.title=caren-helm-reg"
       - "--label=org.opencontainers.image.revision={{.FullCommit}}"
@@ -125,7 +125,7 @@ dockers:
     build_flag_templates:
       - "--platform=linux/arm64"
       - "--pull"
-      - '--build-arg=VERSION=v{{ trimprefix .Version "v" }}'
+      - "--builder=caren"
       - "--label=org.opencontainers.image.created={{.CommitDate}}"
       - "--label=org.opencontainers.image.title=caren-helm-reg"
       - "--label=org.opencontainers.image.revision={{.FullCommit}}"
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -108,6 +108,7 @@ dockers:
       - hack/addons/mindthegap-helm-registry/repos.yaml
     build_flag_templates:
       - "--platform=linux/amd64"
+      - '--version=v{{ trimprefix .Version "v" }}'
       - "--pull"
       - "--label=org.opencontainers.image.created={{.CommitDate}}"
       - "--label=org.opencontainers.image.title=caren-helm-reg"
@@ -124,6 +125,7 @@ dockers:
     build_flag_templates:
       - "--platform=linux/arm64"
       - "--pull"
+      - '--version=v{{ trimprefix .Version "v" }}'
       - "--label=org.opencontainers.image.created={{.CommitDate}}"
       - "--label=org.opencontainers.image.title=caren-helm-reg"
       - "--label=org.opencontainers.image.revision={{.FullCommit}}"
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -108,7 +108,7 @@ dockers:
       - hack/addons/mindthegap-helm-registry/repos.yaml
     build_flag_templates:
       - "--platform=linux/amd64"
-      - '--version=v{{ trimprefix .Version "v" }}'
+      - '--build-arg=version=v{{ trimprefix .Version "v" }}'
       - "--pull"
       - "--label=org.opencontainers.image.created={{.CommitDate}}"
       - "--label=org.opencontainers.image.title=caren-helm-reg"
@@ -125,7 +125,7 @@ dockers:
     build_flag_templates:
       - "--platform=linux/arm64"
       - "--pull"
-      - '--version=v{{ trimprefix .Version "v" }}'
+      - '--build-arg=version=v{{ trimprefix .Version "v" }}'
       - "--label=org.opencontainers.image.created={{.CommitDate}}"
       - "--label=org.opencontainers.image.title=caren-helm-reg"
       - "--label=org.opencontainers.image.revision={{.FullCommit}}"
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -108,6 +108,7 @@ dockers:
       - hack/addons/mindthegap-helm-registry/repos.yaml
     build_flag_templates:
       - "--platform=linux/amd64"
+      - "--pull"
       - '--build-arg=VERSION=v{{ trimprefix .Version "v" }}'
       - "--label=org.opencontainers.image.created={{.CommitDate}}"
       - "--label=org.opencontainers.image.title=caren-helm-reg"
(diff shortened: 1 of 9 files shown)


charts/cluster-api-runtime-extensions-nutanix/templates/helm-repository.yaml
```diff
@@ -26,6 +26,21 @@ spec:
     kind: Issuer
   secretName: helm-repository-tls
 ---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  finalizers:
+  - kubernetes.io/pvc-protection
+  name: charts-pvc
```
supershal, line 34:
```suggestion
  name: addon-charts
```

charts/cluster-api-runtime-extensions-nutanix/templates/helm-repository.yaml
```diff
@@ -45,18 +60,34 @@ spec:
       labels:
         app: helm-repository  # pods have this label, and Service and Deployment depend on it
```
supershal, line 61: nit: if we are renaming to `helmChartsImage`, lets change labels and file name of the template

charts/cluster-api-runtime-extensions-nutanix/templates/helm-repository.yaml
```diff
@@ -45,18 +60,34 @@ spec:
...
     spec:
+      initContainers:
+      - name: init-container
```
supershal, line 64:
```suggestion
      - name: copy-charts

File: charts/cluster-api-runtime-extensions-nutanix/templates/helm-repository.yaml (hunk 1 above)
Comment (faiq) [reply charts/cluster-api-runtime-extensions-nutanix/templates/helm-repository.yaml:34]: i think the original name is fine?

File: charts/cluster-api-runtime-extensions-nutanix/templates/helm-repository.yaml (hunk 3 above)
Comment (faiq) [reply charts/cluster-api-runtime-extensions-nutanix/templates/helm-repository.yaml:64]: sure

File: charts/cluster-api-runtime-extensions-nutanix/templates/helm-repository.yaml [hunk 4]
```diff
@@ -45,18 +60,34 @@ spec:
...
+        image: "{{ .Values.helmChartsImage.repository }}:{{ default $.Chart.AppVersion .Values.helmChartsImage.tag }}"
+        command: ["/bin/sh", "-c", "cp /charts/*.tar /helm-charts"]
+        imagePullPolicy: "{{ .Values.image.pullPolicy }}"
```
supershal, line 67: `volumeMount` for the initContainer is missing

.goreleaser.yml
```diff
@@ -108,6 +108,7 @@ dockers:
       - hack/addons/mindthegap-helm-registry/repos.yaml
     build_flag_templates:
       - "--platform=linux/amd64"
+      - '--build-arg=version=v{{ trimprefix .Version "v" }}'
```
supershal, line 111: Isnt it case sensetive?
```suggestion
      - '--build-arg=VERSION=v{{ trimprefix .Version "v" }}'
```

faiq: this seems to compile just fine for me

charts/cluster-api-runtime-extensions-nutanix/templates/helm-repository.yaml
supershal: sounds good. nice to have.

charts/cluster-api-runtime-extensions-nutanix/templates/helm-repository.yaml
faiq: nice catch here

charts/cluster-api-runtime-extensions-nutanix/templates/helm-repository.yaml
```diff
@@ -26,6 +26,21 @@ spec:
     kind: Issuer
   secretName: helm-repository-tls
 ---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  finalizers:
+  - kubernetes.io/pvc-protection
+  name: charts-pvc
+  namespace: default
```
supershal, line 35:
```suggestion
  namespace: {{ .Release.Namespace }}
```

charts/cluster-api-runtime-extensions-nutanix/templates/helm-repository.yaml
```diff
@@ -26,6 +26,21 @@ spec:
     kind: Issuer
   secretName: helm-repository-tls
 ---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  finalizers:
+  - kubernetes.io/pvc-protection
```
supershal, line 33: The clusterctl upgrade process deletes provider components and redeploys them. But since we are not providing label to the PVC, it should not delete it. https://github.com/kubernetes-sigs/cluster-api/blob/main/cmd/clusterctl/client/cluster/components.go#L143

It would be good to test out upgrade capi-components process.

Since default retainPolicy for our `default` storage class is `delete` it will delete the PV if PVC is deleted. we have to ensure this PVC is not deleted automatically by any of our processes.

faiq: yes, im taking a look at that now.

faiq:
looking into the e2e failures

hack/addons/mindthegap-helm-registry/Dockerfile
```diff
@@ -1,16 +1,9 @@
...
-VOLUME /certs
-CMD [ "serve", "bundle", \
-      "--bundle=/tmp/helm-charts.tar", \
-      "--listen-port=5000", \
-      "--listen-address=0.0.0.0", \
-      "--tls-private-key-file=/certs/tls.key", \
-      "--tls-cert-file=/certs/tls.crt" ]
+FROM --platform=${TARGETPLATFORM} alpine
+COPY --from=bundle_builder /tmp/helm-charts.tar /charts/
+RUN mv /charts/helm-charts.tar /charts/helm-charts-$(date +%s).tar
```
faiq, line 9: I can't figure out a way to make this unique without using date. I tried passing in .ShortCommit from goreleaser as well as version, but from my understanding the run isn't expanding it for whatever reason.

dkoshkin: I think this works?
```
diff --git a/.goreleaser.yml b/.goreleaser.yml
index be51152e..f38cf4c7 100644
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -131,6 +131,7 @@ dockers:
       - "--label=org.opencontainers.image.revision={{.FullCommit}}"
       - '--label=org.opencontainers.image.version=v{{ trimprefix .Version "v" }}'
       - "--label=org.opencontainers.image.source={{.GitURL}}"
+      - '--build-arg=CHARTS_VERSION=v{{ trimprefix .Version "v" }}'
     goarch: arm64
 
 docker_manifests:
diff --git a/hack/addons/mindthegap-helm-registry/Dockerfile b/hack/addons/mindthegap-helm-registry/Dockerfile
index e07b858e..11697c90 100644
--- a/hack/addons/mindthegap-helm-registry/Dockerfile
+++ b/hack/addons/mindthegap-helm-registry/Dockerfile
@@ -6,4 +6,5 @@ RUN --mount=source=./hack/addons/mindthegap-helm-registry/repos.yaml,target=/rep
 
 FROM --platform=${TARGETPLATFORM} alpine
 COPY --from=bundle_builder /tmp/helm-charts.tar /charts/
-RUN mv /charts/helm-charts.tar /charts/helm-charts-$(date +%s).tar
+ARG CHARTS_VERSION
+RUN mv /charts/helm-charts.tar /charts/helm-charts-${CHARTS_VERSION}.tar

```

And it got passed in correctly
```
➜  cluster-api-runtime-extensions-nutanix git:(faiq/copy-charts-to-pvc) ✗ docker run --entrypoint ls -it ghcr.io/nutanix-cloud-native/caren-helm-reg:v0.19.0-dev-arm64 /charts
helm-charts-v0.19.0-dev.tar
```

faiq: yeah moved it down

charts/cluster-api-runtime-extensions-nutanix/templates/helm-repository.yaml
```diff
@@ -45,18 +60,37 @@ spec:
...
       containers:
       - name: helm-repository
         ports:
           - name: serve
             protocol: TCP
             containerPort: 5000
-        image: "{{ .Values.helmRepositoryImage.repository }}:{{ default $.Chart.AppVersion .Values.helmRepositoryImage.tag }}"
-        imagePullPolicy: "{{ .Values.image.pullPolicy }}"
+        image: "{{ .Values.mindthegap.repository }}:{{ .Values.mindthegap.tag }}"
+        imagePullPolicy: "{{ .Values.mindthegap.pullPolicy }}"
```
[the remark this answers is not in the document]
faiq: we are not shipping duplicate images. one is an alpine image and the other is mindthegap

charts/cluster-api-runtime-extensions-nutanix/templates/helm-repository.yaml
```diff
@@ -26,6 +26,21 @@ spec:
     kind: Issuer
   secretName: helm-repository-tls
 ---
+apiVersion: v1
```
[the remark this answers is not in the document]
faiq: sounds reasonable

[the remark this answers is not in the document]
supershal: what cases are you think where we need to disable it dynamically?
Disabling it would need to change the deployment too. it might be good to remove it when it time to disable it?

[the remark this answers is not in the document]
faiq: all of this is already guarded behind 

{{ if .Values.selfHostedRegistry }}

charts/cluster-api-runtime-extensions-nutanix/templates/helm-repository.yaml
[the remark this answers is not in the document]
supershal: +1. we could use the same mindthegap image for init-container too. We are not using versioned `alpine` image either. it would be less configuration to just use single image.

hack/addons/mindthegap-helm-registry/Dockerfile
```diff
@@ -1,16 +1,9 @@
...
-FROM --platform=${TARGETPLATFORM} ghcr.io/mesosphere/mindthegap:${MINDTHEGAP_VERSION}
-COPY --from=bundle_builder /tmp/helm-charts.tar /tmp/helm-charts.tar
-VOLUME /certs
-CMD [ "serve", "bundle", \
-      "--bundle=/tmp/helm-charts.tar", \
-      "--listen-port=5000", \
-      "--listen-address=0.0.0.0", \
-      "--tls-private-key-file=/certs/tls.key", \
-      "--tls-cert-file=/certs/tls.crt" ]
+FROM --platform=${TARGETPLATFORM} alpine
```
faiq, line 7: Add a version here JIC of CVEs or anything like that

charts/cluster-api-runtime-extensions-nutanix/README.md
```diff
@@ -32,9 +32,9 @@ A Helm chart for cluster-api-runtime-extensions-nutanix
 | deployment.replicas | int | `1` |  |
 | env | object | `{}` |  |
 | helmAddonsConfigMap | string | `"default-helm-addons-config"` |  |
-| helmRepositoryImage.pullPolicy | string | `"IfNotPresent"` |  |
-| helmRepositoryImage.repository | string | `"ghcr.io/nutanix-cloud-native/caren-helm-reg"` |  |
-| helmRepositoryImage.tag | string | `""` |  |
+| helmChartsImage.pullPolicy | string | `"IfNotPresent"` |  |
+| helmChartsImage.repository | string | `"ghcr.io/nutanix-cloud-native/caren-helm-reg"` |  |
+| helmChartsImage.tag | string | `""` |  |
```
dkoshkin, line 37: Should we keep the current values to avoid a breaking change, the image name is not changing and is now doing both the serving and in the init container?

dkoshkin (commented):
Looking good, just some minor nits!

charts/cluster-api-runtime-extensions-nutanix/templates/helm-repository.yaml
```diff
@@ -45,18 +60,38 @@ spec:
...
+          - --bundle=/helm-charts/helm-charts-*.tar
+          - --listen-port=5000
+          - --listen-address=0.0.0.0
+          - --tls-private-key-file=/certs/tls.key
+          - --tls-cert-file=/certs/tls.crt
         volumeMounts:
         - name: certs-vol
           mountPath: "/certs"
           readOnly: true
+        - name: charts-volume
```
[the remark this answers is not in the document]
faiq: i don't think this is really necessary?

charts/cluster-api-runtime-extensions-nutanix/README.md
faiq: Since this is deployed using clusterctl and the runtime-extensions config file, we don't have to worry about this.

dkoshkin: Thats true for Konvoy but this is an open source project and there are already [docs](https://confluence.eng.nutanix.com:8443/display/KAR/Create+CAPX+Cluster+using+CAREN+and+Install+Kommander) that have instructions to use it as a Helm chart.
I understand the reasoning why you changed it in the first place when the image only contained the Helm charts, but now thats also has mindthegap and is used as a repository going back to `helmRepositoryImage` is a little more accurate IMO, but has the benefit of not changing the public interface, when not needed.

.goreleaser.yml
```diff
@@ -108,7 +108,7 @@ dockers:
       - hack/addons/mindthegap-helm-registry/repos.yaml
     build_flag_templates:
       - "--platform=linux/amd64"
-      - "--pull"
```
dkoshkin, line 111: Was this done to fix an issue or just speed up the build?
(Want to make sure we have some history of why this changed)

faiq: good catch - this was removed to do something else.

charts/cluster-api-runtime-extensions-nutanix/README.md
faiq: deploying via helm chart is for development purposes and the value already has sane defaults. if someone was deploying this in their own developer workflow they are well equipped to read the docs and change the values as needed.

charts/cluster-api-runtime-extensions-nutanix/templates/helm-repository.yaml
faiq: the labels are automatically applied it would be up to the client to remove them. we did this downstream by implementing a componentsClient t

dkoshkin (approved):
We got there with a real clean solution 🎉

PR #954, opened 2024-10-28
fix: bundle correct arch for the mindthegap binary

dkoshkin:
Latest changes broke the published arm build. The `mindthegap` binary from `bundle_builder` will always be `BUILDPLATFORM`, but we want it to match `TARGETPLATFORM`.

Fixes #

### DIFF
--- a/hack/addons/mindthegap-helm-registry/Dockerfile
+++ b/hack/addons/mindthegap-helm-registry/Dockerfile
@@ -3,10 +3,15 @@ FROM --platform=${BUILDPLATFORM} ghcr.io/mesosphere/mindthegap:${MINDTHEGAP_VERS
 # this gets called by goreleaser so the copy source has to be the path relative to the repo root.
 RUN --mount=source=./hack/addons/mindthegap-helm-registry/repos.yaml,target=/repos.yaml \
     ["/ko-app/mindthegap", "create", "bundle", "--helm-charts-file=/repos.yaml", "--output-file=/tmp/helm-charts.tar"]
+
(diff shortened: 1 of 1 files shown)


supershal (approved):

hack/addons/mindthegap-helm-registry/Dockerfile
```diff
@@ -3,10 +3,15 @@ FROM --platform=${BUILDPLATFORM} ghcr.io/mesosphere/mindthegap:${MINDTHEGAP_VERS
 # this gets called by goreleaser so the copy source has to be the path relative to the repo root.
 RUN --mount=source=./hack/addons/mindthegap-helm-registry/repos.yaml,target=/repos.yaml \
     ["/ko-app/mindthegap", "create", "bundle", "--helm-charts-file=/repos.yaml", "--output-file=/tmp/helm-charts.tar"]
+
+FROM --platform=${TARGETPLATFORM} ghcr.io/mesosphere/mindthegap:${MINDTHEGAP_VERSION} as mindthegap
```
faiq, line 7: Can you recycle the from line on 2 using TARGTPLATFORM

faiq (changes requested):

hack/addons/mindthegap-helm-registry/Dockerfile
supershal: mindthegap from `BUILDPLATFORM` will be used to run `mindthegap create bundle` at runtime. So I think it should be `BUILDPLATFORM` in line 2.

faiq (approved):

PR #955, opened 2024-10-29
fix: use correct securityContext for Alpine based helm-repository Pod

dkoshkin:
User `65532` is the default UID in distroless based images, but in Alpine it's a different user and group. 

Fixes #

### DIFF
--- a/charts/cluster-api-runtime-extensions-nutanix/values.yaml
+++ b/charts/cluster-api-runtime-extensions-nutanix/values.yaml
@@ -156,8 +156,16 @@ resources:
   requests:
     cpu: 100m
     memory: 128Mi
+
 securityContext:
   runAsUser: 65532
+
+# The helm-repository containers are based on an Alpine image with a different nonroot user
+helmRepositorySecurityContext:
+  runAsUser: 405
+  runAsGroup: 100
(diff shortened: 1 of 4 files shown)


charts/cluster-api-runtime-extensions-nutanix/values.yaml
```diff
@@ -156,8 +156,16 @@ resources:
   requests:
     cpu: 100m
     memory: 128Mi
+
 securityContext:
   runAsUser: 65532
+
+# The helm-repository containers are based on an Alpine image with a different nonroot user
+helmRepositorySecurityContext:
```
dkoshkin, line 164: Based this name on `helmRepositoryImage`.

dlipovetsky (approved):

dkoshkin:
Maybe we actually want the `nobody` user?

charts/cluster-api-runtime-extensions-nutanix/values.yaml
```diff
@@ -156,8 +156,16 @@ resources:
...
+  runAsUser: 65534
```
supershal, line 165: Thank you for testing with this user.

supershal (approved):

PR #959, opened 2024-10-30
feat: set default CoreDNS version

dkoshkin:
Automatically upgrades the CoreDNS version. This is done by always setting `dns.imageTag` in KCP in the CoreDNS handler, based on the mapping to the cluster's Kubernetes version.

This component is different from etcd and kube-proxy that is also installed by kubeadm for a few different reasons.
- an etcd upgrade is handled by kubeadm 
- kube-proxy, another "addon", is [upgraded by CAPI core](https://github.com/kubernetes-sigs/cluster-api/blob/75c986db9e38190a2313eaf6e5f97d955fa96b65/controlplane/kubeadm/internal/controllers/controller.go#L476-L480) because its version will match the Kubernetes version

[This functiona call](https://github.com/kubernetes-sigs/cluster-api/blob/75c986db9e38190a2313eaf6e5f97d955fa96b65/controlplane/kubeadm/internal/controllers/controller.go#L482-L485) is misleading and [will only update the version if its set in KCP](https://github.com/kubernetes-sigs/cluster-api/blob/6d7104deacddf540c82734ae8abaf309f2ab3b90/controlplane/kubeadm/internal/workload_cluster_coredns.go#L223-L226).

To not cause a rollout of all managed clusters by changing the defaults, this PR introduces a new API to opt in. To enable this functionality a client can set this new API like so for new clusters and during cluster upgrades:
```
spec:
  topology:
    variables:
      - name: clusterConfig
        value:
          dns:
            coreDNS: {}
```

Fixes #


dkoshkin:
🤔 this would actually cause a rollout of new CP machines for all workload clusters after CAREN is updated. Maybe we should do this in a mutating webhook and set the version on the Cluster object?

pkg/handlers/generic/mutation/coredns/inject.go
```diff
@@ -65,11 +71,10 @@ func (h *coreDNSPatchHandler) Mutate(
...
 	)
 	if err != nil {
-		if variables.IsNotFoundError(err) {
-			log.V(5).Info("coreDNSVar variable not defined")
-			return nil
+		if !variables.IsNotFoundError(err) {
+			return err
 		}
-		return err
+		log.V(5).Info("coreDNSVar variable not defined")
```
jimmidyson, line 77: nit
```suggestion
		log.V(5).Info("coreDNS variable not defined")
```

jimmidyson (approved):

dkoshkin:
@jimmidyson I would like your help on the CEL validation, can you point me to any examples in the repo, I couldn't find any.

dkoshkin:
@jimmidyson I was thinking about this some more. Do we even need/want a new API type to define the upgrade strategy? (In that case I would revert the last commit and change the handler to return if dnsVar is empty)
What if having `dns: {}` enables the auto update logic. This is similar to the addon lifecycle hooks that don't have any "enabled" API, the struct not being nil enables it.

docs/content/customization/generic/dns.md
```diff
@@ -48,3 +48,32 @@ Applying this configuration will result in the following value being set:
...
+
+- `KubeadmControlPlaneTemplate`:
+
+  - ```yaml
+    spec:
+      kubeadmConfigSpec:
+        clusterConfiguration:
+          dns:
+            imageTag: "v1.11.3"
+    ```
```
jimmidyson, line 79: As this should be the default and preferred behaviour, could you move this above the explicit version setting please.

dkoshkin: Done in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/959/commits/b47e1b792e9d0ec5173b8c25e4e47646b89733bb

jimmidyson (approved):
Great work!

thunderboltsid (approved):

docs/content/customization/generic/dns.md
```diff
@@ -15,6 +15,35 @@ If the `dns.coreDNS` property is not specified, then the customization will be s
...
+kind: Cluster
+metadata:
+  name: <NAME>
+spec:
+  topology:
+    variables:
+      - name: clusterConfig
+        value:
+          dns:
+            coreDNS: {}
```
supershal, line 31: we should provide both image and tag in the example to illustrate how it will transform to `KubeadmControlPlaneTemplate`

dkoshkin: @supershal I think that example is already defined further down on this page?

supershal: I see now.

PR #968, opened 2024-11-07
build: set helmRepository tag in list-images target

dkoshkin:
**What problem does this PR solve?**:
Noticed in the [last release](https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/releases/download/v0.22.0/caren-images.txt) this image tag was set to `v0.0.0-dev`, fixing it to get the release version.

```
$ make list-images
...
ghcr.io/nutanix-cloud-native/caren-helm-reg:v0.22.1
```

**Which issue(s) this PR fixes**:
Fixes #

**How Has This Been Tested?**:

**Special notes for your reviewer**:

PR #969, opened 2024-11-07
build: rename caren-helm-reg to better match role

dkoshkin:
While looking at https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/968 I noticed the image name.
The `caren-helm-reg` image is no longer used as a registry. The word registry was also recently changed to repository in the templates.

Renamed it to `cluster-api-runtime-extensions-bundle-initializer`, but open to suggestions!

Fixes #


supershal:
wdyt `cluster-api-runtime-extensions-helm-chart-bundle-initializer`?

dlipovetsky:
No upgrade concerns, right?

dkoshkin:
> No upgrade concerns, right?

Not that I can think of, we've been changing these templates around in the last few releases and haven't seen any issues.

dlipovetsky (approved):

PR #979, opened 2024-11-26
build: Add CAREN release to cluster class artifacts

jimmidyson:
This attaches the updated cluster class artifacts containing the
CAREN release version to the release artifacts on release.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -30,7 +30,7 @@ release:
     - glob: release-metadata.yaml
       name_template: metadata.yaml
     - glob: runtime-extensions-components.yaml
-    - glob: ./charts/{{ .ProjectName }}/defaultclusterclasses/*.yaml
+    - glob: ./hack/examples/release/*-cluster-class.yaml
     - glob: caren-images.txt
 
 before:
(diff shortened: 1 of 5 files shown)


.goreleaser.yml
```diff
@@ -58,6 +58,10 @@ before:
       sh -ec 'if [ {{ .IsSnapshot }} == false ] ; then
         make --no-print-directory CAREN_VERSION=v{{ trimprefix .Version "v" }} list-images >caren-images.txt
       fi'
+    - |
+      sh -ec 'if [ {{ .IsSnapshot }} == false ] ; then
+        make --no-print-directory CAREN_VERSION=v{{ trimprefix .Version "v" }} add-version-to-clusterclasses
```
dkoshkin, line 63: Just curious, does `.IsSnapshot==false` when running e2e tests?

dkoshkin (approved):
Thanks!

.goreleaser.yml
jimmidyson: No, that runs with `release-snapshot` so this will not be run. Do you think there's a problem with that?

dkoshkin: We may need this in the future when running upgrade e2e tests, but probably premature to worry about it now.

dlipovetsky: I'm confused by your reply, because as I read it, it says "this code does not run when we use `release-snapshot`", but then what is the purpose of the conditional?

jimmidyson: `the conditional evaluates to false, so the branch does not run` is correct.

dlipovetsky (approved):

PR #986, opened 2024-12-02
feat: Virtual IP configuration to set different address/port

dkoshkin:
New configuration to support setting different virutalIP address from the external controlPlaneEndpoint. This is useful for Clusters like a Nutanix Cluster in a VPC where the virtual IP needs to be some IP from same Subnet as the VMs, but the conrtol-plane endpoint is a floating IP mapped to the virtual IP accessible from outside the VPC.
```
spec:
  topology:
    variables:
      - name: clusterConfig
        value:
          nutanix:
            controlPlaneEndpoint:
              host: x.x.x.x
              port: 6443
              virtualIP:
                configuration:
                  address: y.y.y.y
```
The "configuration" keyword comes from an existing API here https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/8e2b30770c64d6976a7990274988d53f9626f169/api/v1alpha1/addon_types.go#L260

Fixes #

* New unit tests
* Created a Nutanix Cluster inside a VPC

Assigned a floating IP `10.22.198.34` to an internal IP `172.16.0.101`
image
One of the control-plane VMs got the internal VIP assigned.
image

And the Cluster is accessible outside of the VPC.
```
kubectl get nodes
NAME                                          STATUS   ROLES           AGE    VERSION
dkoshkin-floating-ip-k9d74-d8jxp              Ready    control-plane   122m   v1.30.5
dkoshkin-floating-ip-k9d74-fb5pb              Ready    control-plane   125m   v1.30.5
dkoshkin-floating-ip-k9d74-wtl5w              Ready    control-plane   123m   v1.30.5
dkoshkin-floating-ip-md-0-qkg7q-pm5z9-6zf8f   Ready    <none>          124m   v1.30.5
```


charts/cluster-api-runtime-extensions-nutanix/templates/virtual-ip/kube-vip/manifests/kube-vip-configmap.yaml
```diff
@@ -46,7 +46,7 @@ data:
             - name: vip_retryperiod
               value: "2"
             - name: address
-              value: '{{ `{{ .ControlPlaneEndpoint.Host }}` }}'
+              value: '{{ `{{ .Address }}` }}'
```
supershal, line 49: curious if and how this address is added to the SANs in the certificates referred by the kubeconfig

api/v1alpha1/common_types.go
```diff
@@ -47,6 +47,24 @@ type ControlPlaneVirtualIPSpec struct {
 	// +kubebuilder:validation:Enum=KubeVIP
 	// +kubebuilder:default=KubeVIP
 	Provider string `json:"provider,omitempty"`
+
+	// Configuration for the chosen control-plane virtual IP provider.
+	// +kubebuilder:validation:Optional
+	Configuration *ControlPlaneVirtualIPConfiguration `json:"configuration,omitempty"`
```
supershal, line 53: can we directly use `Address` and `Port` here?  what is motivation behind adding `Configuration` wrapper?

charts/cluster-api-runtime-extensions-nutanix/templates/virtual-ip/kube-vip/manifests/kube-vip-configmap.yaml
dkoshkin: Great question, at first I thought that would be necessary too but it wasn't.
It's not needed because this IP is not used directly by any clients when reaching the API server, everything goes through the controlPlaneEndpoint.

api/v1alpha1/common_types.go
dkoshkin: Mainly for consistency with existing APIs https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/8e2b30770c64d6976a7990274988d53f9626f169/api/v1alpha1/addon_types.go#L260 and https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/8e2b30770c64d6976a7990274988d53f9626f169/api/v1alpha1/addon_types.go#L196.

supershal (approved):
LGTM

api/v1alpha1/common_types.go
```diff
@@ -47,6 +47,24 @@ type ControlPlaneVirtualIPSpec struct {
...
+	// The virtual IP on which the API server is serving.
+	// If left empty, the value from controlPlaneEndpoint.host will be used.
+	// +kubebuilder:validation:Optional
+	Address string `json:"address"`
+
+	// The port on which the API server is serving.
+	// If left empty, the value from controlPlaneEndpoint.port will be used.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=65535
```
supershal, line 66:
```suggestion
	// +kubebuilder:validation:Maximum=65535
	// +kubebuilder:validation:Format=ipv4
```

api/v1alpha1/common_types.go
```diff
@@ -47,6 +47,24 @@ type ControlPlaneVirtualIPSpec struct {
...
+
+	// Configuration for the chosen control-plane virtual IP provider.
+	// +kubebuilder:validation:Optional
+	Configuration *ControlPlaneVirtualIPConfiguration `json:"configuration,omitempty"`
+}
+
+type ControlPlaneVirtualIPConfiguration struct {
+	// The virtual IP on which the API server is serving.
+	// If left empty, the value from controlPlaneEndpoint.host will be used.
+	// +kubebuilder:validation:Optional
```
supershal, line 59:
```suggestion
	// +kubebuilder:validation:Optional
	// +kubebuilder:validation:Format=ipv4
```

jimmidyson: Could also be ipv6...

api/v1alpha1/common_types.go
jimmidyson: I don't think this is a necessary level of indirection in this case. The examples given are there to support generic configuration for LB and StorageClasses, but in this case this case it's a provider specific configuration for Nutanix clusters.

jimmidyson:
@dkoshkin Do we know of other providers that would need to support this? If not, then I wonder if we should make it specific to Nutanix clusters somehow.

docs/content/customization/nutanix/control-plane-endpoint.md
```diff
@@ -2,7 +2,7 @@
 title = "Control Plane Endpoint"
 +++
 
-Configure Control Plane Endpoint. Defines the host IP and port of the CAPX Kubernetes cluster.
+Configure Control Plane Endpoint. Defines the host IP and port of the Nutanix Kubernetes cluster.
```
jimmidyson, line 5: This is not specific to Nutanix clusters is it?

api/v1alpha1/common_types.go
jimmidyson: Let's add ipv4 format as @supershal suggests for now, and once we upgrade to CAPI 1.9 we can start to use `oneOf` schemas to support both ipv4 and ipv6 here.

api/v1alpha1/common_types.go
```diff
@@ -47,6 +47,24 @@ type ControlPlaneVirtualIPSpec struct {
...
+	Address string `json:"address"`
+
+	// The port on which the API server is serving.
+	// If left empty, the value from controlPlaneEndpoint.port will be used.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=65535
+	Port int32 `json:"port"`
```
jimmidyson, line 67:
```suggestion
	Port int32 `json:"port,omitempty"`
```

api/v1alpha1/common_types.go
```diff
@@ -47,6 +47,24 @@ type ControlPlaneVirtualIPSpec struct {
...
+	// Configuration for the chosen control-plane virtual IP provider.
+	// +kubebuilder:validation:Optional
+	Configuration *ControlPlaneVirtualIPConfiguration `json:"configuration,omitempty"`
+}
+
+type ControlPlaneVirtualIPConfiguration struct {
+	// The virtual IP on which the API server is serving.
+	// If left empty, the value from controlPlaneEndpoint.host will be used.
+	// +kubebuilder:validation:Optional
+	Address string `json:"address"`
```
jimmidyson, line 60:
```suggestion
	Address string `json:"address,omitempty"`
```

pkg/handlers/generic/mutation/controlplanevirtualip/providers/providers.go
```diff
@@ -35,11 +36,28 @@ func templateValues(
...
+		port = cmp.Or(
+			controlPlaneEndpoint.VirtualIPSpec.Configuration.Port,
+			controlPlaneEndpoint.Port,
+		)
+	}
 	templateInput := input{
-		ControlPlaneEndpoint: controlPlaneEndpoint,
+		Address: address,
+		Port:    port,
 	}
```
jimmidyson, line 61: Feels a bit odd to use cmp when we've already defaulted above. How about something like this?

```suggestion
	var virtualIPConfig ControlPlaneVirtualIPConfiguration
	if controlPlaneEndpoint.VirtualIPSpec != nil &&
		controlPlaneEndpoint.VirtualIPSpec.Configuration != nil {
		virtualIPConfig = *controlPlaneEndpoint.VirtualIPSpec.Configuration
	}
	templateInput := input{
		Address: cmp.Or(
			virtualIPConfig.Address,
			controlPlaneEndpoint.Host,
		),
		Port:    cmp.Or(
			virtualIPConfig.Port,
			controlPlaneEndpoint.Port,
		),
	}
```

jimmidyson (changes requested):
Actually just realised that Nutanix CCM handler will need updating in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/main/pkg/handlers/generic/lifecycle/ccm/nutanix/handler.go to pass in this VIP as well to be ignored as part of possible node IPs by Nutanix CCM.

pkg/handlers/generic/lifecycle/ccm/nutanix/handler.go
```diff
@@ -174,3 +183,17 @@ func templateValuesFunc(
 		return b.String(), nil
 	}
 }
+
+func ipsToIgnore(nutanixConfig *v1alpha1.NutanixSpec) []string {
+	toIgnore := []string{nutanixConfig.ControlPlaneEndpoint.Host}
+	// Also ignore the virtual IP if it is set.
```
dkoshkin, line 189: Good catch @jimmidyson.
I added the new IP in addition to controlPlane IP but if we want we could also do just a single IP.
LMK what you think.

cc @supershal

jimmidyson: This looks good to me. Better to be safe than sorry IMO and add all IPs just in case.

jimmidyson (approved):

PR #988, opened 2024-12-05
refactor: new waiter functionality in helmAddonApplier

dkoshkin:
This PR is a small refactor to pull existing functionality into the common addon applier. I wanted similar functionality in another handler to wait for the HCP to be ready. 

Fixes #

### DIFF
--- a/pkg/handlers/generic/lifecycle/addons/helmaddon.go
+++ b/pkg/handlers/generic/lifecycle/addons/helmaddon.go
@@ -6,11 +6,13 @@ package addons
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/go-logr/logr"
 	"github.com/spf13/pflag"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/cluster-api/util/conditions"
 	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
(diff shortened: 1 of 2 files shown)


pkg/handlers/generic/lifecycle/addons/helmaddon.go
```diff
@@ -227,7 +227,10 @@ func waitToBeReady(
 			Reader: client,
 			Target: hcp.DeepCopy(),
 			Check: func(_ context.Context, obj *caaphv1.HelmChartProxy) (bool, error) {
-				return conditions.IsTrue(obj, caaphv1.HelmReleaseProxiesReadyCondition), nil
+				if obj.ObjectMeta.Generation != obj.Status.ObservedGeneration {
```
dkoshkin, line 230: I noticed this when moving the function. I think its a good change to prevent the function from returning early, especially during cluster upgrades where the HCP already has HelmReleaseReadyCondition but its out of date.

pkg/handlers/generic/lifecycle/addons/helmaddon.go
```diff
@@ -194,5 +210,32 @@ func (a *helmAddonApplier) Apply(
...
+			Reader: client,
+			Target: hcp.DeepCopy(),
+			Check: func(_ context.Context, obj *caaphv1.HelmChartProxy) (bool, error) {
+				return conditions.IsTrue(obj, caaphv1.HelmReleaseProxiesReadyCondition), nil
+			},
+			Interval: 5 * time.Second,
+			Timeout:  30 * time.Second,
+		},
+	); err != nil {
+		return fmt.Errorf("failed to wait for MetalLB to deploy: %w", err)
```
jimmidyson, line 237:
```suggestion
		return fmt.Errorf("failed to wait for addon %s to deploy: %w", client.ObjectKeyFromObject(hcp), err)
```

jimmidyson (changes requested):

jimmidyson (approved):

PR #996, opened 2024-12-11
fix: validates NUTANIX_ENDPOINT does not fall in the Load Balancer IP Range

manoj-nutanix:
webhook errors out if NUTANIX_ENDPOINT IP falls in Load Balancer IP Range. It only implements dumb check which compares PC IP with Load Balancer IP Range. It's complex to achieve with CEL so going with webhook as we'll need to use regex(`isIP()` func is not working for cluster variables) to extract IP from PC URL and do string compare which not robust as we can do error handling through webhook.

Fixes # 
https://jira.nutanix.com/browse/NCN-102628

```
clusterctl generate cluster ${CLUSTER_NAME} \        
  --from ${CLUSTER_FILE} \
  --kubernetes-version ${KUBERNETES_VERSION} \
  --worker-machine-count 1 | \
kubectl apply --server-side -f -
secret/nutanix-cluster-cilium-helm-addonexport-dockerhub-credentials serverside-applied
secret/nutanix-cluster-cilium-helm-addonexport-pc-creds-for-csi serverside-applied
Warning: Detected changes to resource nutanix-cluster-cilium-helm-addonexport-pc-creds which is currently being deleted.
secret/nutanix-cluster-cilium-helm-addonexport-pc-creds serverside-applied
Error from server (Forbidden): admission webhook "cluster-validator.caren.nutanix.com" denied the request: prism central IP "198.18.1.1" must not be part of MetalLB address range "198.18.1.1"-"198.18.1.10"
```

### DIFF
--- a/pkg/helpers/helpers.go
+++ b/pkg/helpers/helpers.go
@@ -1,0 +1,43 @@
+// Copyright 2024 Nutanix. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+package helpers
+
+import (
+	"fmt"
+	"math/big"
+	"net"
+)
+
+// IsIPInRange checks if the target IP falls within the start and end IP range (inclusive).
+func IsIPInRange(startIP, endIP, targetIP string) (bool, error) {
+	// Parse the IPs
+	start := net.ParseIP(startIP)
+	end := net.ParseIP(endIP)
+	target := net.ParseIP(targetIP)
+
+	// Ensure all IPs are valid
+	if start == nil {
+		return false, fmt.Errorf("invalid start IP: %q", startIP)
+	}
+	if end == nil {
+		return false, fmt.Errorf("invalid end IP: %q", endIP)
+	}
+	if target == nil {
+		return false, fmt.Errorf("invalid target IP: %q", targetIP)
+	}
+
+	// Convert IPs to big integers
+	startInt := ipToBigInt(start)
+	endInt := ipToBigInt(end)
+	targetInt := ipToBigInt(target)
+
+	// Check if target IP is within the range
+	return targetInt.Cmp(startInt) >= 0 && targetInt.Cmp(endInt) <= 0, nil
+}
+
+// ipToBigInt converts a net.IP to a big.Int for comparison.
+func ipToBigInt(ip net.IP) *big.Int {
+	// Normalize to 16-byte representation for both IPv4 and IPv6
+	ip = ip.To16()
+	return big.NewInt(0).SetBytes(ip)
+}
--- a/pkg/helpers/helpers.go
+++ b/pkg/helpers/helpers.go
@@ -4,40 +4,27 @@ package helpers
 
 import (
 	"fmt"
-	"math/big"
-	"net"
+	"net/netip"
 )
 
 // IsIPInRange checks if the target IP falls within the start and end IP range (inclusive).
 func IsIPInRange(startIP, endIP, targetIP string) (bool, error) {
-	// Parse the IPs
-	start := net.ParseIP(startIP)
-	end := net.ParseIP(endIP)
-	target := net.ParseIP(targetIP)
-
-	// Ensure all IPs are valid
-	if start == nil {
-		return false, fmt.Errorf("invalid start IP: %q", startIP)
+	start, err := netip.ParseAddr(startIP)
+	if err != nil {
+		return false, fmt.Errorf("invalid start IP: %w", err)
 	}
-	if end == nil {
-		return false, fmt.Errorf("invalid end IP: %q", endIP)
+	end, err := netip.ParseAddr(endIP)
+	if err != nil {
+		return false, fmt.Errorf("invalid end IP: %w", err)
 	}
-	if target == nil {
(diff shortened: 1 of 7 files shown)


pkg/helpers/helpers.go
```diff
@@ -0,0 +1,43 @@
...
+	// Check if target IP is within the range
+	return targetInt.Cmp(startInt) >= 0 && targetInt.Cmp(endInt) <= 0, nil
+}
+
+// ipToBigInt converts a net.IP to a big.Int for comparison.
+func ipToBigInt(ip net.IP) *big.Int {
+	// Normalize to 16-byte representation for both IPv4 and IPv6
+	ip = ip.To16()
+	return big.NewInt(0).SetBytes(ip)
+}
```
jimmidyson, line 43: Let's use [`net/netip`](https://pkg.go.dev/net/netip) and [`go4.org/netipx`](https://pkg.go.dev/go4.org/netipx) instead:
```suggestion
import (
	"fmt"
	"net/netip"
	
	"go4.org/netipx"
)

// IsIPInRange checks if the target IP falls within the start and end IP range (inclusive).
func IsIPInRange(startIP, endIP, targetIP string) (bool, error) {
	// Parse the IPs
	start, err := netip.ParseAddr(startIP)
	if err != nil {
		return false, fmt.Errorf("invalid start IP %q: %w", start, err)
	}
	end, err := netip.ParseAddr(endIP)
	if err != nil {
		return false, fmt.Errorf("invalid end IP %q: %w", end, err)
	}
	target, err := netip.ParseAddr(targetIP)
	if err != nil {
		return false, fmt.Errorf("invalid target IP %q: %w", target, err)
	}
	
	ipRange := netips.IPRangeFrom(start, end)

	// Check if target IP is within the range
	return ipRange.Contains(target), nil
}
```

pkg/webhook/cluster/nutanix_validator.go
```diff
@@ -0,0 +1,127 @@
...
+	pcHostname, _, err := pcEndpoint.ParseURL()
+	if err != nil {
+		return err
+	}
+
+	pcIP := net.ParseIP(pcHostname)
+	// PC URL can contain IP/FQDN, so compare only if PC is an IP address.
+	if pcIP == nil {
+		return nil
+	}
```
jimmidyson, line 107: Prefer `netip.ParseAddr`

```suggestion
	pcIP, err := netip.ParseAddr(pcHostname)
	// PC URL can contain IP/FQDN, so compare only if PC is an IP address.
	if err != nil {
		return nil
	}
```

jimmidyson (changes requested):
Recently noticed that `net/netip` has improved functions for parsing IPs (errors instead of nil slices if not an IP!) and `netipx` has really useful functions for handling ranges.

pkg/webhook/cluster/nutanix_validator.go
manoj-nutanix: `netip.ParseAddr` would return error if `pcHostname` is FQDN and we'll need to ignore that error as we want avoid dns resolution and compare only if it's IP so `net.ParseIP` suits better for our usecase here(`net.ParseIP` internally uses `netip.ParseAddr` only). wdyt?

jimmidyson: Same same. Personally I prefer the semantics of returning an error and ignoring it (as i suggested) rather than checking if it's a nil value after parsing, but both work I guess.

pkg/helpers/helpers.go
manoj-nutanix: Thanks. I vaguely knew there was a `.Compare`, `.Less` methods, but I couldn't find it, so thought I might be mistaken.
Nit: Did not feel importing ext. pkg `"go4.org/netipx"` was adding any value so used only `"net/netip"`.

pkg/webhook/cluster/nutanix_validator.go
manoj-nutanix: I feel the same but `net.ParseIP` was doing exactly what we wanted than manually doing it with `netip.ParseAddr` so making exception here only.

pkg/helpers/helpers.go
```diff
@@ -0,0 +1,30 @@
...
+	target, err := netip.ParseAddr(targetIP)
+	if err != nil {
+		return false, fmt.Errorf("invalid target IP: %w", err)
+	}
+
+	if start.Compare(target) <= 0 && end.Compare(target) >= 0 {
+		return true, nil
+	}
+
+	return false, nil
```
jimmidyson, line 29: IMO netipx's `iprange.Contains(...)` provides much more readable semantics than this but that's just a preference of mine I guess.

Regardless, a nit but this can be simplified to:
```suggestion
	return start.Compare(target) <= 0 && end.Compare(target) >= 0, nil
```

jimmidyson (approved):

manoj-nutanix:
Closing this PR in favor of https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1001  as forked repos don't have access to github secrets so some e2e tests failing.

PR #1001, opened 2024-12-12
fix: validates PC IP is outside Load Balancer IP Range

manoj-nutanix:
**What problem does this PR solve?**:

webhook errors out if NUTANIX_ENDPOINT IP falls in Load Balancer IP Range. It only implements dumb check which compares PC IP with Load Balancer IP Range. It's complex to achieve with CEL so going with webhook as we'll need to use regex(isIP() func is not working for cluster variables) to extract IP from PC URL and do string compare which not robust as we can do error handling through webhook.

**Which issue(s) this PR fixes**:
Fixes #
https://jira.nutanix.com/browse/NCN-102628

**How Has This Been Tested?**:

**Special notes for your reviewer**:

```
✗ clusterctl generate cluster ${CLUSTER_NAME} \  --from ${CLUSTER_FILE} \
  --kubernetes-version ${KUBERNETES_VERSION} \
  --worker-machine-count 1 | \
  kubectl apply --server-side -f -
secret/nutanix-cluster-cilium-helm-addon-dockerhub-credentials serverside-applied
secret/nutanix-cluster-cilium-helm-addon-pc-creds-for-csi serverside-applied
Warning: Detected changes to resource nutanix-cluster-cilium-helm-addon-pc-creds which is currently being deleted.
secret/nutanix-cluster-cilium-helm-addon-pc-creds serverside-applied
Error from server (Forbidden): admission webhook "cluster-validator.caren.nutanix.com" denied the request: prism central IP "198.18.1.1" must not be part of MetalLB address range "198.18.1.1"-"198.18.1.10"
```

PR #1002, opened 2024-12-13
fix: validates CONTROL_PLANE_ENDPOINT_IP and NUTANIX_ENDPOINT are distinct

manoj-nutanix:
webhook errors out if NUTANIX_ENDPOINT IP is same as PC IP. It only implements dumb check which compares PC IP with control-plane IP.

Fixes #
 https://jira.nutanix.com/browse/NCN-102626

```
 clusterctl generate cluster ${CLUSTER_NAME} \   
  --from ${CLUSTER_FILE} \
  --kubernetes-version ${KUBERNETES_VERSION} \
  --worker-machine-count 1 | \
  kubectl apply --server-side -f -
secret/nutanix-cluster-cilium-helm-addon-dockerhub-credentials serverside-applied
secret/nutanix-cluster-cilium-helm-addon-pc-creds-for-csi serverside-applied
Warning: Detected changes to resource nutanix-cluster-cilium-helm-addon-pc-creds which is currently being deleted.
secret/nutanix-cluster-cilium-helm-addon-pc-creds serverside-applied
Error from server (Forbidden): admission webhook "cluster-validator.caren.nutanix.com" denied the request: prism central and control plane endpoint cannot have the same IP "10.47.10.4"
```


pkg/webhook/cluster/nutanix_validator.go
```diff
@@ -125,3 +134,35 @@ func checkIfPrismCentralIPInLoadBalancerIPRange(
...
+func checkIfPrismCentralAndControlPlaneIPSame(
+	pcEndpoint v1alpha1.NutanixPrismCentralEndpointSpec,
+	controlPlaneEndpointSpec v1alpha1.ControlPlaneEndpointSpec,
+) error {
+	controlPlaneEndpointIP, err := netip.ParseAddr(
+		controlPlaneEndpointSpec.ControlPlaneEndpointIP(),
+	)
+	if err != nil {
+		// controlPlaneEndpointIP is strictly accepted as an IP address from user so
+		// if it is not an IP address, it is invalid.
```
dkoshkin, line 151: I don't think this statement is accurate.
The API does not enforce this because using a VIP is not a requirement. You could use an external LB with DNS.
https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/8c18f88f9d6accdd126106cdf0ae47f546e35c84/api/v1alpha1/common_types.go#L28-L31

dkoshkin: With https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/8c18f88f9d6accdd126106cdf0ae47f546e35c84/api/v1alpha1/common_types.go#L28-L31 we'll also want to check that overlap too now.
That will always be an ipv4.

manoj-nutanix: Thanks for the pointers. I was mistaken that we forgot to put API restriction to allow only ipv4.. would be good to update konvoy CLI [flag](https://github.com/mesosphere/konvoy2/blob/e512f9d5d7c9ab3cf2ec7dd944c9f7df9c8a632b/cmd/konvoy/app/flag/nutanix.go#L317) to string type.

Also, [LB range](https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/8c18f88f9d6accdd126106cdf0ae47f546e35c84/api/v1alpha1/addon_types.go#L272-L278) allows ipv4 only, event though [metalb](https://github.com/metallb/metallb/blob/eeb320f15b70603ec6b93fe8ec3c290384b8a2a0/api/v1beta1/ipaddresspool_types.go#L25-L29) can accpet ipv6, CIDR, etc.

manoj-nutanix: > With https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/986 we'll also want to check that overlap too now.
That will always be an ipv4.

Did not understand what kind of overlap? I'm comparing with IP gotten from [this](https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1002/files#diff-43f16565973dd5eec1b4539f863395d67ab6b52237f0ee811b439eb1512237bbR81-R91) method, did you mean to compare both [host](https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/8c18f88f9d6accdd126106cdf0ae47f546e35c84/api/v1alpha1/common_types.go#L28-L31) and VIP with PC IP?

dkoshkin: RE the ipv4 vs ipv6 we made the decision in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/986#discussion_r1870996440

manoj-nutanix: thanks, done in b02e13ed99c30a834760ebc89de841aa7524bc58 and 7b50e2aba7b3cafee7eaa971ec52c173e73e7d7d

pkg/webhook/cluster/nutanix_validator.go
```diff
@@ -125,3 +134,33 @@ func checkIfPrismCentralIPInLoadBalancerIPRange(
 
 	return nil
 }
+
+// checkIfPrismCentralAndControlPlaneIPSame checks if Prism Central and Control Plane IP are same.
+// It compares strictly IP addresses(no FQDN) and doesn't involve any network calls.
+// This is a temporary check until we have a better way to handle this by reserving IPs
+// using IPAM provider.
+func checkIfPrismCentralAndControlPlaneIPSame(
```
dkoshkin, line 142:
```suggestion
// validatePrismCentralDoesNotEqualControlPlaneIP checks if Prism Central and Control Plane IP are not equal.
// It compares strictly IP addresses(no FQDN) and doesn't involve any network calls.
// This is a temporary check until we have a better way to handle this by reserving IPs
// using IPAM provider.
func validatePrismCentralDoesNotEqualControlPlaneIP(
```

pkg/webhook/cluster/nutanix_validator.go
```diff
@@ -125,3 +134,33 @@ func checkIfPrismCentralIPInLoadBalancerIPRange(
 
 	return nil
 }
+
+// checkIfPrismCentralAndControlPlaneIPSame checks if Prism Central and Control Plane IP are same.
+// It compares strictly IP addresses(no FQDN) and doesn't involve any network calls.
+// This is a temporary check until we have a better way to handle this by reserving IPs
+// using IPAM provider.
```
dkoshkin, line 141: I would suggest we remove this and not have include predictions.
```suggestion
```

pkg/webhook/cluster/nutanix_validator.go
```diff
@@ -125,3 +134,33 @@ func checkIfPrismCentralIPInLoadBalancerIPRange(
...
+func checkIfPrismCentralAndControlPlaneIPSame(
+	pcEndpoint v1alpha1.NutanixPrismCentralEndpointSpec,
+	controlPlaneEndpointSpec v1alpha1.ControlPlaneEndpointSpec,
+) error {
+	controlPlaneEndpointIP, err := netip.ParseAddr(controlPlaneEndpointSpec.VirtualIPAddress())
```
dkoshkin, line 146: nit:
```suggestion
	virtualIPAddress, err := netip.ParseAddr(controlPlaneEndpointSpec.VirtualIPAddress())
```

pkg/webhook/cluster/nutanix_validator.go
```diff
@@ -125,3 +134,33 @@ func checkIfPrismCentralIPInLoadBalancerIPRange(
...
+		// so return directly.
+		return nil
+	}
+
+	pcHostname, _, err := pcEndpoint.ParseURL()
+	if err != nil {
+		return err
+	}
+
+	pcIP, err := netip.ParseAddr(pcHostname)
```
dkoshkin, line 158: This looks very similar to https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1001/files#diff-cbaf86208d1803beb69b004fc24fe8d3dccbc2cba74f10dde9f8e8acf5edebeeR98-R107 but is using slightly different logic, can we reuse some util function please?

dkoshkin (commented):
Looking good, just a few comments

api/v1alpha1/common_types_test.go
```diff
@@ -0,0 +1,67 @@
...
+// SPDX-License-Identifier: Apache-2.0
+package v1alpha1
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestControlPlaneEndpointIP(t *testing.T) {
```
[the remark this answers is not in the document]
manoj-nutanix: 21293e9ffa7fd596f3ba894845b1c7f3f74f177c

pkg/webhook/cluster/nutanix_validator.go
manoj-nutanix: 21293e9ffa7fd596f3ba894845b1c7f3f74f177c

pkg/webhook/cluster/nutanix_validator.go
manoj-nutanix: 21293e9ffa7fd596f3ba894845b1c7f3f74f177c

pkg/webhook/cluster/nutanix_validator.go
manoj-nutanix: 21293e9ffa7fd596f3ba894845b1c7f3f74f177c

pkg/webhook/cluster/nutanix_validator.go
manoj-nutanix: 21293e9ffa7fd596f3ba894845b1c7f3f74f177c

api/v1alpha1/nutanix_clusterconfig_types.go
```diff
@@ -76,3 +77,17 @@ func (s NutanixPrismCentralEndpointSpec) ParseURL() (string, uint16, error) {
 
 	return hostname, uint16(port), nil
 }
+
+func (s NutanixPrismCentralEndpointSpec) ParseIP() (netip.Addr, error) {
+	pcHostname, _, err := s.ParseURL()
+	if err != nil {
+		return netip.Addr{}, err
```
dkoshkin, line 84: nit: returning an empty struct here instead of `nil` kinda smells

dkoshkin (approved):
Thanks!

api/v1alpha1/nutanix_clusterconfig_types.go
manoj-nutanix: yeah, looks weird at first.. I just followed `net/netip` pkg semantic, maybe `Addr` struct doesn't hold much data so they return empty struct instead of nil everywhere.

prajnutanix (approved):

PR #1011, opened 2025-01-16
feat: adds new field for helm values input for cilium CNI

manoj-nutanix:
Introduces new field in CAREN API to add custom helm values input for cilium CNI.

Fixes #
https://jira.nutanix.com/browse/NCN-105147


api/v1alpha1/addon_types.go
```diff
@@ -115,6 +115,11 @@ type CNI struct {
 	// +kubebuilder:default=HelmAddon
 	// +kubebuilder:validation:Enum=ClusterResourceSet;HelmAddon
 	Strategy *AddonStrategy `json:"strategy,omitempty"`
+
+	// ValuesRef is an object reference to Configmap/Secret which contains inline YAML
```
manoj-nutanix, line 119:
```suggestion
	// ValuesRef is an object reference to Configmap which contains inline YAML
```
we could do the above to reduce the scope further.

jimmidyson: Perhaps not applicable for CNI configuration but secrets should be generally referenceable for values for any more confidential data.

api/v1alpha1/addon_types.go
```diff
@@ -115,6 +115,11 @@ type CNI struct {
...
+	// representing the values for the Helm chart when HelmAddon is the strategy.
+	// +kubebuilder:validation:Optional
+	ValuesRef *corev1.ObjectReference `json:"valuesRef,omitempty"`
```
jimmidyson, line 122: Can we use `TypedLocalObjectReference` instead?

api/v1alpha1/addon_types.go
```diff
@@ -115,6 +115,12 @@ type CNI struct {
 	// +kubebuilder:default=HelmAddon
 	// +kubebuilder:validation:Enum=ClusterResourceSet;HelmAddon
 	Strategy *AddonStrategy `json:"strategy,omitempty"`
+
+	// ValuesRef is an object reference to Configmap/Secret inside the same namespace
+	// which contains inline YAML representing the values for the Helm chart
+	// when HelmAddon is the strategy.
+	// +kubebuilder:validation:Optional
+	ValuesRef *corev1.TypedLocalObjectReference `json:"valuesRef,omitempty"`
```
dkoshkin, line 123: @jimmidyson @manoj-nutanix wdyt about wrapping this in another struct to make it extensible in future.
Maybe something like:
```
values:
  ref:
```

In the future we could do things like
```
values:
  ref:
  strategy: merge/override
  inline: |...
```

supershal:
can you please update the doc file. https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/main/docs/content/addons/cni.md

api/v1alpha1/addon_types.go
```diff
@@ -115,6 +115,12 @@ type CNI struct {
 	// +kubebuilder:default=HelmAddon
 	// +kubebuilder:validation:Enum=ClusterResourceSet;HelmAddon
 	Strategy *AddonStrategy `json:"strategy,omitempty"`
+
+	// ValuesRef is an object reference to Configmap/Secret inside the same namespace
+	// which contains inline YAML representing the values for the Helm chart
+	// when HelmAddon is the strategy.
```
supershal, line 121: I think CRS strategy is going to be deprecated eventually. Should we assume that `valuesRef` always used for helm strategy and mentioning `the values for the Cilium Helm chart` in the field description is good enough.
```suggestion
	// ValuesRef is an object reference to Configmap/Secret inside the same namespace
	// which contains inline YAML representing the values for the Cilium Helm chart
```

api/v1alpha1/addon_types.go
manoj-nutanix: sure, will make this change.. would be nice to have it extensible. Only thing is, it would be learning curve for customers to understand meaning of all this.  I'm not sure whether customers even know meaning of strategy ClusterResourceSet vs HelmAddon and its implications.

api/v1alpha1/addon_types.go
manoj-nutanix: don't think it's much of concern right now.. Once we deprecate CRS from [AddonStrategy](https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1011/files#diff-6fd54619366c9d6ade856bbf25da675fea8f7d9384b0a947ea742b637d66cc7bR116), we could always update the `ValuesRef` description here. Are your concerned that we'll forget this later?

api/v1alpha1/addon_types.go
supershal: agree that we should use secrets for more confidential data in the values.
CAREN reads and parse the values from configmaps and set it as inline YAML in the HCP object. If we allow secret to be reference here, we have to decode user's secret and set it in the HCP inline YAML. Is it okay for CAREN to do this?

api/v1alpha1/addon_types.go
supershal: Thanks for making the change. my only concern was that description refers to other field in the schema but we do not strictly validate it in the code. So we can relax the description a bit.
OR
We can also validate in the webhook if ValuesRef is used with wrong strategy.

api/v1alpha1/addon_types.go
jimmidyson: Looks good to me!

api/v1alpha1/addon_types.go
manoj-nutanix: Intention is to inform customer when `ValuesRef` field will be considered, API description is a low hanging fruit. Later, webhook checks can also be introduced if we want to make API input restrictive and otherwise it would be a no-op even if `ValuesRef` is specified as per description. If we don't add it in API description or webhook then, users wouldn't know the field usage.

api/v1alpha1/addon_types.go
manoj-nutanix: any issue you foresee in CAREN doing it? One thing is CAREN needs to have permission to get secret but it would be pre-requisite so not much concerned. [Data format wise](https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1011/files#diff-508d401bc31620f75e53a8c8a331155b1f9b13b815193f30de415fea48cdd97dR65-R98), there's no diff in CM v/s Secret

.pre-commit-config.yaml
```diff
@@ -132,6 +132,7 @@ repos:
   hooks:
   - id: markdownlint
     stages: [pre-commit]
+    args: ["--fix"]
```
supershal, line 135: can we add this as a separate PR.

docs/content/addons/cni.md
```diff
@@ -36,6 +36,70 @@ spec:
...
+
+```yaml
+apiVersion: v1
+data:
+  values.yaml: |-
+    cni:
+      chainingMode: portmap
+      exclusive: false
+    ipam:
+      mode: kubernetes
```
supershal, line 75: we should enforce making this configmaps to be immutable. https://kubernetes.io/docs/concepts/configuration/configmap/#configmap-immutable
This will force users to not modify the existing configmap and expect changes applies to the cluster.

docs/content/addons/cni.md
```diff
@@ -36,6 +36,70 @@ spec:
...
+stringData:
+  values.yaml: |-
+    cni:
+      chainingMode: portmap
+      exclusive: false
+    ipam:
+      mode: kubernetes
+kind: Secret
+metadata:
+  name: default-cilium-cni-helm-values-template-mgmt-cluster
```
supershal, line 95:
```suggestion
  name: <CLUSTER_NAME>-cilium-cni-helm-values-template
```

docs/content/addons/cni.md
```diff
@@ -36,6 +36,70 @@ spec:
...
+  namespace: <cluster's namespace>
```
supershal, line 96: for consistency with other docs.
```suggestion
  namespace: <CLUSTER_NAMESPACE>
```

docs/content/addons/cni.md
```diff
@@ -36,6 +36,70 @@ spec:
...
+data:
+  values.yaml: |-
+    cni:
+      chainingMode: portmap
+      exclusive: false
+    ipam:
+      mode: kubernetes
+kind: ConfigMap
+metadata:
+  name: default-cilium-cni-helm-values-template-mgmt-cluster
```
supershal, line 78:
```suggestion
  name: <CLUSTER_NAME>-cilium-cni-helm-values-template
```

docs/content/addons/cni.md
```diff
@@ -36,6 +36,70 @@ spec:
...
+  namespace: <cluster's namespace>
```
supershal, line 79:
```suggestion
  namespace: <CLUSTER_NAMESPACE>
```

docs/content/addons/cni.md
```diff
@@ -36,6 +36,70 @@ spec:
...
+      - name: clusterConfig
+        value:
+          addons:
+            cni:
+              provider: Cilium
+              strategy: HelmAddon
+              values:
+                sourceRef:
+                  name: <name of Configmap/Secret present in same namespace>
+                  kind: <Configmap/Secret>
```
supershal, line 60: we have to verify if configmaps/secrets referred in the cluster variables will be moved automatically or needs `move` lables. We need this for management clusters.

api/v1alpha1/addon_types.go
supershal: The HCP excepts raw values in its CRD. We will be reading the secret data and setting it as plain text in HCP. I think we should not do that with secrets.

api/v1alpha1/addon_types.go
```diff
@@ -115,6 +115,25 @@ type CNI struct {
...
+	// +kubebuilder:validation:Optional
+	Values *AddonValues `json:"values,omitempty"`
+}
+
+// AddonValues contains the configuration values for the Helm Addon.
+type AddonValues struct {
+	// SourceRef is an object reference to Configmap/Secret inside the same namespace
+	// which contains inline YAML representing the values for the Helm chart.
+	// +kubebuilder:validation:Optional
+	SourceRef *corev1.TypedLocalObjectReference `json:"sourceRef,omitempty"`
```
dkoshkin, line 136: I was going to suggest to try and extend the [existing API](https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/a78b9d4a955fefc3227177423c3debe8bbfe6d2b/api/v1alpha1/common_types.go#L73)

But now I'm not convinced it makes sense anymore:
```
              credentials:
                secretRef:
                  name: my-registry-credentials
```

Its a "secretRef` with a single field `name`. Extending it as is wouldn't make much sense either since we don't users setting Kind/APIGroup for that `secretRef`.

We can stick with `corev1.TypedLocalObjectReference` as you have it, but we may want to copy this type into our API to be able to do the Kind string enforcement through annotations.

dkoshkin (approved):

supershal (approved):
The schema looks good to me. Please address other nits and few comments.

api/v1alpha1/addon_types.go
manoj-nutanix: yes, flux does the same thing https://github.com/fluxcd/pkg/blob/main/apis/meta/reference_types.go#L89-L93 .. wanted to bring this up in y'day grooming but did not get a chance.

docs/content/addons/cni.md
manoj-nutanix: I don't think it moves automatically, will verify.

api/v1alpha1/addon_types.go
manoj-nutanix: Until HCP starts supporting secretRef, we'll need decoding in CAREN. Additional RBAC can be put on HCP, HRP, etc. to allow only legit users.

Later, even if HCP starts supporting secretRef, `helm get values <release-name>` on actual cluster will show all the helm values in plain text so there's no way to avoid this completely. The helm chart maintainer needs to design chart such that no sensitive data gets exposed maybe using external secret store, helm secret plugin, etc. but CAREN, CAAPH and helm has no part to play in this encryption/decryption.

docs/content/addons/cni.md
manoj-nutanix: Even if we enforce making this configmaps to be immutable, users can still re-create configmap with same name but different spec and expect changes applies to the cluster. We'll need to additionally block delete operation too  for referred configmaps if we want to solve this.
OR
If we can communicate that this configmap is used only for bootstrapping and later no changes will be considered, we should be good.

docs/content/addons/cni.md
manoj-nutanix: Verified, it doesn't move automatically.. we'll need that "move" label. Will doc it. Shall we have webhook to add that label if doesn't exist or rely on users to do so?

api/v1alpha1/addon_types.go
manoj-nutanix: done 4ca5455be1ef4cdfc0d6dc154310f207b95f01c0

docs/content/addons/cni.md
supershal: >Shall we have webhook to add that label if doesn't exist or rely on users to do so?

docs/content/addons/cni.md
supershal: we also could recommend or add finalizer and owner references to the configmap to prevent deletion. I am not sure if we want to go that far. we can note it down for now and address it during implementation.

api/v1alpha1/meta_types.go
```diff
@@ -0,0 +1,19 @@
+// Copyright 2024 Nutanix. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package v1alpha1
+
+// TypedLocalObjectReference contains enough information to let you locate the
+// typed referenced object inside the same namespace.
+// This is redacted from the upstream https://pkg.go.dev/k8s.io/api/core/v1#TypedLocalObjectReference
+type TypedLocalObjectReference struct {
```
supershal, line 9: I think separate struct is fine. its gives us more flexibility w.r.t validation.
however this validation is specific for AddonValues where it checks for secrets/configmaps and cant be used by other fields.
Can we rename it to be specific to `ValuesReference` and move it to addon_types.go to ensure it is specific for addons?

api/v1alpha1/addon_types.go
supershal: 👍 .  RBAC should be good enough workaround.
My thoughts: The secrets should be generally referenced in the applications that are allowed to read in-memory. In our case, we ask users to provide secrets and write it plain text in the API server. This gives false sense of security. However RBAC should be fine until we have some native support for secrets in HCP / helm.

docs/content/addons/cni.md
dlipovetsky: An owner reference pointing to the Cluster will [allow the Kubernetes control plane to remove](https://kubernetes.io/docs/concepts/architecture/garbage-collection/#cascading-deletion) the ConfigMap when the Cluster is deleted. If the ConfigMap has multiple owner references point to Clusters, then it will be removed only after all the referenced Clusters are deleted.

If we want the ConfigMap to be garbage collected, then it makes sense for CAREN to set an ownerReference from the ConfigMap to the Cluster.

dlipovetsky: While a finalizer can prevent unintended removal of the ConfigMap,  I think it's not common to ask users to add a finalizer to the resources they create.  Typically, controllers add finalizers to resources they reconcile, so that they have a chance to reconcile delete requests, before the resources are removed from the API. 

How about, instead, we ask users to save their ConfigMap externally, e.g. to a git repository?

docs/content/addons/cni.md
dlipovetsky: The discovery rules are described in https://cluster-api.sigs.k8s.io/developer/providers/contracts/clusterctl.html#move

They include:
> * All the ConfigMap objects from the namespace being moved.

I expect the ConfigMap to be discovered by clusterctl, and moved automatically.

api/v1alpha1/meta_types.go
```diff
@@ -0,0 +1,19 @@
+// Copyright 2024 Nutanix. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package v1alpha1
+
+// TypedLocalObjectReference contains enough information to let you locate the
+// typed referenced object inside the same namespace.
+// This is redacted from the upstream https://pkg.go.dev/k8s.io/api/core/v1#TypedLocalObjectReference
```
dlipovetsky, line 8: nit: I think this is a typo.

```suggestion
// This is copied from the upstream https://pkg.go.dev/k8s.io/api/core/v1#TypedLocalObjectReference
```

api/v1alpha1/meta_types.go
dlipovetsky: > however this validation is specific for AddonValues where it checks for secrets/configmaps and cant be used by other fields.

Good point!

api/v1alpha1/addon_types.go
dlipovetsky: > The helm chart maintainer needs to design chart such that no sensitive data gets exposed

Right; the Chart itself should support references to Secrets for any sensitive information. I _think_ the values should not contain sensitive information. If some Chart requires sensitive information in the values, then the Chart should be fixed. So referencing a Secret here may not be necessary.

As a counterpoint, though, Flux can [read values](https://github.com/fluxcd/pkg/blob/main/apis/meta/reference_types.go#L89-L93) from a Secret, perhaps to support Charts that expect sensitive information in their values.

@supershal w.r.t RBAC, do we want/need more restrictive rules for helm resources? :thinking:

docs/content/addons/cni.md
supershal: After the discovery phase, the configmap will be moved only if 
```
After completing discovery, clusterctl move moves to the target cluster only the objects discovered in the previous phase that are compliant with one of the following rules:

The object is directly or indirectly linked to a Cluster object (linked through the OwnerReference chain).
```
I think CAREN adding owner reference to the configmap should be good enough.

docs/content/addons/cni.md
supershal: Thanks for the details. +1 The owner references should be good enough.

api/v1alpha1/addon_types.go
supershal: > @supershal w.r.t RBAC, do we want/need more restrictive rules for helm resources?

No. I think [current scope](https://github.com/kubernetes-sigs/cluster-api-addon-provider-helm/blob/main/config/rbac/role_binding.yaml) should be good enough.

docs/content/addons/cni.md
manoj-nutanix: I don't think we should add owner references on user created objects as these will get garbage collected and adding "move" label is more suitable than owner references + finalizers

api/v1alpha1/meta_types.go
manoj-nutanix: it was intentional, we are using only 2 fields from upstream struct so used `redacted`

docs/content/addons/cni.md
manoj-nutanix: I'm not in favor of owner reference - https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1011#discussion_r1926490325 .. We are unnecessarily trying to force fit these constructs into use-case. If  "move" label works here, what's the advantage of owner reference?

dlipovetsky: Sorry, my comment above was incorrect. I only mentioned a discovery rule, but no  "selection" rules. @supershal, you mention one "selection" rule, but I agree with @manoj-nutanix, we can rely on this "selection" rule:

> * The object has the clusterctl.cluster.x-k8s.io/move label or the clusterctl.cluster.x-k8s.io/move-hierarchy label, e.g. the CPI config secret.

supershal: sounds good. please update the example in the doc with the label.

supershal (approved):

docs/content/addons/cni.md
manoj-nutanix: sure will add it, but should we rely completely on users to add that label? what if they don't?
How about we ask users to pre-create the configmap before cluster create and CAREN webhook will add that label if missing?

dlipovetsky: > sure will add it, but should we rely completely on users to add that label?

Good question. The label is an implementation detail, so we probably should not.

> How about we ask users to pre-create the configmap before cluster create and CAREN webhook will add that label if missing?

I think we can do this. To avoid having the changes rolled back by another controller, let's use managed fields. I think the Flux docs describe this well: https://fluxcd.io/flux/faq/#why-are-kubectl-edits-rolled-back-by-flux

supershal: we already set owner references for user created secrets for CSI and CPI then copy it to the remote cluster. ex. https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/main/pkg/handlers/generic/lifecycle/csi/nutanix/handler.go#L111
I like to consider using ownerReferences again. Lets pair on it during implementation of this API and finalize the design. currently we can keep this as open question.

manoj-nutanix: sure, let's connect and finalize.

manoj-nutanix: I'll keep things as it is now.. if there's any change later after merge, let's modify.

prajnutanix (approved):

PR #1008, opened 2025-01-07
feat: COSI controller Addon

dkoshkin:
**What problem does this PR solve?**:
This PR adds the COSI controller Addon, without any providers.
The chart comes from https://github.com/mesosphere/charts/tree/master/stable/cosi

The addon does not take in any configuration so the API to enable it is just:
```yaml
apiVersion: cluster.x-k8s.io/v1beta1
kind: Cluster
metadata:
  name: <NAME>
spec:
  topology:
    variables:
      - name: clusterConfig
        value:
          addons:
            cosi: {}
```
Once we add providers, the API can be similar to what it is with CSI with a `providers:` field.

**Which issue(s) this PR fixes**:
Fixes #

**How Has This Been Tested?**:

New unit tests.
Also deployed it on a Nutanix Cluster and the Nutanix COSI driver from https://github.com/nutanix-cloud-native/cosi-driver-nutanix/pull/375

Create the credentials Secret
```
apiVersion: v1
kind: Secret
metadata:
  annotations:
  labels:
  name: objectstorage-provisioner
  namespace: cosi-driver-nutanix
stringData:
  ACCESS_KEY: "<>"
  ACCOUNT_NAME: "<>"
  ENDPOINT: "http://10.1.1.1:80"
  PC_SECRET: "<>:9440:<>:<>"
  SECRET_KEY: "<>"
type: Opaque
```

```
NUTANIX_COSI_DIR=<>
helm install cosi-driver -n cosi-driver-nutanix --create-namespace --set=secret.enabled=false --set=cosiController.enabled=false --skip-crds$NUTANIX_COSI_DIR/charts/

kubectl apply -f $NUTANIX_COSI_DIR/project/examples/bucketclass.yaml
kubectl apply -f $NUTANIX_COSI_DIR/project/examples/bucketclaim.yaml
kubectl apply -f $NUTANIX_COSI_DIR/project/examples/bucketaccessclass.yaml
kubectl apply -f $NUTANIX_COSI_DIR/project/examples/bucketaccess.yaml

# From https://github.com/nutanix-cloud-native/cosi-driver-nutanix/pull/374
kubectl apply -f https://raw.githubusercontent.com/nutanix-cloud-native/cosi-driver-nutanix/944ce9ba66bc34e59bd94519a88da8bf9fae3d0c/project/examples/awscliapppod.yaml
```
The Secret was correctly generated and the test Pod wrote and read to a new bucket.
```
$ kubectl logs awscli
[...]

PR #1014, opened 2025-01-16
test: wait for COSI controller to be ready

dkoshkin:
Missed this in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1008 since I thought that all HCP are already checked.

Fixes #

### DIFF
--- a/test/e2e/cosi_helpers.go
+++ b/test/e2e/cosi_helpers.go
@@ -1,0 +1,74 @@
+//go:build e2e
+
+// Copyright 2024 Nutanix. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package e2e
+
+import (
+	"context"
+	"fmt"
+
+	. "github.com/onsi/ginkgo/v2"
+	appsv1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/cluster-api/test/framework"
+
+	"github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/api/v1alpha1"
+	apivariables "github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/api/variables"
+)
+
+type WaitForCOSIControllerToBeReadyInWorkloadClusterInput struct {
+	COSI                 *apivariables.COSI
+	WorkloadCluster      *clusterv1.Cluster
+	ClusterProxy         framework.ClusterProxy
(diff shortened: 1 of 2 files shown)


test/e2e/cosi_helpers.go
```diff
@@ -0,0 +1,74 @@
+//go:build e2e
+
+// Copyright 2024 Nutanix. All rights reserved.
```
thunderboltsid, line 3: nit: 2025

thunderboltsid (approved):

test/e2e/cosi_helpers.go
```diff
@@ -0,0 +1,74 @@
...
+				*input.COSI.Strategy,
+			),
+		)
+	}
+
+	workloadClusterClient := input.ClusterProxy.GetWorkloadCluster(
+		ctx, input.WorkloadCluster.Namespace, input.WorkloadCluster.Name,
+	).GetClient()
+
+	WaitForDeploymentsAvailable(ctx, framework.WaitForDeploymentsAvailableInput{
```
supershal, line 65: nit:
This additional check is good. 
Does not helm Release proxy's Status will be in "Failed" state if the deployment fails?
I traced the code to https://github.com/helm/helm/blob/main/pkg/kube/wait.go#L63 where it checks for `Ready` before setting on the `status` field on the HelmReleaseProxy.

context: 
I am also looking into ways to report status of addon upgrade back to the CAREN clients.

dkoshkin: Good find! I copied this from existing code https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/162915eb4794e176422c0d8d13c4bc6c2a57cc9b/test/e2e/csi_helpers.go#L316-L324

If its not necessary we can remove all of them in a future PR, WDYT?

supershal (approved):

test/e2e/cosi_helpers.go
jimmidyson: I don't think we wait for HCPs right now so as to not block the CAAPH control loop.

supershal: we have this code to wait for HCP to be ready. https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/main/pkg/handlers/generic/lifecycle/addons/helmaddon.go#L220C6-L220C19

It is only set to be invoked for the metallb addon. 
 > HCPs right now so as to not block the CAAPH control loop.
Is it because we dont want to wait/timeout returning from `AfterControlPlaneInitialization`/`BeforeClusterUpgrade` hook?

So cluster upgrade will continue even if a critical addon deployment failed?

supershal: I am wondering if we should wait for both HCP and HRP to be ready for `upgrade addons` usecase.

PR #1018, opened 2025-01-21
fix: correctly copy Helm charts in init container

dkoshkin:
During upgrades the init container was incorrectly copying the whole directory, instead of just the tars.
```
# bundles directory in the container
$ ls /helm-charts/bundles/
charts                   helm-charts-v0.14.6.tar  helm-charts-v0.14.9.tar  helm-charts-v0.23.1.tar
# incorrectly nested directory with the new charts
$ ls /helm-charts/bundles/charts/
helm-charts-v0.14.6.tar  helm-charts-v0.14.9.tar  helm-charts-v0.25.0.tar
```

Fixes #

The file gets copied correctly when the directory does not exist:
```
$ cp -r charts/ dst/
$ ls dst
foo.tar
```

When the dir already exists (ie upgrade scenario), the whole directory gets copied:
```
$ mkdir dst
$ cp -r charts/ dst/
$ ls dst
charts
```

Copying the file works correctly into an existing dst if using .:
```
$ mkdir dst
$ cp -r charts/. dst/
$ ls dst
foo.tar
```

### DIFF
--- a/charts/cluster-api-runtime-extensions-nutanix/templates/helm-repository.yaml
+++ b/charts/cluster-api-runtime-extensions-nutanix/templates/helm-repository.yaml
@@ -63,7 +63,7 @@ spec:
       initContainers:
       - name: copy-charts
         image: "{{ .Values.helmRepository.images.bundleInitializer.repository }}:{{ default $.Chart.AppVersion .Values.helmRepository.images.bundleInitializer.tag }}"
-        command: ["/bin/cp", "-r", "/charts/",  "/helm-charts/bundles/"]
+        command: ["/bin/cp", "-r", "/charts/.",  "/helm-charts/bundles/"]
         imagePullPolicy: "{{ .Values.image.pullPolicy }}"
         volumeMounts:
         - name: charts-volume
--- a/charts/cluster-api-runtime-extensions-nutanix/templates/helm-repository.yaml
+++ b/charts/cluster-api-runtime-extensions-nutanix/templates/helm-repository.yaml
@@ -63,7 +63,18 @@ spec:
       initContainers:
       - name: copy-charts
         image: "{{ .Values.helmRepository.images.bundleInitializer.repository }}:{{ default $.Chart.AppVersion .Values.helmRepository.images.bundleInitializer.tag }}"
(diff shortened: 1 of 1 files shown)


supershal (approved):

charts/cluster-api-runtime-extensions-nutanix/templates/helm-repository.yaml
```diff
@@ -63,7 +63,7 @@ spec:
       initContainers:
       - name: copy-charts
         image: "{{ .Values.helmRepository.images.bundleInitializer.repository }}:{{ default $.Chart.AppVersion .Values.helmRepository.images.bundleInitializer.tag }}"
-        command: ["/bin/cp", "-r", "/charts/",  "/helm-charts/bundles/"]
+        command: ["/bin/cp", "-r", "/charts/.",  "/helm-charts/bundles/"]
```
jimmidyson, line 66: I don't understand how this works but it does seem to! Do you have any docs on it?

jimmidyson: I wonder if using `--no-target-directory` (`-T`) is better defined? From [https://unix.stackexchange.com/posts/94838/revisions this is equivalent](https://unix.stackexchange.com/a/94838) I guess, but perhaps more understandable than the difference between `/charts/` and `/charts/.`?

Regardless, let's add a comment here explaining why this command is specified as it is.

Optional suggestion:

```suggestion
        # Copy the charts bundled in the image to the mounted PVC.
        # The combination of flags `--recursive --no-target-directory` ensure that the bundled directory contents
        # only are copied to the destination, rather than copying the source directory itself to the destination,
        # which would lead to incorrect nesting. See https://unix.stackexchange.com/a/94838 for further
        # explanation.
        # Globs cannot be used here because we are not running in a shell and `cp` does not natively support
        # globbing.
        command: ["/bin/cp", "--recursive", "--no-target-directory", "--verbose", "/charts/",  "/helm-charts/bundles/"]
```

dkoshkin: Thank you for the `--no-target-directory` and the comment! 
Tested locally too upgrading on a live cluster.

jimmidyson (approved):

PR #1019, opened 2025-01-22
refactor: Remove api module dependency from common module

jimmidyson:
Move CAPA and CAPX specific test requests to root module. This allows
other projects to use the common module without also pulling in the
CAREN api module.

Also remove some unused code.


supershal (approved):
Thanks.

PR #1041, opened 2025-02-10
fix: Specify PriorityClass for Node Feature Discovery components

dlipovetsky:
NFD components were missing a PriorityClass.

Fixes
- https://jira.nutanix.com/browse/NCN-105633
- https://jira.nutanix.com/browse/NCN-105634

### DIFF
--- a/charts/cluster-api-runtime-extensions-nutanix/addons/nfd/values-template.yaml
+++ b/charts/cluster-api-runtime-extensions-nutanix/addons/nfd/values-template.yaml
@@ -1,3 +1,5 @@
+priorityClass: system-node-critical
+
 master:
   extraLabelNs:
     - nvidia.com


faiq (approved):

supershal (approved):

charts/cluster-api-runtime-extensions-nutanix/addons/nfd/values-template.yaml
```diff
@@ -1,3 +1,5 @@
+priorityClass: system-node-critical
```
supershal, line 1:
```suggestion
priorityClassName: system-node-critical
```
https://kubernetes-sigs.github.io/node-feature-discovery/v0.16/deployment/helm.html
we also need to set
```
worker:
  priorityClassName: system-node-critical
```

dlipovetsky: Thanks for being vigilant. However, having looked at the Chart, the top level `priorityClassName` field value is propagated to all components--master, workers, and gc :smile:

supershal: :) . Thanks for checking the code. I will file bug/fix upstream. 
We can just change top level `priorityClassName` then.

prajnutanix (approved):

charts/cluster-api-runtime-extensions-nutanix/addons/nfd/values-template.yaml
dlipovetsky: You think there's a bug to report?

It seems acceptable to have both the top-level field and lower-level `priorityClassName` fields in the Helm chart values. :shrug:

PR #1039, opened 2025-02-10
fix: Correctly configure non-mirror registry certificates

jimmidyson:
This PR updates the runtime extensions to include two mutation handlers for Nutanix clusters:

- `nutanixclusterconfigpatch-gp.cluster-api-runtime-extensions-nutanix` - original name and original (buggy functionality). This will still be referenced from existing CCs and so must retain the same name.
- `nutanixclusterv2configpatch-gp.cluster-api-runtime-extensions-nutanix` - new name with fixed functionality. This is now referenced from the example CC in this project and can be used by existing clusters by rebasing on a CC that references this handler name.

The new handler fixes are:

- CA certificates are now written to `/etc/containerd/certs.d/<registryHost>/ca.crt`
as per https://github.com/containerd/containerd/blob/main/docs/hosts.md#support-for-dockers-certificate-file-pattern. This does not require a `hosts.toml` file for configuration, although that explicit configuration may be preferred (can do in future if so).
- Remove non-mirror registry config from `/etc/containerd/certs.d/_default/hosts.toml`
which was causing all registries to be configured as mirror registry.

PR #1042, opened 2025-02-11
fix: don't generate empty _default containerd mirror file

dkoshkin:
Now that we're fixing how the `_default` file is being generated just for mirrors, we should skip creating the file if only registries are provided.

Depends on #1039.

Fixes #


jimmidyson (approved):

supershal (approved):

dlipovetsky (dismissed):
Makes sense. Thanks!

supershal (approved):

dlipovetsky (approved):

PR #1043, opened 2025-02-11
feat: Update COSI controller Addon

dlipovetsky:
Chart now defines priorityClass and tolerations for the controller Pod.

Fixes https://jira.nutanix.com/browse/NCN-105644

### DIFF
--- a/make/addons.mk
+++ b/make/addons.mk
@@ -25,7 +25,7 @@ export KUBE_VIP_VERSION := v0.8.3
 
 export METALLB_CHART_VERSION := 0.14.8
 
-export COSI_CONTROLLER_VERSION := 0.0.1-alpha.4
+export COSI_CONTROLLER_VERSION := 0.0.1-alpha.5
 
 .PHONY: addons.sync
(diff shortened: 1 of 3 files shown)


make/addons.mk
```diff
@@ -25,7 +25,7 @@ export KUBE_VIP_VERSION := v0.8.3
 
 export METALLB_CHART_VERSION := 0.14.8
 
-export COSI_CONTROLLER_VERSION := 0.0.1-alpha.4
+export COSI_CONTROLLER_VERSION := 0.0.1-alpha.5
```
jimmidyson, line 28: This version doesn't defaults to the necessary priority class, etc? No need for overrides here?

jimmidyson (changes requested):
Please run `make addons.sync`.

make/addons.mk
dlipovetsky: No need for overrides; the chart has the defaults we need: https://github.com/mesosphere/charts/blob/2970620efaa1ad449abe0025ba803a3b26d70d47/stable/cosi/values.yaml#L10-L18

dlipovetsky:
> make addons.sync

Will do. I created this PR before the chart updated merged :smile:

supershal (approved):

dkoshkin (approved):

jimmidyson (approved):

PR #1045, opened 2025-02-11
fix: Configure priorityClassName for Cilium Hubble

dlipovetsky:
Assigns the `system-cluster-critical` PriorityClassName to Hubble. This is the same class name assigned to Cilium itself.

Fixes https://jira.nutanix.com/browse/NCN-105637

### DIFF
--- a/charts/cluster-api-runtime-extensions-nutanix/addons/cni/cilium/values-template.yaml
+++ b/charts/cluster-api-runtime-extensions-nutanix/addons/cni/cilium/values-template.yaml
@@ -13,6 +13,7 @@ hubble:
     enabled: true
     image:
       useDigest: false
+    priorityClassName: system-cluster-critical
 ipam:
   mode: kubernetes
 image:


hack/addons/kustomize/cilium/kustomization.yaml.tmpl
```diff
@@ -29,5 +29,6 @@ helmCharts:
       enabled: false
       relay:
         enabled: false
+        priorityClassName: system-cluster-critical
```
jimmidyson, line 32: This is only applicable to the CRS generation and will have no effect either for that because the line above disables relay in CRS!

You want to change the Helm addon values in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/main/charts/cluster-api-runtime-extensions-nutanix/addons/cni/cilium/values-template.yaml.

jimmidyson (changes requested):

hack/addons/kustomize/cilium/kustomization.yaml.tmpl
dlipovetsky: Thanks.

dlipovetsky: Fixed.

jimmidyson (approved):

supershal (approved):

PR #1050, opened 2025-02-13
fix: Use new patch handler for AWS and Docker ClusterClasses

jimmidyson:
This means no automatic rollout of clusters using the existing
ClusterClasses and will not break any downstream project tests
that might be checking for that.

Same approach as previously done for Nutanix ClusterClass -
should have done this at the same time.

### DIFF
--- a/pkg/handlers/deleteinv0280/aws/mutation/metapatch_handler.go
+++ b/pkg/handlers/deleteinv0280/aws/mutation/metapatch_handler.go
@@ -1,4 +1,4 @@
-// Copyright 2023 Nutanix. All rights reserved.
+// Copyright 2025 Nutanix. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 package mutation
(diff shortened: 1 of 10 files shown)


pkg/handlers/deleteinv0280/aws/mutation/metapatch_handler.go
```diff
@@ -0,0 +1,43 @@
+// Copyright 2023 Nutanix. All rights reserved.
```
thunderboltsid, line 1: nit: 2025

pkg/handlers/deleteinv0280/docker/mutation/metapatch_handler.go
```diff
@@ -0,0 +1,29 @@
+// Copyright 2023 Nutanix. All rights reserved.
```
thunderboltsid, line 1: same

thunderboltsid (approved):

thunderboltsid (approved):

PR #1062, opened 2025-02-28
feat: properly support kube-vip upgrades

dkoshkin:
We ran into an issue where upgrading CAREN causes an unwanted rollout of all control plane Nodes.
This happens when CAREN is shipped with a new kube-vip version and the difference in the static Pod spec changes the `file` contents in KCP and forces new Nodes to be created.

We do not want this behavior and only want this rollout to happen at controlled times.

In a recent change we introduced a mechanism of making backend changes by using versioned handlers.https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1039/commits/7bc90941f0f848d5be1b05cad87b8ae26b782bd9

This change follows a similar pattern. A new handler is introduced that will instead of reading the kube-vip template from a global ConfigMap will read it directly from the KCPTemplate associated with the ClusterClass.
We can then continue to change the kube-vip version (or any other spec changes there) and downstream projects will just pick up whatever changes through a new ClusterClass.

Fixes #

Tested locally by creating a cluster with the new and old handlers.

This PR introduces a "v3" version of the handlers with the modified kube-vip handler, and preserves the previous behavior from v2.


pkg/handlers/v2/generic/mutation/controlplanevirtualip/inject.go
```diff
@@ -0,0 +1,189 @@
+// Copyright 2023 Nutanix. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package controlplanevirtualip
```
dkoshkin, line 4: This is unchanged from the previous version and is the only handler that was copied over.

pkg/handlers/generic/mutation/controlplanevirtualip/providers/kubevip_test.go
```diff
@@ -26,7 +23,7 @@ func Test_GenerateFilesAndCommands(t *testing.T) {
 		name                        string
 		controlPlaneEndpointSpec    v1alpha1.ControlPlaneEndpointSpec
 		cluster                     *clusterv1.Cluster
-		configMap                   *corev1.ConfigMap
+		kcp                         *controlplanev1.KubeadmControlPlaneTemplate
```
dkoshkin, line 26: Notice how the tests output did not change, only template input

pkg/handlers/nutanix/mutation/controlplanevirtualip/inject.go
```diff
@@ -4,19 +4,12 @@
...
 )
 
-func NewPatch(
-	cl ctrlclient.Client,
-	cfg *controlplanevirtualip.Config,
-) *controlplanevirtualip.ControlPlaneVirtualIP {
+func NewPatch() *controlplanevirtualip.ControlPlaneVirtualIP {
 	return controlplanevirtualip.NewControlPlaneVirtualIP(
-		cl,
-		cfg,
```
dkoshkin, line 19: We no longer need to pass these along

pkg/handlers/generic/mutation/controlplanevirtualip/inject_test.go
```diff
@@ -52,28 +60,24 @@ var _ = Describe("Generate ControlPlane virtual IP patches", func() {
...
+						Path:         "/spec/template/spec/kubeadmConfigSpec/files/0/owner",
+						ValueMatcher: gomega.Equal("root:root"),
+					},
+					{
+						Operation:    "add",
+						Path:         "/spec/template/spec/kubeadmConfigSpec/files/0/permissions",
+						ValueMatcher: gomega.Equal("0600"),
+					},
+					{
+						Operation: "replace",
```
dkoshkin, line 76: The Patches look different now, with this replace, but the resulting KCP is the same.

pkg/handlers/generic/mutation/controlplanevirtualip/inject.go
```diff
@@ -187,3 +167,36 @@ func (h *ControlPlaneVirtualIP) Mutate(
...
+		for _, path := range filePathsToDelete {
+			if obj.Spec.Template.Spec.KubeadmConfigSpec.Files[i].Path == path {
+				obj.Spec.Template.Spec.KubeadmConfigSpec.Files = slices.Delete(
+					obj.Spec.Template.Spec.KubeadmConfigSpec.Files, i, i+1,
+				)
+				break
+			}
+		}
+	}
+}
```
thunderboltsid, line 182: Consider rewriting it for better readability
```go
func deleteFiles(obj *controlplanev1.KubeadmControlPlaneTemplate, filePathsToDelete ...string) {
	toDelete := make(map[string]struct{}, len(filePathsToDelete))
	for _, path := range filePathsToDelete {
		toDelete[path] = struct{}{}
	}

	// Filter the Files slice in place.
	files := obj.Spec.Template.Spec.KubeadmConfigSpec.Files
	filtered := files[:0] // reuse the underlying array (or create a new one, either is fine)
	for _, file := range files {
		if _, ok := toDelete[file.Path]; !ok {
			filtered = append(filtered, file)
		}
	}
	obj.Spec.Template.Spec.KubeadmConfigSpec.Files = filtered
}
```

thunderboltsid: alternatively, we can pull out the update logic between delete and merge
```go
func updateFiles(files []bootstrapv1.File, updateFn func(map[string]bootstrapv1.File)) []bootstrapv1.File {
	fileMap := make(map[string]bootstrapv1.File, len(files))
	for _, file := range files {
		fileMap[file.Path] = file
	}

	updateFn(fileMap)
	updated := make([]bootstrapv1.File, 0)
	for _, file := range fileMap {
		updated = append(updated, file)
	}

	return updated
}

func deleteFiles(obj *controlplanev1.KubeadmControlPlaneTemplate, filePathsToDelete ...string) {
	obj.Spec.Template.Spec.KubeadmConfigSpec.Files = updateFiles(
		obj.Spec.Template.Spec.KubeadmConfigSpec.Files,
		func(fileMap map[string]bootstrapv1.File) {
			// Remove each file whose path is in filePathsToDelete.
			for _, path := range filePathsToDelete {
				delete(fileMap, path)
			}
		},
	)
}

func mergeFiles(obj *controlplanev1.KubeadmControlPlaneTemplate, filesToMerge ...bootstrapv1.File) {
	obj.Spec.Template.Spec.KubeadmConfigSpec.Files = updateFiles(
		obj.Spec.Template.Spec.KubeadmConfigSpec.Files,
		func(fileMap map[string]bootstrapv1.File) {
			// Replace or add each file from filesToMerge.
			for _, file := range filesToMerge {
				fileMap[file.Path] = file
			}
		},
	)
}
```

pkg/handlers/generic/mutation/controlplanevirtualip/inject_test.go
```diff
@@ -268,3 +275,247 @@ spec:
...
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			deleteFiles(tt.obj, tt.filesToDelete...)
+			assert.Equal(t, tt.expectedFiles, tt.obj.Spec.Template.Spec.KubeadmConfigSpec.Files)
```
thunderboltsid, line 372: Consider using ElementsMatch so test doesn't rely on order
```go
assert.ElementsMatch(
				t,
				tt.expectedFiles,
				tt.obj.Spec.Template.Spec.KubeadmConfigSpec.Files,
			)
```

pkg/handlers/generic/mutation/controlplanevirtualip/inject_test.go
```diff
@@ -268,3 +275,247 @@ spec:
...
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			mergeFiles(tt.obj, tt.files...)
+			assert.Equal(t, tt.expectedFiles, tt.obj.Spec.Template.Spec.KubeadmConfigSpec.Files)
```
thunderboltsid, line 518: Consider using ElementsMatch so test doesn't rely on order
```go
assert.ElementsMatch(
				t,
				tt.expectedFiles,
				tt.obj.Spec.Template.Spec.KubeadmConfigSpec.Files,
			)
```

pkg/handlers/generic/mutation/controlplanevirtualip/providers/kubevip.go
```diff
@@ -39,39 +38,32 @@ var (
...
 	}
 }
 
-func (p *kubeVIPFromConfigMapProvider) Name() string {
+func (p *kubeVIPFromKCPTemplateProvider) Name() string {
 	return "kube-vip"
 }
 
 // GenerateFilesAndCommands returns files and pre/post kubeadm commands for kube-vip.
 // It reads kube-vip template from a ConfigMap and returns the content a File, templating the required variables.
```
thunderboltsid, line 58: update the comment too.

pkg/handlers/v2/generic/mutation/controlplanevirtualip/inject.go
```diff
@@ -0,0 +1,189 @@
+// Copyright 2023 Nutanix. All rights reserved.
```
thunderboltsid, line 1: nit: s/2023/2025

thunderboltsid: nvm; archived file

pkg/handlers/generic/mutation/controlplanevirtualip/inject.go
dkoshkin: The introduction of the `map` here makes it indeterministic  and will lead to unwanted rollouts during periodic reconciles.

pkg/handlers/generic/mutation/controlplanevirtualip/inject_test.go
dkoshkin: We do want to preserve the order, please see https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1062#discussion_r1989907485

pkg/handlers/generic/mutation/controlplanevirtualip/inject_test.go
dkoshkin: We do want to preserve the order, please see https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1062#discussion_r1989907485

pkg/handlers/generic/mutation/controlplanevirtualip/providers/kubevip.go
dkoshkin: Thanks, done in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1062/commits/1aae2c6919ae4dce46d266cb2576992fd541d553

pkg/handlers/generic/mutation/controlplanevirtualip/inject.go
dkoshkin: Oops, I see your first comment, let me try that.

hack/examples/bases/nutanix/clusterclass/kustomization.yaml.tmpl
```diff
@@ -46,16 +46,23 @@ patches:
...
+# The handler will set the variables if needed, or remove it.
 - target:
     kind: KubeadmControlPlaneTemplate
   patch: |-
     - op: test
       path: /spec/template/spec/kubeadmConfigSpec/files/0/path
       value: "/etc/kubernetes/manifests/kube-vip.yaml"
     - op: "remove"
-      path: "/spec/template/spec/kubeadmConfigSpec/files/0"
+      path: "/spec/template/spec/kubeadmConfigSpec/files/0/owner"
```
supershal, line 58: why are we removing the owner. should we set it to `root:root` explicitly or it will be defaulted to it?
also we should be setting file permissions `0600` too. https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/main/pkg/handlers/generic/mutation/controlplanevirtualip/providers/kubevip.go#L22-L24

pkg/handlers/generic/mutation/controlplanevirtualip/inject.go
```diff
@@ -117,23 +95,27 @@ func (h *ControlPlaneVirtualIP) Mutate(
...
 		vars,
 		&holderRef,
 		selectors.ControlPlane(),
 		log,
 		func(obj *controlplanev1.KubeadmControlPlaneTemplate) error {
+			if controlPlaneEndpointVar.VirtualIPSpec == nil {
+				log.V(5).Info("ControlPlane VirtualIP not set")
+				// if VirtualIPSpec is not set, delete all template files
+				// as we do not want them to end up in the generated KCP
+				deleteFiles(obj, providers.TemplateFileNames...)
```
supershal, line 109: This is actually removing only KubeVIP related files only right?
the comment should reflect that.
Also it would be nice to rename `TemplateFileNames` to `VirtualIPProviderFileNames`

hack/examples/bases/nutanix/clusterclass/kustomization.yaml.tmpl
dkoshkin: We only need the templated Pod spec, all other fields here are set by the handler.
I removed it intentionally so that we don't have unused, (and potentially different information) in this KCPTemplate.

pkg/handlers/generic/mutation/controlplanevirtualip/inject.go
```diff
@@ -187,3 +167,36 @@ func (h *ControlPlaneVirtualIP) Mutate(
 		},
 	)
 }
+
+func deleteFiles(obj *controlplanev1.KubeadmControlPlaneTemplate, filePathsToDelete ...string) {
```
supershal, line 171: nit: 
It would be nice if the function accepts `srcFiles []bootstrapv1.File` args and keep assignment to obj.Spec.Template.Spec.KubeadmConfigSpec.Files from the returned value.
```
deleteFiles(files []bootstrapv1.File, filePathsToDelete ...string) []bootstrapv1.File
```
it can be useful to use it in `KubeadmConfigTemplate`

pkg/handlers/generic/mutation/controlplanevirtualip/inject.go
```diff
@@ -187,3 +167,36 @@ func (h *ControlPlaneVirtualIP) Mutate(
...
+				)
+				break
+			}
+		}
+	}
+}
+
+// mergeFiles will merge the files into the KubeadmControlPlaneTemplate,
+// overriding any file with the same path and appending the rest.
+func mergeFiles(obj *controlplanev1.KubeadmControlPlaneTemplate, filesToMerge ...bootstrapv1.File) {
```
supershal, line 186: nit. same as above.

```
func mergeFiles(files []bootstrapv1.File, filesToMerge ...bootstrapv1.File) []bootstrapv1.File
```

pkg/handlers/generic/mutation/controlplanevirtualip/inject.go
dkoshkin: Good idea, done in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1062/commits/49381076065595dd41dc82dc169e480170ab9ca1

pkg/handlers/generic/mutation/controlplanevirtualip/inject.go
dkoshkin: Good idea, done in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1062/commits/49381076065595dd41dc82dc169e480170ab9ca1

pkg/handlers/generic/mutation/controlplanevirtualip/inject.go
dkoshkin: Makes sense, I didn't include "KubeVIP" here because by this points in the code its dealing with the virtualIPProvider interface, but updated the comment with more details, PTAL.

dkoshkin: Renamed the varibale too, thanks for the suggestion.

hack/examples/bases/nutanix/clusterclass/kustomization.yaml.tmpl
supershal: 👍 I noticed it in handler code (hidden in github diff view). Also the unit tests are asserting this. Thanks.

supershal (approved):
Thanks the solution and suggesting alternatives.

jimmidyson (approved):
This is great!

PR #1063, opened 2025-03-02
fix: correctly handle multiple registries with the same Host

dkoshkin:
Write out a single CA file for multiple registries with the same url.Host. Return an error if the CA content for those registries do not match.

This fix is needed after another recent fix https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1039/commits/17bc80da9bfe98670c2af6c7d16b67475acd4564#diff-3b57926153158559026700fdb9392d259a62201d7eb485f3d0d62fdeaa976ffdL57-L64 that now correctly generates the CA files without the url.Path.

Fixes #

New unit tests.

### DIFF
--- a/pkg/handlers/generic/mutation/mirrors/containerd_files.go
+++ b/pkg/handlers/generic/mutation/mirrors/containerd_files.go
@@ -6,13 +6,14 @@ package mirrors
 import (
 	"bytes"
 	_ "embed"
+	"errors"
 	"fmt"
 	"net/url"
 	"path"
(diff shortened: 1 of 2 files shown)


dkoshkin:
This fix will also need to be backported to `release/v0.27.x`

jimmidyson (approved):

jimmidyson (approved):

mhrabovcin (approved):

PR #1073, opened 2025-03-13
fix: Retain existing join and init kubeadm config when adding taints

jimmidyson:
Previously a dodgy piece of logic overwrote the whole of the join and init
kubeadm config when adding taints due to an incorrect `!=` which should have
been `==`.


fatz (approved):

dkoshkin (approved):

supershal (approved):
Thanks.

PR #1074, opened 2025-03-13
fix: Tolerate all NoSchedule taints for NFD and CSI deployments

jimmidyson:
This allows tainting nodepools and still using e.g. GPUs or CSI.


dkoshkin (approved):
Thanks

supershal (approved):

PR #1085, opened 2025-03-20
feat: use latest sidecar images for Nutanix CSI

dkoshkin:
Wanted the team's thoughts on this approach. The main driver is picking up CVE fixes, but also reducing the air-gapped bundle size not needing to have duplicate image versions.

Fixes #


jimmidyson:
My concern is testing compatibility: how do we ensure that bumping these images here doesn't break anything before we merge it?

dkoshkin:
The upstream CSI team will update and test the sidecars.

PR #1116, opened 2025-05-02
feat: add registry addon

dkoshkin:
This PR adds a new addon `registryMirror` that deploys https://github.com/distribution/distribution as a StatefulSet
 and a [sidecar container to sync images](https://github.com/regclient/regclient/blob/main/docs/regsync.md) across instances.
 
```
$ kubectl get pods -n registry-system 
NAME                                           READY   STATUS    RESTARTS   AGE
cncf-distribution-registry-docker-registry-0   2/2     Running   0          2m13s
cncf-distribution-registry-docker-registry-1   2/2     Running   0          8s

```
 
This addon is designed to only be a mirror and not used a regular registry, hence the name and the lack of external access to the Service.
In a follow up PR, the in-cluster Service will be used a Containerd mirror. 
In a follow up PR, it will also be deployed with randomly generated credentials to further prevent direct use.
 
Fixes #


api/v1alpha1/addon_types.go
```diff
@@ -335,3 +340,10 @@ type AddressRange struct {
 	// +kubebuilder:validation:Format=ipv4
 	End string `json:"end"`
 }
+
+type RegistryMirror struct {
```
dkoshkin, line 344: Did not add the `AddonStrategy` field since only the HelmAddon was implemented

charts/cluster-api-runtime-extensions-nutanix/addons/registry-mirror/distribution/values-template.yaml
```diff
@@ -0,0 +1,12 @@
+replicaCount: 2
```
supershal, line 1: should we dynamically set the replicaCount = # of CP nodes? 
also set anti-affinities to schedule one statefulset per CP node to help during downtime, upgrade and single node cluster? https://github.com/mesosphere/charts/blob/master/staging/docker-registry/templates/statefulset.yaml  we need to enhance the chart.

charts/cluster-api-runtime-extensions-nutanix/templates/helm-config.yaml
```diff
@@ -27,6 +27,10 @@ data:
     ChartName: cosi
     ChartVersion: 0.0.1-alpha.5
     RepositoryURL: '{{ if .Values.helmRepository.enabled }}oci://helm-repository.{{ .Release.Namespace }}.svc/charts{{ else }}https://mesosphere.github.io/charts/stable/{{ end }}'
+  distribution-registry-mirror: |
+    ChartName: docker-registry
```
supershal, line 31: nit: original chart name could be `distribution-registry` :)

api/v1alpha1/constants.go
```diff
@@ -28,6 +28,8 @@ const (
 	ClusterAutoscalerVariableName = "clusterAutoscaler"
 	// ServiceLoadBalancerVariableName is the Service LoadBalancer config patch variable name.
 	ServiceLoadBalancerVariableName = "serviceLoadBalancer"
+	// RegistryMirrorVariableName is the OCI registry config patch variable name.
```
supershal, line 31: we have `globalRegistryMirror` and now `registryMirror`. that could be confusing for end users. I cant of think of any better suggestion now. however can we take it offline and think about specific name that shows intention of this registry?

pkg/handlers/generic/lifecycle/registrymirror/distribution/handler.go
```diff
@@ -0,0 +1,148 @@
...
+		ctx,
+		"",
+		n.client,
+		ctrlclient.ObjectKeyFromObject(cluster),
+	)
+	if err != nil {
+		return fmt.Errorf("error creating remote cluster client: %w", err)
+	}
+
+	err = handlersutils.EnsureNamespaceWithMetadata(
```
supershal, line 82: Why do we need to explicitly create helm release namespace? Does not HRP creates it? also are we expecting additional metadata on the namespace later? currently it is set to `nil`

charts/cluster-api-runtime-extensions-nutanix/addons/registry-mirror/distribution/values-template.yaml
supershal: correction: affinity support is in chart. https://github.com/mesosphere/charts/blob/master/staging/docker-registry/values.yaml#L217
we should add affinity in this Values.yaml.

pkg/handlers/generic/lifecycle/registrymirror/doc.go
```diff
@@ -0,0 +1,4 @@
+// Copyright 2025 Nutanix. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
```
supershal, line 3: can we add package level comments or remote this file.

charts/cluster-api-runtime-extensions-nutanix/addons/registry-mirror/distribution/values-template.yaml
dkoshkin: Can I create a follow up ticket? This probably needs some more investigation on the cases that we don't really have e2e tests for (ie upgrades, more cp Nodes, etc)

charts/cluster-api-runtime-extensions-nutanix/templates/helm-config.yaml
dkoshkin: Yeah :( , this name comes from the upstream chart https://github.com/twuni/docker-registry.helm/tree/main
It would have been too intrusive to change it and still be able to sync upstream changes.

pkg/handlers/generic/lifecycle/registrymirror/distribution/handler.go
dkoshkin: Ahh good eye, I copied this from a different handler that needs to install additional resources into this namespace.
Let me remove it here and then we can add it back in a follow up if we do need to guarantee that it exists.

pkg/handlers/generic/lifecycle/registrymirror/doc.go
dkoshkin: 👍 I'll add a package level comment.

api/v1alpha1/constants.go
dkoshkin: Here is an example of what that would look like. When thinking of the name I intentionally based it on the existing field, but I can see now that it may be confusing since its not clear what if these should be set together or separately.

I do want to convey with the name that this registry is specifically designed to be a mirror and not just a standalone registry (that can be a different addon if we ever want to add that in the future).

```
    - name: clusterConfig
      value:
        addons:
          registryMirror: {}
          ....
        globalImageRegistryMirror:
          url: https://example.com
          credentials:
            secretRef:
              name: my-mirror-ca-cert
```

test/e2e/statefulset_helpers.go
```diff
@@ -0,0 +1,44 @@
...
+		if err := input.Getter.Get(ctx, key, input.StatefulSet); err == nil {
+			if input.StatefulSet.Status.ObservedGeneration != input.StatefulSet.Generation {
+				return false
+			}
+
+			return input.StatefulSet.Status.AvailableReplicas == input.StatefulSet.Status.Replicas
+		}
+		return false
+	}, intervals...).Should(BeTrue())
+	Logf("Deployment %s is now available, took %v", key, time.Since(start))
```
supershal, line 43: nit
```suggestion
	Logf("StatefulSet %s is now available, took %v", key, time.Since(start))
```

supershal (approved):
Thanks. The code changes looks good. Lets discuss if we can simplify providing all three `registries`, `registryMirror` and `globalImageRegistry`

api/v1alpha1/constants.go
supershal: >specifically designed to be a mirror 

can you please remind why we designed it to be a mirror?

supershal: If we have two mirrors, registryDistribution mirror and global registry mirror, how containerd will work? 
Which mirror it would select to cache the images it cant find and pull it from the backend registry?

I believe we initially decided to use single global image registry mirror per cluster. Trying to recall reason behind that call. 🤔

dkoshkin: > can you please remind why we designed it to be a mirror?

Some of the images needed for the cluster will be baked into the OS, but most images will need to come from this internal registry.
Setting it as a mirror automatically (done in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1117) makes it so that users do really not need to understand the internals of how this registry is deployed and don't need to manually set it as a mirror.

dkoshkin: RE: multiple mirrors.

The single mirror limitation comes from just CAREN's API and IIRC it was done that way for simplicity.
But the [handler code](https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/3640635d3b4f00fa177a9545c345e41351829241/pkg/handlers/generic/mutation/mirrors/templates/hosts.toml.gotmpl#L1) and [Containerd](https://github.com/containerd/containerd/blob/main/docs/hosts.md#setup-a-local-mirror-for-docker) both handle a list, Containerd will just go down the list trying each mirror until it can find the image, with the final one being the registry from the image name.

In this case CAREN will add this internal registry as the last mirror in the list.

api/v1alpha1/addon_types.go
```diff
@@ -100,6 +102,9 @@ type GenericAddons struct {
 
 	// +kubebuilder:validation:Optional
 	ServiceLoadBalancer *ServiceLoadBalancer `json:"serviceLoadBalancer,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	RegistryMirror *RegistryMirror `json:"registryMirror,omitempty"`
```
jimmidyson, line 107: I'm not keen on this naming. Although we're using it as a registry mirror, from CAREN's perspective this is just a registry... I suggest changing it to simply `registry`
```suggestion
	Registry *RegistryAddon `json:"registry,omitempty"`
```

api/v1alpha1/constants.go
jimmidyson: I proposed above to use `registry` as the addon name. I think that makes things clearer. Although we are expecting to use this as a registry mirror, from CAREN's perspective it could equally be used as a simple way to deploy a registry.

docs/content/addons/registry-mirror.md
```diff
@@ -0,0 +1,32 @@
...
+  name: <NAME>
+spec:
+  topology:
+    variables:
+      - name: clusterConfig
+        value:
+          addons:
+            registryMirror: {}
+```
+
```
jimmidyson, line 30: Let's add some more details about what will be deployed here, e.g. multiple replicas, synced, etc.

api/v1alpha1/constants.go
dkoshkin: Are we ok calling it `registry` and always configuring it as a `mirror` https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1117?
Do we want to make that configurable?

That was the intent of calling it `registryMirror` since users would not have control if its a mirror or not.

dkoshkin: @supershal does `registry` sound good to you?

docs/content/addons/registry-mirror.md
dkoshkin: Done

api/v1alpha1/addon_types.go
dkoshkin: Done

test/e2e/statefulset_helpers.go
```diff
@@ -0,0 +1,44 @@
...
+			if input.StatefulSet.Status.ObservedGeneration != input.StatefulSet.Generation {
+				return false
+			}
+
+			return input.StatefulSet.Status.AvailableReplicas == input.StatefulSet.Status.Replicas
+		}
+		return false
+	}, intervals...).Should(BeTrue())
+	Logf("StatefulSet %s is now available, took %v", key, time.Since(start))
+}
```
manoj-nutanix, line 44: we can use [this](https://github.com/helm/helm/blob/b25a51b291b930ef05ed0f4a9ea7cce073c10f63/pkg/kube/ready.go#L85) method then, we won't need to maintain separate funcs for each Kind.

test/e2e/registry.go
```diff
@@ -0,0 +1,58 @@
...
+
+	WaitForStatefulSetsAvailable(ctx, WaitForStatefulSetAvailableInput{
+		Getter: workloadClusterClient,
+		StatefulSet: &appsv1.StatefulSet{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "cncf-distribution-registry-docker-registry",
+				Namespace: "registry-system",
+			},
+		},
+	}, input.StatefulSetIntervals...)
```
manoj-nutanix, line 57: If we set [wait: true](https://github.com/kubernetes-sigs/cluster-api-addon-provider-helm/blob/7fa137968e2aabe79d179d81e9cad810b2a6849d/api/v1alpha1/helmchartproxy_types.go#L106-L112) , HRP will report Ready only if when [func](https://github.com/helm/helm/blob/b25a51b291b930ef05ed0f4a9ea7cce073c10f63/pkg/kube/ready.go#L85) returns true.. we won't need these WaitForStatefulSetsAvailable funcs then

manoj-nutanix (approved):
LG

charts/cluster-api-runtime-extensions-nutanix/addons/registry/cncf-distribution/values-template.yaml
```diff
@@ -0,0 +1,12 @@
+replicaCount: 2
+persistence:
+  enabled: true
+  size: 50Gi
```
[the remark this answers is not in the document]
dkoshkin: Thats correct, each release is about 15GB of images, we can re-evaluate in a follow up once we start doing some e2e testing if this needs to change.

test/e2e/statefulset_helpers.go
dkoshkin: Good idea! Created a follow up ticket.

test/e2e/registry.go
dkoshkin: Good idea! Created a follow up ticket.

pkg/handlers/generic/lifecycle/registry/cncfdistribution/handler.go
```diff
@@ -0,0 +1,121 @@
...
+func (c *Config) AddFlags(prefix string, flags *pflag.FlagSet) {
+	flags.StringVar(
+		&c.defaultValuesTemplateConfigMapName,
+		prefix+".default-values-template-configmap-name",
+		"default-cncf-distribution-registry-helm-values-template",
+		"default values ConfigMap name",
+	)
+}
+
+type Distribution struct {
```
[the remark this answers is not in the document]
dkoshkin: Good catch, changed in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1116/commits/c97c129ee159590541ade04b14e397837720e397

pkg/handlers/generic/lifecycle/registry/utils/ip.go
```diff
@@ -0,0 +1,53 @@
...
+		return "", fmt.Errorf("error getting a service IP for a cluster: %w", err)
+	}
+
+	return serviceIP, nil
+}
+
+func getServiceIP(serviceSubnetStrings []string) (string, error) {
+	if len(serviceSubnetStrings) == 0 {
+		serviceSubnetStrings = []string{v1alpha1.DefaultServicesSubnet}
+	}
```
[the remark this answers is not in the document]
dkoshkin: Good idea https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1116/commits/90a9f1442231fe2b90271adb684c42afb781d192

PR #1117, opened 2025-05-02
feat: use registryMirror addon as Containerd mirror

dkoshkin:
Depends on https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1116

Automatically sets the registryMirror as a Containerd mirror. We're not updating the Cluster object because this should not be a user controller configuration and the IP used as the mirror is determined based on the addon handler and the Service CIDRs.

Tested in a Docker cluster:
```
$ kubectl port-forward \
  --address=0.0.0.0 \
  --namespace registry-mirror-system \
  pod/registry-mirror-docker-registry-0 5000:5000
# Push an image tag that doesn't exist in dockerhub  
$ crane copy nginx:latest 0.0.0.0:5000/library/nginx:dkoshkin --insecure
$ kubectl run nginx-working --image=docker.io/library/nginx:dkoshkin
$ kubectl run nginx-should-be-broken --image=docker.io/library/nginx:dne
$ kubectl get pods 
NAME                                                              READY   STATUS              RESTARTS   AGE
cluster-autoscaler-0196931c-cb53-7abf-aa89-49c82c42ced5-86w5j8c   0/1     ContainerCreating   0          19m
nginx-should-be-broken                                            0/1     ErrImagePull        0          11m
nginx-working                                                     1/1     Running             0          11m
```

Fixes #

### DIFF
--- a/pkg/handlers/generic/mutation/mirrors/inject.go
+++ b/pkg/handlers/generic/mutation/mirrors/inject.go
@@ -143,11 +143,7 @@ func (h *globalMirrorPatchHandler) Mutate(
 			return err
 		}
 
-		registryConfig, err := containerdConfigFromRegistryMirrorAddon(
-			ctx,
-			h.client,
-			cluster,
-		)
+		registryConfig, err := containerdConfigFromRegistryMirrorAddon(cluster)
 		if err != nil {
 			return err
 		}
@@ -266,11 +262,7 @@ func containerdConfigFromImageRegistry(
 	return configWithOptionalCACert, nil
 }
 
-func containerdConfigFromRegistryMirrorAddon(
-	_ context.Context,
-	_ ctrlclient.Client,
-	cluster *clusterv1.Cluster,
-) (containerdConfig, error) {
+func containerdConfigFromRegistryMirrorAddon(cluster *clusterv1.Cluster) (containerdConfig, error) {
 	serviceIP, err := registrymirrorutils.ServiceIPForCluster(cluster)
 	if err != nil {
 		return containerdConfig{}, fmt.Errorf("error getting service IP for the registry mirror addon: %w", err)
--- a/pkg/handlers/generic/mutation/mirrors/inject.go
+++ b/pkg/handlers/generic/mutation/mirrors/inject.go
@@ -11,6 +11,7 @@ import (
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	bootstrapv1 "sigs.k8s.io/cluster-api/bootstrap/kubeadm/api/v1beta1"
 	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
 	runtimehooksv1 "sigs.k8s.io/cluster-api/exp/runtime/hooks/api/v1alpha1"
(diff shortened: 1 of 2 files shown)


pkg/handlers/generic/mutation/mirrors/inject.go
```diff
@@ -234,6 +266,25 @@ func containerdConfigFromImageRegistry(
 	return configWithOptionalCACert, nil
 }
 
+func containerdConfigFromRegistryMirrorAddon(
+	_ context.Context,
+	_ ctrlclient.Client,
```
supershal, line 271:
```suggestion
```
nit: we can add it back when we need it in the function.

pkg/handlers/generic/mutation/mirrors/inject.go
```diff
@@ -82,14 +84,24 @@ func (h *globalMirrorPatchHandler) Mutate(
...
+	_, registryMirrorErr := variables.Get[v1alpha1.RegistryMirror](
+		vars,
+		v1alpha1.ClusterConfigVariableName,
+		[]string{"addons", v1alpha1.RegistryMirrorVariableName}...)
+
 	switch {
-	case variables.IsNotFoundError(imageRegistriesErr) && variables.IsNotFoundError(globalMirrorErr):
-		log.V(5).Info("Image Registry Credentials and Global Registry Mirror variable not defined")
+	case variables.IsNotFoundError(imageRegistriesErr) &&
+		variables.IsNotFoundError(globalMirrorErr) &&
```
supershal, line 94: what happens in the case where global registry is provided but the distribution registry is not? 
i.e
```
globalImageRegistry:
  url: <global_registry_url>
registries:
  url: <global_registry_url>
  secretRef: <secret>
```

It should not add the containerd config for the distribution registry. It looks like this patch will add it??
Can we add a negative unit test for this?

Also I feel it could be clean to write its own patch handler for the distribution registry.

dkoshkin: The client will not need to know about the internals for the distribution registry like the url to use, ie. Kubernetes Service IP for the registry, is being automatically appended here to the list of mirrors.
A client can still set their own mirror if they want, and the in-cluster distribution registry mirror will be used last.

In follow up PRs the CA and credentials will also be generated at runtime in which case the client would definitely not.

supershal: Thank you going through the code with me.

supershal (approved):

supershal (approved):
LGTM

pkg/handlers/generic/mutation/mirrors/inject.go
```diff
@@ -121,6 +133,22 @@ func (h *globalMirrorPatchHandler) Mutate(
...
 	}
+	if registryAddonErr == nil {
+		cluster, err := clusterGetter(ctx)
+		if err != nil {
+			log.Error(
+				err,
+				"failed to get cluster from Global Mirror mutation handler",
+			)
+			return err
+		}
```
manoj-nutanix, line 144: can we pass `Cluster` object to Mutate handler instead so we can avoid this additional call?

pkg/handlers/generic/mutation/mirrors/inject.go
```diff
@@ -234,6 +262,21 @@ func containerdConfigFromImageRegistry(
 	return configWithOptionalCACert, nil
 }
 
+func containerdConfigFromRegistryAddon(cluster *clusterv1.Cluster) (containerdConfig, error) {
```
manoj-nutanix, line 265: shall we add unit tests for this func?
Also, is there any unit test which verifies generated containerd file content?

manoj-nutanix (approved):
one q. if user removes registry config from cluster spec after deployment, will it generate new containerd patch without bootstrap registry config and apply it to live cluster?

pkg/handlers/generic/mutation/mirrors/inject.go
dkoshkin: Passing in `clusterGetter` is a pattern we've decided in this interface. We can consider the change but that would be out of scope for this PR, there might be some assumptions in existing handlers that it needs to operator on an updated Cluster.

pkg/handlers/generic/mutation/mirrors/inject.go
dkoshkin: >Also, is there any unit test which verifies generated containerd file content?

Yes those exist in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/main/pkg/handlers/generic/mutation/mirrors/containerd_files_test.go
From the perspective of those functions it won't know where this `containerdConfig` came from it just tests that the correct containerd file is generated based on the input it gets.

dkoshkin:
>one q. if user removes registry config from cluster spec after deployment, will it generate new containerd patch without bootstrap registry config and apply it to live cluster?

Thats correct, let me create a ticket, maybe we should have a validation webhook here to prevent this.

Not sure though because there are lots of other input we expect clients to understand what changing it means and nothing prevents it there.

pkg/handlers/generic/mutation/mirrors/inject.go
manoj-nutanix: cool, this is good enough for unit tests.

supershal (approved):

PR #1120, opened 2025-05-07
test(e2e): Disable Calico tests

jimmidyson:
These are currently unused and causing numerous flakes. Disabling until
we get time to properly investigate and fix.


dkoshkin (approved):
Calico is not a priority at this point, I'm ok with disabling it. Thanks

supershal (approved):

PR #1123, opened 2025-05-09
ci: use templating for capx tests

faiq:

uses templating for capx tests so we don't have to constantly keep updating images.

Fixes #


dlipovetsky (approved):
Thanks!

Does this need backports?

faiq:
> Does this need backports?

I don't think we can backport this -- it is only available in capx version 1.6+

pkg/handlers/nutanix/mutation/machinedetails/inject.go
```diff
@@ -100,10 +100,12 @@ func (h *nutanixMachineDetailsPatchHandler) Mutate(
 			spec.Cluster = nutanixMachineDetailsVar.Cluster
 
 			switch {
-			case nutanixMachineDetailsVar.Image != nil:
+			case nutanixMachineDetailsVar.Image != nil && nutanixMachineDetailsVar.ImageLookup == nil:
```
dkoshkin, line 103: Is this change necessary?
The API validation will prevent from both being set, but even if things change in the future, I think the more specific one`nutanixMachineDetailsVar.Image`, should take higher precedence than `nutanixMachineDetailsVar.ImageLookup`.

Right now if both are somehow set it returns `ErrNoImageOrImageLookupSet`.

examples/capi-quick-start/nutanix-cluster-cilium-helm-addon.yaml
```diff
@@ -104,9 +104,9 @@ spec:
               cluster:
                 name: ${NUTANIX_PRISM_ELEMENT_CLUSTER_NAME}
                 type: name
-              image:
-                name: ${NUTANIX_MACHINE_TEMPLATE_IMAGE_NAME}
-                type: name
+              imageLookup:
```
dkoshkin, line 107: Can `clusterctl` templating support if/else statements to still allow setting `NUTANIX_MACHINE_TEMPLATE_IMAGE_NAME` if set?

dkoshkin (commented):
Thanks for making this simpler, just a couple of questions.

pkg/handlers/nutanix/mutation/machinedetails/inject.go
faiq: i found that it was only after adding these changes the e2e tests passed.

> I think the more specific onenutanixMachineDetailsVar.Image, should take higher precedence than nutanixMachineDetailsVar.ImageLookup.

that is the existing logic and it wasn't changed.

examples/capi-quick-start/nutanix-cluster-cilium-helm-addon.yaml
faiq: i don't believe it can.

jimmidyson: No... it's just simple env var replacement really.

jimmidyson (approved):
Thank you!

.github/workflows/checks.yml
```diff
@@ -64,9 +64,10 @@ jobs:
     strategy:
       matrix:
         config:
-          - {"provider": "Nutanix", "kubernetesMinor": "v1.30", "kubernetesVersion": "v1.30.10", "osImage": "nkp-rocky-9.5-release-1.30.10-20250417212833"}
-          - {"provider": "Nutanix", "kubernetesMinor": "v1.31", "kubernetesVersion": "v1.31.4", "osImage": "nkp-rocky-9.5-release-1.31.4-20250409153435"}
-          - {"provider": "Nutanix", "kubernetesMinor": "v1.32", "kubernetesVersion": "v1.32.3", "osImage": "nkp-rocky-9.5-release-1.32.3-20250430150550"}
+          - {"provider": "Nutanix", "kubernetesMinor": "v1.30", "kubernetesVersion": "v1.30.10", "baseOS": "rocky-9.5"}
+          - {"provider": "Nutanix", "kubernetesMinor": "v1.31", "kubernetesVersion": "v1.31.4", "baseOS": "rocky-9.5"}
+          - {"provider": "Nutanix", "kubernetesMinor": "v1.32", "kubernetesVersion": "v1.32.3", "baseOS": "rocky-9.5"}
+          - {"provider": "Docker", "kubernetesMinor": "v1.30", "kubernetesVersion": "v1.30.12"}
```
jimmidyson, line 70: Oops missed this was re-added here, will remove in another PR.

PR #1124, opened 2025-05-09
refactor: Register types with scheme using pattern from Cluster API

dlipovetsky:
This pattern removes the dependency on controller-runtime from the API module. It is recommended by Cluster API.

https://cluster-api.sigs.k8s.io/developer/providers/getting-started/implement-api-types#registering-apis-in-the-scheme

Fixes #

Because the api module contains external types that do have a dependency on controller-runtime, we have to make a similar change to these types upstream before we can remove our dependency on controller-runtime.


jimmidyson (approved):
Thanks!

dkoshkin:
>This pattern removes the dependency on controller-runtime from the API module.

Should we expect a `go.mod` change here?

jimmidyson:
> Should we expect a `go.mod` change here?

From the PR description:

> Because the api module contains external types that do have a dependency on controller-runtime, we have to make a similar change to these types upstream before we can remove our dependency on controller-runtime.

dkoshkin (approved):

PR #1127, opened 2025-05-17
feat: generate a self-signed cert for registry addon

dkoshkin:
This PR adds the functionality to set a self-signed certificate for the registry addon.
A new `Certificate` deployed by the Helm chart generates a 10 year CA Secret `registry-addon-root-ca` in the namespace where this controller is running.

The `ca.crt` data is copied into a Cluster specific Secret (with an OwnerRef set) in a `BeforeClusterCreate` hook and is then used by the mirror handler.

This CA is then used to sign a new 2 year TLS certificate for every cluster that is copied to the remote cluster `registry-system/registry-tls` and used by the registry Pods. A new cert is generated and copied on every invocation of the handler, on `BeforeClusterUpgrade` and `AfterControlPlaneInitialized` lifecycle hooks, so that we can punt on auto renewal for now.

Fixes #

1. Tested manually by create a Docker cluster and verifying I can push an image into the registry, observe it get synced across the replicas, and then get pulled with Containerd.

```
$ kubectl get secret -n registry-system
NAME                                                TYPE                 DATA   AGE
cncf-distribution-registry-docker-registry-secret   Opaque               3      5h16m
registry-tls                                        kubernetes.io/tls    3      5h16m
sh.helm.release.v1.cncf-distribution-registry.v1    helm.sh/release.v1   1      5h16m


$ kubectl port-forward --address=127.0.0.1 --namespace registry-system pod/cncf-distribution-registry-docker-registry-0 5000:5000
$ crane copy nginx:stable 127.0.0.1:5000/library/nginx:dkoshkin --insecure
$ kubectl run nginx-working --image=docker.io/library/nginx:dkoshkin
$ kubectl get pods
NAME                                                              READY   STATUS              RESTARTS   AGE
cluster-autoscaler-0196db54-35b0-73fd-ad5b-14f998751820-7bzwnm9   0/1     ContainerCreating   0          5h15m
nginx-working                                                     1/1     Running             0          53m
```

2. New integration tests.
3. Existing e2e tests already wait for the STS Pods to be Ready which won't happen unless the TLS Secret is there and is valid.

I've considered other approaches here:
1. Using cert-manager to generate a unique CA per cluster. This makes it more difficult for clients to trust the registries when pushing images to multiple clusters. 
2. Use cert-manager to also generate the TLS certificate, it got pretty complicated to trigger a renewal and needing to wait for it to reconcile the cert Secret. Using cert-manager was also not providing much value as we want better control when the certs get rotated - https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/commit/118b39e1b3b7a6dba6f070e1e3cb511a70398318.
3. Use the CAPI generated CA cert to sign a new cert, but decided against using an externally managed CA for a separate usecase.


charts/cluster-api-runtime-extensions-nutanix/templates/certificates.yaml
```diff
@@ -32,3 +32,20 @@ spec:
     kind: {{ .Values.certificates.issuer.kind }}
     name: {{ template "chart.issuerName" . }}
   secretName: {{ template "chart.name" . }}-admission-tls
+---
+# CA used to sign certificates for the clusters' registry addons
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: registry-addon-tls
```
supershal, line 40:
```suggestion
  name: registry-addon-CA
```

pkg/handlers/generic/lifecycle/registry/utils/tls.go
```diff
@@ -0,0 +1,267 @@
...
+	return &corev1.Secret{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: corev1.SchemeGroupVersion.String(),
+			Kind:       "Secret",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      handlersutils.SecretNameForRegistryAddonCA(cluster),
+			Namespace: cluster.Namespace,
+		},
+		Data: data,
```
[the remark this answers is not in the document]
dkoshkin: The root CA will also have `tls.key` used to sign the CA, we don't want to expose the private key and just copying the ca.crt here.
Will add a comment.

pkg/handlers/generic/lifecycle/registry/handler.go
```diff
@@ -74,11 +92,97 @@ func (r *RegistryHandler) BeforeClusterUpgrade(
...
+	handler, ok := r.ProviderHandler[registryVar.Provider]
+	if !ok {
+		err = fmt.Errorf("unknown RegistryAddon Provider")
+		log.Error(err, "provider", registryVar.Provider)
+		resp.SetStatus(runtimehooksv1.ResponseStatusFailure)
+		resp.SetMessage(
+			fmt.Sprintf("%s %s", err, registryVar.Provider),
+		)
+		return
+	}
```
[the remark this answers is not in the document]
dkoshkin: I tried to extract it but it doesn't really work because of how errors work in these handlers where we just `return` and set the errors in the `resp` and log msgs.

pkg/handlers/generic/mutation/mirrors/inject.go
```diff
@@ -262,16 +267,27 @@ func containerdConfigFromImageRegistry(
...
+		return containerdConfig{}, errors.New("CA certificate not found in the secret")
+	}
 	config := containerdConfig{
-		// FIXME: Generate a self-signed CA.
-		URL:    fmt.Sprintf("http://%s", serviceIP),
-		Mirror: true,
+		URL:          fmt.Sprintf("https://%s", serviceIP),
+		Mirror:       true,
+		CASecretName: secret.Name,
+		CACert:       string(secret.Data[secretKeyForCACert]),
```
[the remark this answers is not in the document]
dkoshkin: Good idea.

pkg/handlers/generic/lifecycle/registry/cncfdistribution/handler.go
```diff
@@ -59,14 +64,71 @@ func New(
...
+	log.Info("Copying TLS Certificate for CNCF Distribution registry to remote cluster")
+	// Ensure the CA secret exists and is up to date in the cluster's namespace before trying to use it.
+	err := utils.EnsureCASecretForCluster(
+		ctx,
+		n.client,
+		cluster,
+	)
+	if err != nil {
+		return fmt.Errorf("failed to ensure CA secret for CNCF Distribution registry addon: %w", err)
+	}
```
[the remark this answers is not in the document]
manoj-nutanix: > makes sense. Could we call `setup()` in `AfterControlPlaneInitialized` handler too and remove `EnsureCASecretForCluster` from `apply()` .. are there any side effect if we do so?

I see it's already done.. thanks

pkg/handlers/generic/lifecycle/registry/cncfdistribution/handler.go
```diff
@@ -59,14 +64,64 @@ func New(
...
+			Name:      tlsSecretName,
+			Namespace: DefaultHelmReleaseNamespace,
+		},
+		Spec: utils.CertificateSpec{
+			CommonName:  stsName,
+			DNSNames:    certificateDNSNames(),
+			IPAddresses: certificateIPAddresses(serviceIP),
+		},
+	}
+	err = utils.EnsureTLSCertificateSecretOnRemoteCluster(
```
supershal, line 111: for separate PR: It would be good to provide a dev doc with diagram to demonstrate creation of global root CA, registry CA for a cluster and TLS certificates for a registry and how they will be copied, updated etc. 
Also good to document how client can push to this registry.

pkg/handlers/generic/lifecycle/registry/utils/tls.go
```diff
@@ -0,0 +1,271 @@
...
+) error {
+	globalTLSCertificateSecret, err := handlersutils.SecretForRegistryAddonRootCA(ctx, c)
+	if err != nil {
+		return err
+	}
+
+	clusterCASecret := buildClusterCASecret(globalTLSCertificateSecret, cluster)
+
+	// Copy the global CA certificate to a cluster CA secret.
+	err = handlersutils.EnsureSecretForLocalCluster(ctx, c, clusterCASecret, cluster)
```
supershal, line 73: can you confirm the cluster CA will stay on the management cluster and regenerated every time on upgrade?

dkoshkin: Yes we expect the CA Secret for the cluster to exist, this handler will fail and mirror mutation handler will also fail.
Initially I copied the CA every time but removed that https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1127/commits/9e22cd5ab45586013f765d27badd63e76f22901c#diff-d062d8df83a4dbee7f8632cbcb52dd12c887dad576ba3c5e9b4d1028e60953f7L91 since it was confusing why its in both places.

I think we can treat this the same like we do other cluster specific Secrets and just assume they are present.
Maybe using finalizers to prevent deletion but thats out of scope of this change IMO.

pkg/handlers/generic/lifecycle/registry/cncfdistribution/handler.go
dkoshkin: Great idea, added a new section on cert handling.

I don't yet know how we want the push to happen since this registry is not exposed publicly so left that out.

dkoshkin:
See certificate management:

```mermaid
sequenceDiagram
  participant CAREN Deployment as CAREN Deploymen
  participant BCC as BCC
  participant Topology Mutation as Topology Mutation
  participant ACPI as ACPI
  participant BCU as BCU
  participant Remote Cluster as Remote Cluster
  CAREN Deployment --> CAREN Deployment: Certificate: <caren-namespace>/registry-addon-root-ca
  CAREN Deployment --> CAREN Deployment: Secret: <caren-namespace>/registry-addon-root-ca
  BCC --> BCC: EnsureCASecretForCluster()
  BCC --> Topology Mutation: Secret: <cluster-namespace>/<cluster-name>-registry-addon-ca
  Topology Mutation --> Topology Mutation: containerdConfigFromRegistryAddon()
  ACPI --> ACPI: EnsureRegistryServerCertificateSecretOnRemoteCluster()
  ACPI ->> Remote Cluster: Secret: registry-system/registry-tls
  BCU --> BCU: EnsureRegistryServerCertificateSecretOnRemoteCluster()
  BCU ->> Remote Cluster: Secret: registry-system/registry-tls

```

PR #1129, opened 2025-05-20
feat: Add preflight checks framework

dlipovetsky:
Adds a framework for preflight checks.

A validating webhook on the Cluster resource executes all preflight checks, and returns failures, and warnings to the client. 

Fixes #

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -41,6 +41,7 @@ import (
 	"github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pkg/handlers/nutanix"
 	"github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pkg/handlers/options"
 	"github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pkg/webhook/cluster"
+	"github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pkg/webhook/preflight"
 )
 
 func main() {
@@ -219,6 +220,13 @@ func main() {
 		Handler: cluster.NewValidator(mgr.GetClient(), admission.NewDecoder(mgr.GetScheme())),
 	})
 
+	mgr.GetWebhookServer().Register("/preflight-v1beta1", &webhook.Admission{
+		Handler: preflight.New(mgr.GetClient(), admission.NewDecoder(mgr.GetScheme()),
+			[]preflight.Checker{
+				// Add your preflight checkers here.
+			}...,
+		),
+	})
 	if err := mgr.Start(signalCtx); err != nil {
 		setupLog.Error(err, "unable to start controller manager")
 		os.Exit(1)
--- a/pkg/webhook/preflight/preflight.go
+++ b/pkg/webhook/preflight/preflight.go
@@ -1,0 +1,133 @@
+// Copyright 2025 Nutanix. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+package preflight
+
+import (
+	"context"
+	"net/http"
+
+	admissionv1 "k8s.io/api/admission/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	"github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/common/pkg/capi/utils"
+)
+
+type CheckResult struct {
+	Allowed bool
+	Field   string
+	Message string
+	Warning string
+	Error   bool
+}
+
+type Check = func(ctx context.Context) CheckResult
+
+type Checker interface {
+	Checks(ctx context.Context, client ctrlclient.Client, cluster *clusterv1.Cluster) ([]Check, error)
+	Provider() string
+}
+
+type WebhookHandler struct {
+	client             ctrlclient.Client
+	decoder            admission.Decoder
+	checkersByProvider map[string]Checker
+}
+
+func New(client ctrlclient.Client, decoder admission.Decoder, checkers ...Checker) *WebhookHandler {
+	h := &WebhookHandler{
+		client:             client,
+		decoder:            decoder,
+		checkersByProvider: make(map[string]Checker, len(checkers)),
+	}
+	for _, checker := range checkers {
+		h.checkersByProvider[checker.Provider()] = checker
+	}
+	return h
+}
+
+func (h *WebhookHandler) Handle(ctx context.Context, req admission.Request) admission.Response {
+	if req.Operation == admissionv1.Delete {
+		return admission.Allowed("")
+	}
+
+	cluster := &clusterv1.Cluster{}
+	err := h.decoder.Decode(req, cluster)
+	if err != nil {
+		return admission.Errored(http.StatusBadRequest, err)
+	}
+
+	// Checks run only for ClusterClass-based clusters.
+	if cluster.Spec.Topology == nil {
+		return admission.Allowed("")
+	}
+
+	// Checks run only for the known infrastructure providers.
+	checker, ok := h.checkersByProvider[utils.GetProvider(cluster)]
+	if !ok {
+		return admission.Allowed("")
+	}
+
+	resp := admission.Response{
+		AdmissionResponse: admissionv1.AdmissionResponse{
+			Allowed: true,
+			Result: &metav1.Status{
+				Details: &metav1.StatusDetails{},
+			},
+		},
+	}
+
+	checks, err := checker.Checks(ctx, h.client, cluster)
+	if err != nil {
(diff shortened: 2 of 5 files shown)


pkg/webhook/preflight/preflight.go
```diff
@@ -0,0 +1,156 @@
...
+			result := check(ctx)
+			resultCh <- result
+			wg.Done()
+		}(ctx, check)
+	}
+	wg.Wait()
+	close(resultCh)
+
+	// Collect check results.
+	for result := range resultCh {
```
dkoshkin, line 126: Very clean, on any Error to run a Check or when a check returns a `Allowed==false` the `resp` will always be set to `resp.Allowed = false` and will keep appending to the causes.

(just mainly thinking out loud)

dkoshkin (approved):
Thanks for keeping the PR small and targeted!

cmd/main.go
```diff
@@ -219,6 +220,13 @@ func main() {
 		Handler: cluster.NewValidator(mgr.GetClient(), admission.NewDecoder(mgr.GetScheme())),
 	})
 
+	mgr.GetWebhookServer().Register("/preflight-v1beta1", &webhook.Admission{
```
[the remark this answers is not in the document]
dlipovetsky: Good call. We're only registering it for the Cluster. If we end up registering it for other resources, we can change the name then.

dlipovetsky:
@dkoshkin Thanks for reviewing!

As I'm working on #1130, I'm making some changes here.

In hindsight, I should have marked this a draft PR. Sorry that I didn't. I won't force-push any changes here. I hope that will make it easier to review the new changes. :pray:

cmd/main.go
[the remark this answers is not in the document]
dlipovetsky: Fixed in 4a518b3100824ca88b33584cabc8b71ac063a8da

pkg/webhook/preflight/preflight.go
dlipovetsky: That's right. We want to keep track of whether we had a rejection, or an error, but collect all the check results, regardless.

pkg/webhook/preflight/preflight.go
```diff
@@ -0,0 +1,153 @@
...
+	// Because we have some causes, we construct the response message and code.
+	resp.Result.Message = "preflight checks failed"
+	resp.Result.Code = http.StatusForbidden
+	resp.Result.Reason = metav1.StatusReasonForbidden
+	if internalError {
+		// Internal errors take precedence over check failures.
+		resp.Result.Code = http.StatusInternalServerError
+		resp.Result.Reason = metav1.StatusReasonInternalError
+	}
+
```
jimmidyson, line 151: I don't checking if causes is empty is enough here, does it handle the error response correctly? How about a switch instead to cover error and allowed state? Also think it should use `StatusReasonInvalid` rather than `StatusReasonForbidden`.

```suggestion
	switch {
		case internalError:
			// Internal errors take precedence over check failures.
			resp.Result.Code = http.StatusInternalServerError
			resp.Result.Reason = metav1.StatusReasonInternalError
		case !resp.Allowed:
			// Because the response is not allowed, preflights must have failed.
			resp.Result.Message = "preflight checks failed"
			resp.Result.Code = http.StatusInvalid
			resp.Result.Reason = metav1.StatusReasonInvalid
	}

```

pkg/webhook/preflight/preflight.go
```diff
@@ -0,0 +1,153 @@
...
+			checkerWG.Done()
+		}(ctx, checker, resultCh)
+	}
+
+	// Close the channel when all checkers are done.
+	go func(wg *sync.WaitGroup, resultCh chan CheckResult) {
+		wg.Wait()
+		close(resultCh)
+	}(checkerWG, resultCh)
```
faiq, line 114: can the function return before this go routing closes the channel?

what happens with slow checks?

pkg/webhook/preflight/preflight.go
```diff
@@ -0,0 +1,153 @@
...
+			internalError = true
+		}
+
+		if !result.Allowed {
+			resp.Allowed = false
+		}
+
+		for _, cause := range result.Causes {
+			resp.Result.Details.Causes = append(resp.Result.Details.Causes, metav1.StatusCause{
+				Type:    metav1.CauseType(fmt.Sprintf("FailedPreflight%s", result.Name)),
```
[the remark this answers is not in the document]
dlipovetsky: What concerns do you have? Do you want to impose a limit?

There is no documented limit to the number of causes that can be returned, and I find nothing in the API server's [webhook dispatcher code](https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/validating), either.

pkg/webhook/preflight/preflight.go
```diff
@@ -0,0 +1,153 @@
...
+					result := check(ctx)
+					resultCh <- result
+					checksWG.Done()
+				}(ctx, check, resultCh)
+			}
+			checksWG.Wait()
+
+			checkerWG.Done()
+		}(ctx, checker, resultCh)
+	}
```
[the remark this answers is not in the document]
dlipovetsky: Because I need to initialize checkers concurrently, I end up needing to order by checker index _and_ check index.

One priority queue won't work. I considered using one priority queue for checkers, and another for checks. I then decided to trade off space for complexity, and stored results in a two-dimensional slice, shared among all the goroutines. WDYT?

pkg/webhook/preflight/preflight.go
dlipovetsky: I replaced this code when I switched approaches, but your question is still valid. And it's a good question :smile: 

> still would like to know how slow checks end up getting propogated up to the user!

The short answer is: they don't.

The webhook configuration has a configurable timeout, up to 30 seconds. The API server abandons the request when the time is up. No results are returned at all, and the user merely sees a "context deadline exceeded" error.

We want to return the results of some checks to the user, even if we can't complete all the checks in time. To do that, the framework will need to (a) impose its own timeout (and it must be shorter than the one configured for the webhook), (b) pre-empt checks that fail to return within the timeout.

dlipovetsky: I think it's important to return partial results instead of simply timing out. I'll make sure we address this in a future PR. How does that sound?

dlipovetsky: I filed https://jira.nutanix.com/browse/NCN-107885

pkg/webhook/preflight/preflight.go
```diff
@@ -0,0 +1,162 @@
...
+		resp.Result.Code = http.StatusUnprocessableEntity
+		resp.Result.Reason = metav1.StatusReasonInvalid
+	}
+
+	return resp
+}
+
+// run runs all checks for the cluster, concurrently, and returns the results ordered by checker and check.
+// Checker are initialized concurrently, and checks runs concurrently as well.
+func run(ctx context.Context,
```
dlipovetsky, line 122: From @thunderboltsid: Let's make sure we recover from a panic in a check.

pkg/webhook/preflight/preflight.go
```diff
@@ -0,0 +1,162 @@
...
+
+type (
+	// CheckerFactory returns a Checker for a given cluster.
+	CheckerFactory func(client ctrlclient.Client, cluster *clusterv1.Cluster) Checker
+	Checker        interface {
+		// Init returns the checks that should run for the cluster.
+		Init(ctx context.Context) []Check
+	}
+
+	Check       = func(ctx context.Context) CheckResult
```
dlipovetsky, line 26: It would help to know the name of the Check, even if it fails to run (e.g. it panics). Maybe this should be an interface with two methods, `Run` and `Name`?

dlipovetsky: If a check panics, the client may want to skip it, and doing that requires knowing its name.

thunderboltsid (approved):

pkg/webhook/preflight/preflight.go
```diff
@@ -0,0 +1,210 @@
...
+		for _, result := range results {
+			if result.Error {
+				internalError = true
+			}
+			if !result.Allowed {
+				resp.Allowed = false
+			}
+			for _, cause := range result.Causes {
+				resp.Result.Details.Causes = append(resp.Result.Details.Causes, metav1.StatusCause{
+					Type:    metav1.CauseType(fmt.Sprintf("FailedPreflight%s", result.Name)),
```
yanhua121, line 118: I think it is better to predefine the possible preflight causes as metav1.CauseType constants, see those examples in the metav1 package, such as:

const (
	// CauseTypeFieldValueNotFound is used to report failure to find a requested value
	// (e.g. looking up an ID).
	CauseTypeFieldValueNotFound CauseType = "FieldValueNotFound"
        .........
}

dlipovetsky: I definitely considered this, but we need a field to hold the check name, so that clients can identiffy the check, for example, to offer the option of skipping the check.

We have three available fields; `Message` to hold the explanation, `Field` to reference the configuration, e.g., the field that needs to be changed to make the check pass, and finally, `Type` to hold the check name.

PR #1132, opened 2025-05-21
Add hardened example

WinsonSou:
**What problem does this PR solve?**:
CIS Level 1 + 2 Benchmarks for Control Planes and Workers
**Which issue(s) this PR fixes**:
N/A

**How Has This Been Tested?**:
Create NKP 2.14.0 Cluster
Clone and Patch kubeadmcontrolplanetemplate and kubeadmconfigtemplate
Patch ClusterClass to reference new kubeadmcontrolplanetemplate and kubeadmconfigtemplate
Perform Tenable CIS Benchmark Scan
Upgrade Cluster to NKP 2.15.0
Clone and Patch upgraded kubeadmcontrolplanetemplate and kubeadmconfigtemplate
Patch Upgraded ClusterClass to reference new kubeadmcontrolplanetemplate and kubeadmconfigtemplate
Perform Tenable CIS Benchmark Scan

PR #1137, opened 2025-05-29
feat: CIS benchmark improvements

jimmidyson:
Pulling in most of the changes from #1132.

This commit does not include Kubelet CSR Approver because that requires
more invasive changes (e.g. deploying the `kubelet-csr-approver` service
and reconfiguring kubelet after initial startup.


charts/cluster-api-runtime-extensions-nutanix/defaultclusterclasses/nutanix-cluster-class.yaml
```diff
@@ -200,18 +223,45 @@ spec:
...
             kubeletExtraArgs:
               cloud-provider: external
+              event-qps: "5"
               eviction-hard: nodefs.available<10%,nodefs.inodesFree<5%,imagefs.available<15%,memory.available<100Mi,imagefs.inodesFree<10%
-              tls-cipher-suites: TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256,TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256
+              make-iptables-util-chains: "true"
+              pod-max-pids: "4096"
+              read-only-port: "0"
+              streaming-connection-idle-timeout: 5m
+              tls-cipher-suites: TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_RSA_WITH_AES_256_GCM_SHA384,TLS_RSA_WITH_AES_128_GCM_SHA256
```
dkoshkin, line 253: FYI, these 2 are newL
```
TLS_RSA_WITH_AES_256_GCM_SHA384
TLS_RSA_WITH_AES_128_GCM_SHA256
```

And this one was removed:
```
TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256
```

Any concerns about that?

charts/cluster-api-runtime-extensions-nutanix/defaultclusterclasses/nutanix-cluster-class.yaml
```diff
@@ -123,13 +131,28 @@ spec:
         clusterConfiguration:
           apiServer:
             extraArgs:
+              admission-control-config-file: /etc/kubernetes/admission.yaml
               cloud-provider: external
+              enable-admission-plugins: DenyServiceExternalIPs,EventRateLimit,NodeRestriction
```
dkoshkin, line 136: https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#denyserviceexternalips
https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#noderestriction

I think both of these may be considered a breaking change. At minimum we will need to document this, but do we also want to expose this in the API?

pkg/handlers/generic/mutation/auditpolicy/inject.go
```diff
@@ -25,7 +26,7 @@ type auditPolicyPatchHandler struct{}
 //go:embed embedded/apiserver-audit-policy.yaml
 var auditPolicy string
 
-const auditPolicyPath = "/etc/kubernetes/audit-policy/apiserver-audit-policy.yaml"
+const auditPolicyPath = "/etc/kubernetes/audit-policy.yaml"
```
dkoshkin, line 29: This will cause a rollout on all existing clusters when CAREN version is upgraded.
We will need to bump the handler version and copy the old implementation to https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/tree/main/pkg/handlers/v2 (but changed to v3 and removed the kube-vip handler from there)

(can be done as a separate PR)

dkoshkin (approved):
Thanks for the changes here, please follow up on my comment for a new handler version.

pkg/handlers/generic/mutation/auditpolicy/inject.go
jimmidyson: Good catch! I'll open a follow up PR before we merge this one.

charts/cluster-api-runtime-extensions-nutanix/defaultclusterclasses/aws-cluster-class.yaml
```diff
@@ -84,25 +84,74 @@ spec:
...
             kubeletExtraArgs:
               cloud-provider: external
+              event-qps: "5"
+              make-iptables-util-chains: "true"
+              pod-max-pids: "4096"
+              read-only-port: "0"
+              streaming-connection-idle-timeout: 5m
+              tls-cipher-suites: TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_RSA_WITH_AES_256_GCM_SHA384,TLS_RSA_WITH_AES_128_GCM_SHA256
             name: '{{ ds.meta_data.local_hostname }}'
+      postKubeadmCommands:
```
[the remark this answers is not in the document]
jimmidyson: Oops! Thanks for catching this!

charts/cluster-api-runtime-extensions-nutanix/defaultclusterclasses/nutanix-cluster-class.yaml
jimmidyson: Good catch!

While the 2 new ciphers are actually recommended by CIS benchmark (even in CIS latest 1.11...) they are indeed considered insecure so I think we are correct in previously omitting these so will continue to do so going forward.

Conversely, `TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256` is not included in CIS 1.11 benchmark recommendation so removed just in case this has to be justified...

charts/cluster-api-runtime-extensions-nutanix/defaultclusterclasses/nutanix-cluster-class.yaml
jimmidyson: Technicall I agree...

`NodeRestriction` shouldn't have a direct impact on users.

From [the docs](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#denyserviceexternalips), enabling `DenyServiceExternalIPs` will not break existing clusters or services but would prevent creating new ones with `externalIPs`:

```
When enabled, users of the cluster may not create new Services which use externalIPs and may not add new values to externalIPs on existing Service objects. Existing uses of externalIPs are not affected, and users may remove values from externalIPs on existing Service objects.

Most users do not need this feature at all, and cluster admins should consider disabling it. Clusters that do need to use this feature should consider using some custom policy to manage usage of it.
```

winsonsou-nutanix (approved):
LGTM

charts/cluster-api-runtime-extensions-nutanix/defaultclusterclasses/aws-cluster-class.yaml
```diff
@@ -84,24 +84,73 @@ spec:
...
               profiling: "false"
+        files:
+        - content: |
+            apiVersion: apiserver.config.k8s.io/v1
+            kind: AdmissionConfiguration
+            plugins:
+            - name: EventRateLimit
+              path: /etc/kubernetes/eventratelimit-config.yaml
+          path: /etc/kubernetes/admission.yaml
+          permissions: "0600"
```
thunderboltsid, line 118: nit: indentation

charts/cluster-api-runtime-extensions-nutanix/defaultclusterclasses/docker-cluster-class.yaml
```diff
@@ -83,17 +83,68 @@ spec:
...
               profiling: "false"
+        files:
+        - content: |
+            apiVersion: apiserver.config.k8s.io/v1
+            kind: AdmissionConfiguration
+            plugins:
+            - name: EventRateLimit
+              path: /etc/kubernetes/eventratelimit-config.yaml
+          path: /etc/kubernetes/admission.yaml
+          permissions: "0600"
```
thunderboltsid, line 116: nit: indentation

charts/cluster-api-runtime-extensions-nutanix/defaultclusterclasses/docker-cluster-class.yaml
```diff
@@ -83,17 +83,68 @@ spec:
...
+        - content: |
+            apiVersion: eventratelimit.admission.k8s.io/v1alpha1
+            kind: Configuration
+            limits:
+            - type: Server
+              qps: 10000
+              burst: 40000
+          path: /etc/kubernetes/eventratelimit-config.yaml
+          permissions: "0600"
```
thunderboltsid, line 125: nit: indentation

charts/cluster-api-runtime-extensions-nutanix/defaultclusterclasses/aws-cluster-class.yaml
jimmidyson: This is indented correctly, sibling fields to content... bad highlighting in GitHub! Kustomize writes out property keys alphabetically so it looks a bit odd, but is correct.

dkoshkin (approved):

charts/cluster-api-runtime-extensions-nutanix/defaultclusterclasses/aws-cluster-class.yaml
```diff
@@ -84,24 +84,73 @@ spec:
...
+        - content: |
+            apiVersion: eventratelimit.admission.k8s.io/v1alpha1
+            kind: Configuration
+            limits:
+            - type: Server
+              qps: 10000
+              burst: 40000
+          path: /etc/kubernetes/eventratelimit-config.yaml
+          permissions: "0600"
```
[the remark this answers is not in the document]
jimmidyson: See https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1137#discussion_r2121683213

charts/cluster-api-runtime-extensions-nutanix/defaultclusterclasses/docker-cluster-class.yaml
jimmidyson: See https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1137#discussion_r2121683213

charts/cluster-api-runtime-extensions-nutanix/defaultclusterclasses/docker-cluster-class.yaml
jimmidyson: See https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1137#discussion_r2121683213

thunderboltsid (approved):

PR #1154, opened 2025-06-09
feat: Enable kube-vip CP load-balancer in masquerade mode

jimmidyson:
This enables actual load balancing across CP nodes, an improvement on the
current configuration that only routes traffic to the API server running on
the CP node where kube-vip leader is running.

With this new configuration, the ipvs service is created on the node where the
current leader is running and will be created on new leader when the VIP moves.
This means there will be minimal downtime when the leader is reelected, which
usually happens only when a node is rolled out (e.g. for CP upgrade). This is the
same behaviour as previously with no ipvs, just adding the details here
to align on expected behaviour.

### DIFF
--- a/charts/cluster-api-runtime-extensions-nutanix/defaultclusterclasses/nutanix-cluster-class.yaml
+++ b/charts/cluster-api-runtime-extensions-nutanix/defaultclusterclasses/nutanix-cluster-class.yaml
@@ -194,20 +194,21 @@ spec:
                       value: "10"
                     - name: vip_retryperiod
                       value: "2"
+                    - name: lb_enable
+                      value: "true"
+                    - name: lb_port
+                      value: "6443"
+                    - name: lb_fwdmethod
+                      value: masquerade
                     - name: address
                       value: '{{ .Address }}'
                     - name: prometheus_server
-                  image: ghcr.io/kube-vip/kube-vip:v0.9.1
+                  image: ghcr.io/kube-vip/kube-vip-iptables:v0.9.1
                   imagePullPolicy: IfNotPresent
                   name: kube-vip
                   resources: {}
                   securityContext:
-                    capabilities:
-                      add:
-                        - NET_ADMIN
-                        - NET_RAW
-                      drop:
-                        - ALL
+                    privileged: true
                   volumeMounts:
                     - mountPath: /etc/kubernetes/admin.conf
                       name: kubeconfig
--- a/hack/addons/update-kube-vip-manifests.sh
+++ b/hack/addons/update-kube-vip-manifests.sh
@@ -22,17 +22,20 @@ docker container run --rm ghcr.io/kube-vip/kube-vip:"${KUBE_VIP_VERSION}" \
   --arp \
   --address='127.0.0.1' \
   --controlplane \
+  --enableLoadBalancer \
+  --lbForwardingMethod=masquerade \
   --leaderElection \
   --leaseDuration=15 \
   --leaseRenewDuration=10 \
   --leaseRetry=2 \
   --prometheusHTTPServer='' |
   gojq --yaml-input --yaml-output \
-    'del(.metadata.creationTimestamp, .status) |
-     .spec.containers[].imagePullPolicy |= "IfNotPresent" |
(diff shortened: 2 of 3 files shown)


charts/cluster-api-runtime-extensions-nutanix/defaultclusterclasses/nutanix-cluster-class.yaml
```diff
@@ -194,20 +194,21 @@ spec:
...
+                      value: "true"
+                    - name: lb_port
+                      value: "6443"
+                    - name: lb_fwdmethod
+                      value: masquerade
                     - name: address
                       value: '{{ .Address }}'
                     - name: prometheus_server
-                  image: ghcr.io/kube-vip/kube-vip:v0.9.1
+                  image: ghcr.io/kube-vip/kube-vip-iptables:v0.9.1
```
dkoshkin, line 206: Do you know what the difference is between this image and the regular kube-vip image?

jimmidyson: The image includes `iptables` and as such is based on `alpine` rather than `scratch`. The difference between these images is the following rather than `FROM scratch`:

```
FROM alpine:3.21.3
# Update pkgs and add iptables
RUN apk upgrade && \
    apk add --no-cache iptables iptables-legacy
```

dkoshkin:
Any concerns here during upgrades as CPs are being rolled out and some Pods have this enabled when others don't?

jimmidyson:
Need to test that out, but I don't think so. The ipvs service will only be created on the elected kube vip leader once a new CP becomes kube-vip leader. When that happens it will load balance across all discovered CP nodes, whether they are old or new. Either way it should just work.

charts/cluster-api-runtime-extensions-nutanix/defaultclusterclasses/nutanix-cluster-class.yaml
```diff
@@ -194,20 +194,21 @@ spec:
                       value: "10"
                     - name: vip_retryperiod
                       value: "2"
+                    - name: lb_enable
+                      value: "true"
+                    - name: lb_port
+                      value: "6443"
```
dkoshkin, line 200: For future us, this `lb_port` is the target port, ie kube-apiserver which in our case is always running on port 6443.
The other config here `port: {{ .Port }}` is where the VIP will be listening on and can be configured in the API.

Enabling IPVS LB here also fixes the bug where changing the API Server port did not actually work.

thunderboltsid (approved):

charts/cluster-api-runtime-extensions-nutanix/defaultclusterclasses/nutanix-cluster-class.yaml
dkoshkin: It does not work, I set `controlPlaneEndpoint.port=443`

This causes kube-vip to not even start up:
```
E0624 18:03:03.693671       1 leaderelection.go:436] error retrieving resource lock kube-system/plndr-cp-lock: Get "https://kubernetes:443/apis/coordination.k8s.io/v1/namespaces/kube-system/leases/plndr-cp-lock?timeout=10s": dial tcp 127.0.0.1:443: connect: connection refused
```

Switching kube-vip's `port` and `lb_port`, it does not respond on 443. It does respond on 6443 but at this point the kubeconfig is already referencing 6443.
```
[root@dk-kube-vip-5jkk7-5rqmb manifests]# curl -k https://10.22.203.1:443
curl: (7) Failed to connect to 10.22.203.1 port 443: Connection refused
[root@dk-kube-vip-5jkk7-5rqmb manifests]# curl -k https://10.22.203.1:6443
{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {},
  "status": "Failure",
  "message": "forbidden: User \"system:anonymous\" cannot get path \"/\"",
  "reason": "Forbidden",
  "details": {},
  "code": 403
}
```

dkoshkin:
@jimmidyson curious why you closed this? Do you think this is not the way to go?

PR #1167, opened 2025-06-17
refactor: add global feature.Gate variable

dkoshkin:
Refactoring the feature gate code to make it easier to use.
Added a new `feature.Gate` readonly variable that can be used anywhere in the codebase to determine if a specific feature gate was set.

This is based on CAPI's implementation https://github.com/kubernetes-sigs/cluster-api/tree/main/feature

Fixes #


pkg/feature/feature.go
```diff
@@ -1,13 +1,15 @@
 // Copyright 2024 Nutanix. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-package features
+package feature
```
dkoshkin, line 4: Renamed the package to make it more readable with `feature.Gates`

thunderboltsid (approved):

PR #1169, opened 2025-06-18
feat: Enforce MD replicas within cluster autoscaler bounds

jimmidyson:
When the current MD `.spec.replicas` is outside of the bounds of the cluster autoscaler min/max annotations, cluster autoscaler will not scale the MD up or down. This can occur when either the annotations are edited or `spec.replicas` is updated.

In a topology-based cluster, CAPI propagates the cluster autoscaler annotations from `cluster.spec.topology.workers.machineDeployments[].metadata.annotations` to the templated `MachineDeployment` resources. If these annotations are changed so that the range they define does not contain the current number of replicas, then no scaling up or down will happen to the MachineDeployment by the cluster autoscaler.

This commit adds a controller that does the following to MachineDeployments:

- if `spec.replicas` is nil do nothing
- if cluster autoscaler annotations for min or max are missing do nothing
- if `spec.replicas` is within cluster autoscaler annotations bounds do nothing
- if `spec.replicas` is not within CA annotations bounds then set `spec.replicas`
  to nil (CAPI MD defaulting webhook will correctly set MD replicas as per
https://github.com/kubernetes-sigs/cluster-api/blob/v1.10.3/internal/webhooks/machinedeployment.go#L353-L376.

This ensures that scaling up and down by specifying cluster autoscaler annotations on the machine deployments in `cluster.spec.topology.workers.machineDeployments[].metadata.annotations` will always be respected, even if the current number of replicas is outside the autoscaling specified bounds.

### DIFF
--- a/charts/cluster-api-runtime-extensions-nutanix/README.md
+++ b/charts/cluster-api-runtime-extensions-nutanix/README.md
@@ -30,6 +30,7 @@ A Helm chart for cluster-api-runtime-extensions-nutanix
 | certificates.issuer.selfSigned | bool | `true` |  |
 | deployDefaultClusterClasses | bool | `true` |  |
 | deployment.replicas | int | `1` |  |
+| enforceClusterAutoscalerLimits.enabled | bool | `true` |  |
 | env | object | `{}` |  |
 | helmAddonsConfigMap | string | `"default-helm-addons-config"` |  |
 | helmRepository.enabled | bool | `true` |  |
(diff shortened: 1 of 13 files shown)


dkoshkin (approved):
Thanks! Tested locally too in a docker cluster

dlipovetsky:
@jimmidyson also reminded me that, if the CA annotations are set in `spec.topology.workers.machineDeployments[].metadata.annotations`, then `spec.topology.workers.machineDeployments[].replicas` is explicitly [blocked by the Cluster API Cluster webhook](https://github.com/kubernetes-sigs/cluster-api/blob/ce4c050cce7cec8194c1cc449627e47fb57b0123/internal/webhooks/cluster.go#L1007-L1009).

This means that Cluster API core is "aware" of these annotations, and though they are not fields, they are effectively part of the API spec. Therefore, I think the behavior proposed in this PR should be implemented in core Cluster API.

But we can first implement it here.

jimmidyson:
> I think the behavior proposed in this PR should be implemented in core Cluster API.

I think it should be in cluster-autoscaler, enforce max the equivalent of the existing enforce min there.

PR #1175, opened 2025-06-23
feat: auto enable registry addon in workload clusters

dkoshkin:
This adds a new feature gate `AutoEnableWorkloadClusterRegistry=true` that will automatically enable the registry addon on workload cluster creation when it is enabled on the management cluster.

Fixes #


pkg/webhook/addons/registry/webhook_suite_test.go
```diff
@@ -0,0 +1,71 @@
...
+		M: m,
+		WebhookInstallOptions: ctrlenvtest.WebhookInstallOptions{
+			MutatingWebhooks: []*admissionv1.MutatingWebhookConfiguration{
+				mutatingWebhook,
+			},
+		},
+		SetupEnv: func(e *envtest.Environment) {
+			e.GetWebhookServer().Register("/mutate-v1beta1-registry-addon", &webhook.Admission{
+				Handler: NewWorkloadClusterAutoEnabler(e.GetClient(), admission.NewDecoder(e.GetScheme())).Defaulter(),
+			})
```
supershal, line 66: nice. good to learn how to set it up in test.

pkg/webhook/addons/registry/autoenabler.go
```diff
@@ -0,0 +1,185 @@
...
+	if err != nil {
+		return admission.Errored(
+			http.StatusInternalServerError,
+			fmt.Errorf(
+				"failed to get management cluster: %w",
+				err,
+			))
+	}
+	// Check if creating a workload cluster, ie managementCluster is not nil. If it is nil just return.
+	if managementCluster == nil {
```
supershal, line 95: nit: we can move this check before we start inspecting topology variables.

pkg/webhook/addons/registry/autoenabler.go
```diff
@@ -0,0 +1,185 @@
...
+	clusterRegistry, err := registryAddonFromCluster(cluster)
+	if err != nil {
+		return admission.Errored(
+			http.StatusInternalServerError,
+			fmt.Errorf(
+				"failed to check if registry addon is enabled in cluster: %w",
+				err,
+			))
+	}
+	if clusterRegistry != nil {
```
supershal, line 67: not directly related to this default webhook. So if the management cluster does not have registry addon enabled, but workload cluster has it set, we will deploy in-cluster registry on the workload cluster. If the other flag to sync images is not enabled then this registry on workload cluster will not be used. 
do we want to allow such usecase? 
Or separately we should block cluster creation in above case?

dkoshkin: >If the other flag to sync images is not enabled then this registry on workload cluster will not be used.

The other webhook will also check if its enabled on the mgmt cluster and only set if it is.

> do we want to allow such usecase?

Yes I think we should, its possible to seed the registry in other ways.

pkg/webhook/addons/registry/webhook_suite_test.go
dkoshkin: Yeah shamelessly copied from another webhook in CAREN 😆

supershal (approved):

thunderboltsid (approved):

PR #1180, opened 2025-06-24
fix(preflight): improved error reporting for storage container and VM image checks

thunderboltsid:
- Handle nil responses of get storage container and get cluster API calls.
- Improve error wording. 
- Correctly construct error responses of list API calls.
- Test error response of list clusters API call.
- Treat a missing storageContainer as a failed check, but not an internal error.


dlipovetsky:
Should this handling of the List call be similar to what we have in `getVMImages`? https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/35fd1698256ef86f38eb11bff40ff6a1b3a1ba02/pkg/webhook/preflight/nutanix/image.go#L125-L137

PR #1203, opened 2025-07-08
fix: allow http registries by disabling TLS check

faiq:
Disables TLS if the registry is HTTP 

Fixes #

### DIFF
--- a/pkg/webhook/preflight/generic/registry.go
+++ b/pkg/webhook/preflight/generic/registry.go
@@ -93,8 +93,11 @@ func (r *registryCheck) checkRegistry(
 		result.Allowed = false
 		result.Causes = append(result.Causes,
 			preflight.Cause{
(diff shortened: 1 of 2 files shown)


pkg/webhook/preflight/generic/registry.go
```diff
@@ -86,9 +86,23 @@ func (r *registryCheck) checkRegistry(
...
 	}
-	mirrorHost := config.Host{
+	registryHost := config.Host{
 		Name: registryURLParsed.Host,
 	}
+	if registryURLParsed.Scheme != "http" && registryURLParsed.Scheme != "https" {
+		result.Allowed = false
+		result.Causes = append(result.Causes,
+			preflight.Cause{
+				Message: fmt.Sprintf("failed to parse registry url must be http or https %s", registryURL),
```
jimmidyson, line 96:
```suggestion
				Message: fmt.Sprintf(`incorrect scheme for registry URL %q: must be either "http" or "https"`, registryURL),
```

PR #1217, opened 2025-07-17
NCN-108428: When failureDomain is configured, machineDetails "cluster" and "subnets" should be optional

yanhua121:
https://jira.nutanix.com/browse/NCN-108428:
When failureDomain is configured, machineDetails "cluster" and "subnets" should be optional

### DIFF
--- a/pkg/handlers/nutanix/mutation/machinedetails/inject.go
+++ b/pkg/handlers/nutanix/mutation/machinedetails/inject.go
@@ -97,7 +97,14 @@ func (h *nutanixMachineDetailsPatchHandler) Mutate(
 			spec := obj.Spec.Template.Spec
 
 			spec.BootType = nutanixMachineDetailsVar.BootType
-			spec.Cluster = nutanixMachineDetailsVar.Cluster
+			if nutanixMachineDetailsVar.Cluster != nil {
+				spec.Cluster = *nutanixMachineDetailsVar.Cluster
+			} else {
+				// Have to set the required Type field.
+				spec.Cluster = capxv1.NutanixResourceIdentifier{
+					Type: capxv1.NutanixIdentifierName,
+				}
+			}
 
 			switch {
 			case nutanixMachineDetailsVar.Image != nil:
@@ -118,7 +125,6 @@ func (h *nutanixMachineDetailsPatchHandler) Mutate(
 			spec.Subnets = slices.Clone(nutanixMachineDetailsVar.Subnets)
 			spec.AdditionalCategories = slices.Clone(nutanixMachineDetailsVar.AdditionalCategories)
 			spec.GPUs = slices.Clone(nutanixMachineDetailsVar.GPUs)
-
 			spec.Project = nutanixMachineDetailsVar.Project.DeepCopy()
 
 			obj.Spec.Template.Spec = spec
--- a/pkg/webhook/cluster/nutanix_validator.go
+++ b/pkg/webhook/cluster/nutanix_validator.go
@@ -11,6 +11,7 @@ import (
 	"net/netip"
 
 	v1 "k8s.io/api/admission/v1"
+	"k8s.io/apimachinery/pkg/util/validation/field"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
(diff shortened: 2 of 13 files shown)


pkg/webhook/cluster/nutanix_validator.go
```diff
@@ -89,9 +90,113 @@ func (a *nutanixValidator) validate(
...
+	cluster *clusterv1.Cluster,
+) error {
+	fldErrs := field.ErrorList{}
+	fldPath := field.NewPath("spec", "topology")
+	// Validate that either failureDomains is set, or cluster and subnets are set with machineDetails, for control plane.
+	if clusterConfig.ControlPlane != nil &&
+		clusterConfig.ControlPlane.Nutanix != nil &&
+		len(clusterConfig.ControlPlane.Nutanix.FailureDomains) == 0 {
+		machineDetails := clusterConfig.ControlPlane.Nutanix.MachineDetails
+		if machineDetails.Cluster == nil || (!machineDetails.Cluster.IsName() && !machineDetails.Cluster.IsUUID()) {
```
faiq, line 112: do IsName and IsUUID give mutually exclusive results? is there a way to simplify this if statement?

pkg/webhook/cluster/nutanix_validator.go
```diff
@@ -89,9 +90,113 @@ func (a *nutanixValidator) validate(
...
+			))
+		}
+	}
+
+	// Validate either failureDomains is set, or cluster and sebnets are set with machineDetails, for workers.
+	if cluster.Spec.Topology.Workers != nil {
+		for i := range cluster.Spec.Topology.Workers.MachineDeployments {
+			md := cluster.Spec.Topology.Workers.MachineDeployments[i]
+			if md.FailureDomain != nil && *md.FailureDomain != "" {
+				// failureDomain is configured
```
faiq, line 148:
```suggestion
				// failureDomain is configured so we can skip checking if cluster and subnet is set 
```

dlipovetsky:
We are changing a field from required to optional. That is a backward compatible change, because clients can continue to provide specify the field.

However, the fields are only optional in some cases. I am concerned about this. Consider this example: I configure a worker pool with a failureDomain. I therefore do not configure the cluster and subnet in the Cluster spec itself. Later, I remove the failureDomain (it is mutable), but I don't update the Cluster spec. At this point, there is no source for the cluster and subnet. Will this change be rejected by validation?

I think the correct direction is to make FailureDomains mandatory in CAPX. That will simplify this API by having at most a single source of truth. But that's out of the scope of this PR, of course.

thunderboltsid:
We need to take care of cases where the variables override isn't present but the top-level worker config is. Feel free to use changes/tests I hacked up in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/compare/jira/NCN-108428...issue/failuredomain-cluster-webhook

thunderboltsid (changes requested):
We need to take care of cases where the variables override isn't present but the top-level worker config is. Feel free to use changes/tests I hacked up in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/compare/jira/NCN-108428...issue/failuredomain-cluster-webhook if it helps.

yanhua121:
> We are changing a field from required to optional. That is a backward compatible change, because clients can continue to provide specify the field.
> 
> However, the fields are only optional in some cases. I am concerned about this. Consider this example: I configure a worker pool with a failureDomain. I therefore do not configure the cluster and subnet in the Cluster spec itself. Later, I remove the failureDomain (it is mutable), but I don't update the Cluster spec. At this point, there is no source for the cluster and subnet. Will this change be rejected by validation?
> 
> I think the correct direction is to make FailureDomains mandatory in CAPX. That will simplify this API by having at most a single source of truth. But that's out of the scope of this PR, of course.

@dlipovetsky In the above scenario you described, when user trying to update the Cluster topology to remove the  failureDomain(s) reference but the corresponding machineDetail's does not have the cluster and/or subnets configuration, the Cluster webhook would reject this update and report the validation errors (see the sample errors I put in the PR description area).
When we bump up the caren and capx api version, we can have FailureDomains mandatory.

pkg/handlers/nutanix/mutation/machinedetails/inject.go
```diff
@@ -97,7 +97,14 @@ func (h *nutanixMachineDetailsPatchHandler) Mutate(
...
 
 			spec.BootType = nutanixMachineDetailsVar.BootType
-			spec.Cluster = nutanixMachineDetailsVar.Cluster
+			if nutanixMachineDetailsVar.Cluster != nil {
+				spec.Cluster = *nutanixMachineDetailsVar.Cluster
+			} else {
+				// Have to set the required Type field.
+				spec.Cluster = capxv1.NutanixResourceIdentifier{
+					Type: capxv1.NutanixIdentifierName,
+				}
```
thunderboltsid, line 106: why is this needed?

yanhua121: Since the NutanixMachine.spec.Cluster is not a pointer, when it's an empty object with default field values, the Type will get the default empty string "" value, this will fail when validation with the Type field rule of the enum values. See:

type NutanixResourceIdentifier struct {
	// Type is the identifier type to use for this resource.
	// +kubebuilder:validation:Enum:=uuid;name
	Type NutanixIdentifierType `json:"type"`

pkg/webhook/cluster/nutanix_validator_test.go
```diff
@@ -221,3 +226,196 @@ func TestValidatePrismCentralIPDoesNotEqualControlPlaneIP(t *testing.T) {
...
+		{
+			name:          "worker failureDomain not configured and missing machineDetail cluster and subnets",
+			clusterConfig: fakeClusterConfigSpec(true, false, false),
+			cluster:       fakeCluster(t, false, false, false),
+			expectedErr:   true,
+			expectedErrMessages: []string{
+				"spec.topology.workers.machineDeployments.variables.overrides.workerConfig.value.nutanix.machineDetails.cluster: Required value: \"cluster\" must set when failureDomain is not configured.", //nolint:lll // Message is long.
+				"spec.topology.workers.machineDeployments.variables.overrides.workerConfig.value.nutanix.machineDetails.subnets: Required value: \"subnets\" must set when failureDomain is not configured.", //nolint:lll // Message is long.
+			},
+		},
```
thunderboltsid, line 307: There are few cases to this:
* top-level workerConfig does not exist and only variable overrides exist
* top-level workerConfig exists but variable overrides don't exist
* top-level workerConfig and variable overrides both exist

pkg/webhook/cluster/nutanix_validator.go
```diff
@@ -89,9 +90,144 @@ func (a *nutanixValidator) validate(
...
+							err,
+						),
+					))
+				}
+
+				path := fldPath2
+				if workerConfig == nil {
+					workerConfig = defaultWorkerConfig
+					path = fldPath1
+				}
```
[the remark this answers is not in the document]
thunderboltsid: example contrary cases.
```go
    - class: default-worker
        metadata:
          annotations:
            cluster.x-k8s.io/cluster-api-autoscaler-node-group-max-size: "3"
            cluster.x-k8s.io/cluster-api-autoscaler-node-group-min-size: "3"
        name: md-no-overrides
      - class: default-worker
        failureDomain: fd-2
        metadata:
          annotations:
            cluster.x-k8s.io/cluster-api-autoscaler-node-group-max-size: "3"
            cluster.x-k8s.io/cluster-api-autoscaler-node-group-min-size: "3"
        name: md-no-overrides-failure-domain
```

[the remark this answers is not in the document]
thunderboltsid: See the conditional this conditional is in. https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1217/files#diff-cbaf86208d1803beb69b004fc24fe8d3dccbc2cba74f10dde9f8e8acf5edebeeR188

pkg/webhook/cluster/nutanix_validator.go
```diff
@@ -89,9 +90,144 @@ func (a *nutanixValidator) validate(
...
+
+	return fldErrs
+}
+
+// validateWorkerFailureDomainConfig validates either failureDomains is set,
+// or cluster and subnets are set with machineDetails, for workers.
+func validateWorkerFailureDomainConfig(cluster *clusterv1.Cluster) field.ErrorList {
+	fldErrs := field.ErrorList{}
+	fldPath1 := field.NewPath("spec", "topology", "variables", "workerConfig")
+	fldPath2 := field.NewPath(
```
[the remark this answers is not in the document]
yanhua121: changed

pkg/handlers/nutanix/mutation/machinedetails/inject.go
thunderboltsid: Go 1.24 can save us here https://www.bytesizego.com/blog/go-124-omitzero

thunderboltsid (approved):

thunderboltsid:
Merging this to unblock dev builds. Let's sync up next week on whether we should be using XOR semantics instead.

PR #1218, opened 2025-07-17
fix(auto-cert-renewal): adds 0 as valid value for daysBeforeExpiry

atulv7:
As of now to disable CAPI based automated certificate renewal we have to set `autoRenewCertificates` to `nil` which results in SSA patch [issue](https://github.com/kubernetes/kubernetes/issues/117447). To avoid this added 0 as a valid value for `autoRenewCertificates.daysBeforeExpiry` denoting the feat is disabled.

Fixes [NCN-108329](https://jira.nutanix.com/browse/NCN-108329)

- Tested using the dev workflow along with the konvoy [changes](https://github.com/mesosphere/konvoy2/pull/3959)

### DIFF
--- a/api/v1alpha1/controlplane_types.go
+++ b/api/v1alpha1/controlplane_types.go
@@ -13,9 +13,10 @@ type GenericControlPlaneSpec struct {
 type AutoRenewCertificatesSpec struct {
 	// DaysBeforeExpiry indicates a rollout needs to be performed if the
 	// certificates of the control plane will expire within the specified days.
+	// Set to 0 to disable automated certificate renewal.
 	// +kubebuilder:validation:Required
-	// +kubebuilder:validation:Minimum=7
-	DaysBeforeExpiry int32 `json:"daysBeforeExpiry,omitempty"`
+	// +kubebuilder:validation:XValidation:rule="self == 0 || self >= 7",message="Value must be 0 or at least 7"
+	DaysBeforeExpiry int32 `json:"daysBeforeExpiry"`
 }
 
 // DockerControlPlaneSpec defines the desired state of the control plane for a Docker cluster.
--- a/pkg/handlers/generic/mutation/autorenewcerts/inject.go
+++ b/pkg/handlers/generic/mutation/autorenewcerts/inject.go
@@ -67,12 +67,7 @@ func (h *autoRenewCerts) Mutate(
 		h.variableName,
 		h.variableFieldPath...,
 	)
-	if err != nil {
-		if variables.IsNotFoundError(err) {
-			log.V(5).Info("Control Plane auto renew certs variable not defined")
-			return nil
-		}
-
+	if err != nil && !variables.IsNotFoundError(err) {
 		return err
 	}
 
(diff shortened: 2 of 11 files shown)


pkg/handlers/generic/mutation/autorenewcerts/inject.go
```diff
@@ -67,12 +67,7 @@ func (h *autoRenewCerts) Mutate(
...
 	)
-	if err != nil {
-		if variables.IsNotFoundError(err) {
-			log.V(5).Info("Control Plane auto renew certs variable not defined")
-			return nil
-		}
-
+	if err != nil && !variables.IsNotFoundError(err) {
 		return err
 	}
```
manoj-nutanix, line 72: revert this, maybe earlier code was meant to log if variable is not provided.

manoj-nutanix (approved):

api/v1alpha1/controlplane_types.go
```diff
@@ -13,9 +13,10 @@ type GenericControlPlaneSpec struct {
 type AutoRenewCertificatesSpec struct {
 	// DaysBeforeExpiry indicates a rollout needs to be performed if the
 	// certificates of the control plane will expire within the specified days.
+	// Set to 0 to disable automated certificate renewal.
 	// +kubebuilder:validation:Required
-	// +kubebuilder:validation:Minimum=7
-	DaysBeforeExpiry int32 `json:"daysBeforeExpiry,omitempty"`
```
jimmidyson, line 18: As this is required, it should never have had `omitempty` here anyway. Thanks for fixing!

pkg/handlers/generic/mutation/autorenewcerts/inject_test.go
```diff
@@ -115,6 +115,31 @@ var _ = Describe("Generate auto renew patches", func() {
...
+					),
+				},
+				RequestItem: request.NewKubeadmControlPlaneTemplateRequestItem(""),
+				UnexpectedPatchMatchers: []capitest.JSONPatchMatcher{{
+					Operation:    "add",
+					Path:         "/spec/template/spec",
+					ValueMatcher: gomega.HaveKey("rolloutBefore"),
+				}},
+			},
+		},
```
jimmidyson, line 142: How about a negative test here too to check that 1-6 are invalid?

dkoshkin (approved):
Good solution.

I was not able to reproduce what you were seeing from https://github.com/kubernetes/kubernetes/issues/117447

But I also tried many different changes to get the SSA with `sigs.k8s.io/controller-runtime/pkg/client` client to remove the field when `AutoRenewCertificatesSpec` is set to nil but it was not removing it, even after removing `omitempty` and explicitly setting
```
autoRenewCertificates: null
```

I also verified that the SSA manager was the same for create and patch and made no difference.
Using Force on the SSA client also did not help.

pkg/handlers/generic/mutation/autorenewcerts/inject_test.go
faiq: That would be very useful

jimmidyson (approved):

pkg/handlers/generic/mutation/autorenewcerts/inject_test.go
dkoshkin: I wanted to help out @atulv7 and get this PR in today, but it turned out to be pretty involved.
Adding a test exposed 2 existing issues in our test code:
1. https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1218/commits/40b373de8c4aad794de235904476aa4a6b990fb3 Now that we're testing `DaysBeforeExpiry int32` with CEL its incompatible with standard `json.Unmarshal` since it treats every number as a float64, leading to an error like this:
```
invalid data, expected int, got float64 evaluating rule: Value must be 0 or at least 7
```
2. https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1218/commits/e5bfc50913dab61fa21c119944a66932c2959d54 Our CEL validator code in the test package was not properly traversing the struct so it was not failing as expected.
3. https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1218/commits/a3aa911bdf93ec1a756981cdb9b1053efcb10d02 added passing and a failing test. Since the test is just setting `ExpectError: true` see screenshot from my IDE with correct error:

image

PR #1227, opened 2025-07-22
feat: Add scale from zero cluster-autoscaler support

jimmidyson:
Added documentation for how to use with providers that do not natively support Cluster Autoscaler scale-from-zero.

Tested with CAPA (native support) and CAPD (via annotations) and works nicely.

Fixes #


docs/content/addons/cluster-autoscaler.md
```diff
@@ -50,6 +50,78 @@ spec:
...
+        - class: default-worker
+          metadata:
+            annotations:
+              # Set the following annotations to configure the Cluster Autoscaler
+              # The initial MachineDeployment will have 0 Machines
+              cluster.x-k8s.io/cluster-api-autoscaler-node-group-max-size: "3"
+              cluster.x-k8s.io/cluster-api-autoscaler-node-group-min-size: "0"
+              capacity.cluster-autoscaler.kubernetes.io/cpu: "8"
+              capacity.cluster-autoscaler.kubernetes.io/memory: "8112564Ki"
+            capacity.cluster-autoscaler.kubernetes.io/labels: "node-restriction.kubernetes.io/my-app="
```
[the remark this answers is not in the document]
jimmidyson: This is incorrect - see https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/clusterapi/README.md#scale-from-zero-support for details.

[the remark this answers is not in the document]
dkoshkin: I think its getting confused because of incorrect indentation.

dkoshkin, line 118:
```suggestion
              capacity.cluster-autoscaler.kubernetes.io/labels: "node-restriction.kubernetes.io/my-app="
```

charts/cluster-api-runtime-extensions-nutanix/addons/cluster-autoscaler/values-template.yaml
```diff
@@ -33,3 +33,35 @@ extraVolumeSecrets:
 rbac:
   # Create a Role instead of a ClusterRoles to update cluster-api objects
   clusterScoped: false
+{{- $capiProvider := index .Cluster.Labels "cluster.x-k8s.io/provider" }}
```
dkoshkin, line 36: We talked about this label `cluster.x-k8s.io/provider` in the context of different work and how its technically optional.
Since the example Cluster templates have this label, Im ok with continuing to use it here and in a [couple of other places](https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/main/pkg/webhook/cluster/nutanix_validator.go#L61), but we should consider enforcing it exists, or default based on some other information in the Cluster and its templates.

dkoshkin (approved):

charts/cluster-api-runtime-extensions-nutanix/addons/cluster-autoscaler/values-template.yaml
jimmidyson: Agreed. I've added some docs on this now to hopefully make this easier to use and understand.

PR #1229, opened 2025-07-23
fix: Namespacesync copies resources after partial copy failure

dlipovetsky:
Backport of #1228

By design, the namespacesync controller does not update resources in the target namespace. However, if it only copies some resources before being interrupted (i.e. the controller restarts), it will never copy the remaining resources.

This PR adds a (failing) test that simulates this scenario, and then adds a fix.

Fixes https://jira.nutanix.com/browse/NCN-108612

Fixes #


jimmidyson (approved):
Thanks!

supershal (approved):

PR #1233, opened 2025-07-23
fix(webhook): Update cluster webhook to validate failure domains

thunderboltsid:
Use XOR instead of OR for cluster and subnet presence check against failure domains on control plane and worker overrides.

### DIFF
--- a/pkg/webhook/cluster/nutanix_validator.go
+++ b/pkg/webhook/cluster/nutanix_validator.go
@@ -117,8 +117,8 @@ func validateTopologyFailureDomainConfig(
 	return fldErrs.ToAggregate()
 }
 
-// validateControlPlaneFailureDomainConfig validates that either failureDomains are set,
-// or cluster and subnets are set with machineDetails, for control plane.
+// validateControlPlaneFailureDomainConfig validates XOR behavior: either failureDomains are set
+// OR cluster and subnets are set with machineDetails, but not both, for control plane.
 func validateControlPlaneFailureDomainConfig(clusterConfig *variables.ClusterConfigSpec) field.ErrorList {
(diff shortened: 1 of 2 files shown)


pkg/webhook/cluster/nutanix_validator.go
```diff
@@ -132,31 +132,54 @@ func validateControlPlaneFailureDomainConfig(clusterConfig *variables.ClusterCon
...
+	if clusterConfig.ControlPlane != nil && clusterConfig.ControlPlane.Nutanix != nil {
 		machineDetails := clusterConfig.ControlPlane.Nutanix.MachineDetails
-		if machineDetails.Cluster == nil || (!machineDetails.Cluster.IsName() && !machineDetails.Cluster.IsUUID()) {
-			fldErrs = append(fldErrs, field.Required(
-				fldPath.Child("cluster"),
-				"\"cluster\" must be set when failureDomains are not configured.",
-			))
-		}
+		hasFailureDomains := len(clusterConfig.ControlPlane.Nutanix.FailureDomains) > 0
+		hasCluster := machineDetails.Cluster != nil &&
```
dkoshkin, line 138: nit: there are 2 other places that we have the same checks, we could add a Method on NutanixMachineDetails to do this

dkoshkin (approved):

dkoshkin (approved):

mhrabovcin (approved):
TIL about `omitzero`

thunderboltsid:
> TIL about `omitzero`

Yeah, it started with go1.24

thunderboltsid:
had to manually push commits due to signed commit requirement

PR #1234, opened 2025-07-23
feat: Support context for all Nutanix client calls

dlipovetsky:
This allows us to handle context cancellation when using the Nutanix v4 go client, which itself does not support context. That, together with #1235, allows us to return check results before the API server webhook timeout. This means the client sees results, instead of an uninformative (and misleading) "failed to call webhook" error.

## Before

Notice that no results are returned.

```
│ kubectl apply --dry-run=server -f invalid-bad-prismcentral-url.yaml -v=4
I0723 12:12:17.523594 2625398 envvar.go:172] "Feature gate default state" feature="ClientsPreferCBOR" enabled=false
I0723 12:12:17.523624 2625398 envvar.go:172] "Feature gate default state" feature="InformerResourceVersion" enabled=false
I0723 12:12:17.523627 2625398 envvar.go:172] "Feature gate default state" feature="InOrderInformers" enabled=true
I0723 12:12:17.523630 2625398 envvar.go:172] "Feature gate default state" feature="WatchListClient" enabled=false
I0723 12:12:17.523633 2625398 envvar.go:172] "Feature gate default state" feature="ClientsAllowCBOR" enabled=false
I0723 12:12:27.582271 2625398 helpers.go:246] server response object: %s[{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {},
  "status": "Failure",
  "message": "error when creating \"invalid-bad-prismcentral-url.yaml\": Internal error occurred: failed calling webhook \"preflight.cluster.caren.nutanix.com\": failed to call webhook: Post \"https://cluster-api-runtime-extensions-nutanix-admission.default.svc:443/preflight-v1beta1-cluster?timeout=10s\": context deadline exceeded",
  "reason": "InternalError",
  "details": {
    "causes": [
      {
        "message": "failed calling webhook \"preflight.cluster.caren.nutanix.com\": failed to call webhook: Post \"https://cluster-api-runtime-extensions-nutanix-admission.default.svc:443/preflight-v1beta1-cluster?timeout=10s\": context deadline exceeded"
      }
    ]
  },
  "code": 500
}]
Error from server (InternalError): error when creating "invalid-bad-prismcentral-url.yaml": Internal error occurred: failed calling webhook "preflight.cluster.caren.nutanix.com": failed to call webhook: Post "https://cluster-api-runtime-extensions-nutanix-admission.default.svc:443/preflight-v1beta1-cluster?timeout=10s": context deadline exceeded
```

## After

Notice that the results are returned.

```
│ kubectl apply --dry-run=server -f invalid-bad-prismcentral-url.yaml -v=4
I0723 12:06:59.458667 2603987 envvar.go:172] "Feature gate default state" feature="ClientsAllowCBOR" enabled=false
I0723 12:06:59.458703 2603987 envvar.go:172] "Feature gate default state" feature="ClientsPreferCBOR" enabled=false
I0723 12:06:59.458706 2603987 envvar.go:172] "Feature gate default state" feature="InformerResourceVersion" enabled=false
I0723 12:06:59.458709 2603987 envvar.go:172] "Feature gate default state" feature="InOrderInformers" enabled=true
I0723 12:06:59.458711 2603987 envvar.go:172] "Feature gate default state" feature="WatchListClient" enabled=false
I0723 12:07:07.583019 2603987 helpers.go:246] server response object: %s[{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {},
  "status": "Failure",
  "message": "error when creating \"invalid-bad-prismcentral-url.yaml\": admission webhook \"preflight.cluster.caren.nutanix.com\" denied the request: preflight checks failed due to an internal error",
  "reason": "InternalError",
  "details": {
    "causes": [
      {
        "reason": "NutanixCredentials",
        "message": "Failed to validate credentials: Get \"https://example.com:9441/api/nutanix/v3/users/me\": context deadline exceeded. This is usually a temporary error. Please retry.",
        "field": "$.spec.topology.variables[?@.name==\"clusterConfig\"].value.nutanix.prismCentralEndpoint"
      }
    ]
  },
  "code": 500
}]
Error from server (InternalError): error when creating "invalid-bad-prismcentral-url.yaml": admission webhook "preflight.cluster.caren.nutanix.com" denied the request: preflight checks failed due to an internal error
```
 
Fixes #

### DIFF
--- a/pkg/webhook/preflight/nutanix/clients.go
+++ b/pkg/webhook/preflight/nutanix/clients.go
@@ -368,31 +368,29 @@ func (c *clientWrapper) ListSubnets(
 // callWithContext is a helper function that immediately responds to context cancellation,
 // while calling a long-running, non-preemptible function. The long-running function always
 // runs to completion, but its result is only returned if the context is not cancelled.
-func callWithContext[T any](ctx context.Context, f func() (T, error)) (resp T, err error) {
-	respCh := make(chan T)
-	errCh := make(chan error)
+func callWithContext[T any](ctx context.Context, f func() (T, error)) (T, error) {
+	type result[T any] struct {
+		val T
+		err error
+	}
+
+	// The buffered channel allows us to send the result of the function without knowing
+	// whether the channel will be read. If the context is not cancelled, this function will
+	// read from the channel and return the value to caller. If the context is cancelled, we
+	// will not read from the channel. Once this function returns, the channel will go out
+	// of scope, and be garbage collected.
+	ch := make(chan result[T], 1)
 
 	go func() {
 		resp, err := f()
-		select {
-		case <-ctx.Done():
-			// Context was cancelled before function returned. We assume no one wants the result anymore.
-		default:
-			if err != nil {
-				errCh <- err
-			}
-			respCh <- resp
-		}
-		close(respCh)
-		close(errCh)
+		ch <- result[T]{val: resp, err: err}
 	}()
 
 	select {
 	case <-ctx.Done():
-		return resp, ctx.Err()
-	case err := <-errCh:
-		return resp, err
-	case resp := <-respCh:
-		return resp, nil
+		var zero T
+		return zero, ctx.Err()
+	case res := <-ch:
+		return res.val, res.err
 	}
 }
--- a/pkg/webhook/preflight/nutanix/clients.go
+++ b/pkg/webhook/preflight/nutanix/clients.go
@@ -26,6 +26,7 @@ type client interface {
 	)
 
 	GetImageById(
+		ctx context.Context,
 		uuid *string,
 		args ...map[string]interface{},
 	) (
@@ -34,6 +35,7 @@ type client interface {
 	)
 
 	ListImages(
+		ctx context.Context,
 		page_ *int,
 		limit_ *int,
 		filter_ *string,
@@ -46,6 +48,7 @@ type client interface {
 	)
 
 	GetClusterById(
+		ctx context.Context,
 		uuid *string,
 		args ...map[string]interface{},
 	) (
@@ -53,6 +56,7 @@ type client interface {
 	)
 
 	ListClusters(
+		ctx context.Context,
 		page_ *int,
 		limit_ *int,
 		filter_ *string,
@@ -65,6 +69,7 @@ type client interface {
 		error,
 	)
 	ListStorageContainers(
+		ctx context.Context,
 		page_ *int,
 		limit_ *int,
 		filter_ *string,
@@ -77,6 +82,7 @@ type client interface {
 	)
 
 	GetSubnetById(
+		ctx context.Context,
 		uuid *string,
 		args ...map[string]interface{},
 	) (
@@ -84,6 +90,7 @@ type client interface {
 	)
 
 	ListSubnets(
+		ctx context.Context,
 		page_ *int,
 		limit_ *int,
 		filter_ *string,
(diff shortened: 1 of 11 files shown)


pkg/webhook/preflight/nutanix/clients.go
```diff
@@ -83,126 +197,202 @@ func newClient(
...
+		)
+	})
+}
+
+// callWithContext is a helper function that immediately responds to context cancellation,
+// while calling a long-running, non-preemptible function. The long-running function always
+// runs to completion, but its result is only returned if the context is not cancelled.
+func callWithContext[T any](ctx context.Context, f func() (T, error)) (resp T, err error) {
+	respCh := make(chan T)
+	errCh := make(chan error)
```
thunderboltsid, line 373: consider using a single channel
```
type result[T any] struct {
	val T
	err error
}

func callWithContext[T any](ctx context.Context, f func() (T, error)) (T, error) {
	ch := make(chan result[T], 1)
        ...
}
```

pkg/webhook/preflight/nutanix/clients.go
```diff
@@ -83,126 +197,202 @@ func newClient(
...
+			// Context was cancelled before function returned. We assume no one wants the result anymore.
+		default:
+			if err != nil {
+				errCh <- err
+			}
+			respCh <- resp
+		}
+		close(respCh)
+		close(errCh)
+	}()
```
thunderboltsid, line 388: this should be sufficient when you use a buffered channel
```go
go func() {
        val, err := f()
        ch <- result[T]{val, err}
}()
```

thunderboltsid, line 388: If you use a buffered channel, you can simplify to
```go
go func() {
        v, e := f()
        ch <- result[T]{value: v, err: e}
}()
```

dlipovetsky: It's simpler, but I think it's harder to read/understand.

How do you know what happens to the result that's sent to to `ch`? That is, how can you be sure the channel will be garbage collected? You have to know that the result is sent to the buffer, and that if the `select` statement already terminated, then the channel will never be read, so once the enclosing function goes out of scope, the channel will be garbage-collected.

I guess I could explain this in a comment, though.

pkg/webhook/preflight/nutanix/clients.go
dlipovetsky: Nice. Thanks!

pkg/webhook/preflight/nutanix/clients.go
dlipovetsky: You convinced me :smile:. Thanks!

Done in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1234/commits/c74b9a615e300de61a6e0259a065d1a5fdfb0fae

pkg/webhook/preflight/nutanix/clients.go
dlipovetsky: Done in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1234/commits/c74b9a615e300de61a6e0259a065d1a5fdfb0fae

pkg/webhook/preflight/nutanix/clients.go
```diff
@@ -83,126 +197,200 @@ func newClient(
...
+			select_,
+			args...,
+		)
+	})
+}
+
+// callWithContext is a helper function that immediately responds to context cancellation,
+// while calling a long-running, non-preemptible function. The long-running function always
+// runs to completion, but its result is only returned if the context is not cancelled.
+func callWithContext[T any](ctx context.Context, f func() (T, error)) (T, error) {
```
dkoshkin, line 371: 👏

pkg/webhook/preflight/nutanix/clients.go
```diff
@@ -18,10 +18,25 @@ import (
...
+		ctx context.Context,
+	) (
+		*prismv3.UserIntentResponse,
+		error,
+	)
+
+	GetImageById(
+		ctx context.Context,
+		uuid *string,
+		args ...map[string]interface{},
```
dkoshkin, line 31: Can you please explain how these `args` are used with the `ctx` change?

dkoshkin (approved):
Nice solution!

pkg/webhook/preflight/nutanix/clients.go
dlipovetsky: Good question, to which I have no answer. I don't know their purpose :cry: . But because the actual client accepts them as arguments, I didn't want to remove them from the wrapper.

pkg/webhook/preflight/nutanix/clients.go
dlipovetsky: Thanks! It took a bunch of iterations, but I found a way to do it! And then Sid made a couple suggestions that made it simpler.

thunderboltsid (approved):
lgtm

PR #1252, opened 2025-08-06
fix: Correctly handle variable overrides

jimmidyson:
Currently the metamutator only passes the override values when they are
present on a machine deployment or control plane config. This commit
fixes that by merging the global variables into the overrides, resulting
in the behaviour that we always expected.

As overrides are not that commonly used yet, we have not seen this issue
but as taints, etc become more commonly used this will become an issue.

Tests added to prove the fix as well.


common/pkg/capi/clustertopology/handlers/mutation/meta.go
```diff
@@ -108,20 +127,88 @@ func (mgp metaGeneratePatches) GeneratePatches(
...
 			}
 
 			log.V(3).Info("Mutation pipeline completed successfully", "handlerCount", len(mgp.mutators))
 			return nil
 		},
 	)
 }
+
+func mergeVariableDefinitions(
+	vars, globalVars map[string]apiextensionsv1.JSON,
```
faiq, line 166:
```suggestion
	overrideVars, globalVars map[string]apiextensionsv1.JSON,
```

Can you use a name that's better reflects what this function is intending to do? A comment here would also be helpful.

common/pkg/capi/clustertopology/handlers/mutation/meta.go
```diff
@@ -108,20 +127,88 @@ func (mgp metaGeneratePatches) GeneratePatches(
...
+				log.Error(err, "Failed to merge global variables")
+				return err
+			}
 
-				if err := h.Mutate(ctx, obj.(*unstructured.Unstructured), vars, holderRef, clusterKey, clusterGetter); err != nil {
-					log.Error(err, "Mutator failed", "index", i, "handler", handlerName)
+			for i, h := range mgp.mutators {
+				mutatorType := fmt.Sprintf("%T", h)
+				log.V(5).
+					Info("Running mutator", "index", i, "handler", mutatorType, "vars", vars)
```
mhrabovcin, line 142:
```suggestion
					Info("Running mutator", "index", i, "handler", mutatorType, "vars", mergedVars)
```

common/pkg/capi/clustertopology/handlers/mutation/meta.go
jimmidyson: Done.

common/pkg/capi/clustertopology/handlers/mutation/meta.go
```diff
@@ -108,20 +127,90 @@ func (mgp metaGeneratePatches) GeneratePatches(
 
 			log.V(3).Info("Starting mutation pipeline", "handlerCount", len(mgp.mutators))
 
-			for i, h := range mgp.mutators {
-				handlerName := fmt.Sprintf("%T", h)
-				log.V(5).Info("Running mutator", "index", i, "handler", handlerName)
+			// Merge the global variables to the current resource vars. This allows the handlers to access
+			// the variables defined in the cluster class or cluster configuration and use these correctly when
+			// overrides are specified on machine deployment or control plane configuration.
+			mergedVars, err := mergeVariableOverridesWithGlobal(vars, globalVars)
```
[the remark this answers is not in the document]
jimmidyson: I also thought CAPI did the merging, but as you can see it only does a shallow merge, rather than a deep merge.

That CAPI code is called to build the variables to call the handlers with. The request includes the global variables and then also passes the shallow merged variables per templated resource - this latter is what we were using prior to this fix and hence we were only using the override variable if one was specified.

This breaks things like specifying defaults in global vars (e.g. instance type) and only specifying overrides per MachineDeployment, e.g. taints. Prior to this fix, we would only pass the taints to the handler and the instance type would not have been set. After this fix, the handler gets the merged global and override vars.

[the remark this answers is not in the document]
dkoshkin: >The request includes the global variables and then also passes the shallow merged variables per templated resource
Aha, that makes sense, thanks for the explanation!

dkoshkin (approved):
Thanks for the new tests

jimmidyson:
@thunderboltsid So we 'd need to do merges in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/main/pkg/webhook/cluster/nutanix_validator.go#L238?

common/pkg/capi/clustertopology/variables/merge_test.go
```diff
@@ -0,0 +1,207 @@
...
+				vars: map[string]apiextensionsv1.JSON{
+					"a": {Raw: []byte(`{"x":1,"y":{"a": 2,"b": null}}`)},
+				},
+				globalVars: map[string]apiextensionsv1.JSON{
+					"a": {Raw: []byte(`{"y":{"a": 2,"b":{"c": 5, "d": 6}},"z":4}`)},
+				},
+			},
+			want: map[string]apiextensionsv1.JSON{
+				"a": {Raw: []byte(`{"x":1,"y":{"a": 2,"b":{"c": 5, "d": 6}},"z":4}`)},
+			},
```
thunderboltsid, line 125: We need to ensure that overrides have a mechanism to nullify the global vars.

PR #1255, opened 2025-08-06
feat: EKS ClusterClass and Cluster example

jimmidyson:

Fixes #

The bases are included in the hack directory rather than pulling in from external projects as we do for other providers because the EKS CC support has not been released in CAPA yet. Once that has happened, we can switch to the same strategy and use the upstream CC and Cluster as the base for patching to create the CAREN CC/Cluster examples.

Requires #1254 and #1252.


make/clusterctl.mk
```diff
@@ -25,9 +25,9 @@ clusterctl.init:
 	      --infrastructure docker:$(CAPD_VERSION),aws:$(CAPA_VERSION),nutanix:$(CAPX_VERSION) \
 	      --addon helm:$(CAAPH_VERSION) \
 	      --wait-providers
-	kubectl apply --server-side --force-conflicts \
+	kubectl --kubeconfig=$(KIND_KUBECONFIG) apply --server-side --force-conflicts \
```
dkoshkin, line 28: How did this work before 🤔

dkoshkin (approved):
Nice! Lets wait for the stacked PRs and other dependent PRs to merege

dkoshkin:
Until we decide what to do with https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1252

Should we set the vars directly in the MD variable? https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/commit/28183a4fb87823fac68c3e26c6417e5230c2420c

PR #1257, opened 2025-08-06
feat: Make the Nutanix top-level node specs required

dlipovetsky:
The current schema marks the Nutanix top-level node specs optional. (The same is true for top-level specs of other providers, such as AWS and Docker, but I'll get to those in the future).

This excerpt tries to demonstrate what I mean:
````yaml
        controlPlane:
          nutanix: # <- This is marked optional! But if it is missing, no VMs can be created.
            machineDetails:
              bootType: uefi
              cluster:
                name: example
                type: name
              image:
                name: example
                type: name
              memorySize: 16Gi
              subnets:
              - name: example
                type: name
              systemDiskSize: 80Gi
              vcpuSockets: 4
              vcpusPerSocket: 1
````

This draft PR marks the specs required. This is a breaking API change. I have not bumped the API version, because I want to first understand the code changes that are required, but are unrelated an API version change.    
 
Fixes #

### DIFF
--- a/api/v1alpha1/controlplane_types.go
+++ b/api/v1alpha1/controlplane_types.go
@@ -31,7 +31,7 @@ type DockerControlPlaneSpec struct {
 
 // NutanixControlPlaneSpec defines the desired state of the control plane for a Nutanix cluster.
 type NutanixControlPlaneSpec struct {
-	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Required
 	Nutanix *NutanixControlPlaneNodeSpec `json:"nutanix,omitempty"`
 
 	GenericControlPlaneSpec `json:",inline"`
--- a/api/v1alpha1/crds/caren.nutanix.com_nutanixclusterconfigs.yaml
+++ b/api/v1alpha1/crds/caren.nutanix.com_nutanixclusterconfigs.yaml
@@ -341,6 +341,7 @@ spec:
                           type: array
                           x-kubernetes-list-type: set
                         machineDetails:
+                          description: machineDetails specifies the details of the Nutanix machine that will be created
                           properties:
                             additionalCategories:
                               description: |-
@@ -553,6 +554,8 @@ spec:
                           - key
                         type: object
                       type: array
+                  required:
+                    - nutanix
                   type: object
                 dns:
                   description: DNS defines the DNS configuration for the cluster.
--- a/api/v1alpha1/crds/caren.nutanix.com_nutanixworkernodeconfigs.yaml
+++ b/api/v1alpha1/crds/caren.nutanix.com_nutanixworkernodeconfigs.yaml
@@ -61,6 +61,7 @@ spec:
                 nutanix:
                   properties:
                     machineDetails:
+                      description: machineDetails specifies the details of the Nutanix machine that will be created
                       properties:
                         additionalCategories:
                           description: |-
@@ -273,6 +274,8 @@ spec:
                       - key
                     type: object
                   type: array
+              required:
+                - nutanix
               type: object
           type: object
       served: true
--- a/api/v1alpha1/nodeconfig_types.go
+++ b/api/v1alpha1/nodeconfig_types.go
@@ -96,7 +96,7 @@ func (s NutanixWorkerNodeConfig) VariableSchema() clusterv1.VariableSchema { //n
 
 // NutanixWorkerNodeConfigSpec defines the desired state of NutanixWorkerNodeSpec.
 type NutanixWorkerNodeConfigSpec struct {
-	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Required
 	Nutanix *NutanixWorkerNodeSpec `json:"nutanix,omitempty"`
 
 	GenericNodeSpec `json:",inline"`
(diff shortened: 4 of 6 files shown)

### CODE REVIEW

**Comment by dlipovetsky** [comment]
Currently, multiple variable unit tests fail. I thought they were caused by missing specs, so I added those.

However, after I added the specs to one of the tests, it continues to fail. I don't really understand the origin of the error. As you can see, the file reported in the output (`pkg/handlers/generic/mutation/autorenewcerts/variables.go`) does not exist. I'm not sure why this file is reported. I assume the correct file is `common/pkg/testutils/capitest/variables.go`. The error appears to be returned by `openapi.ValidateClusterVariable`.

```
    /home/dlipovetsky/projects/cluster-api-runtime-extensions-nutanix/pkg/handlers/generic/mutation/autorenewcerts/variables.go:121:
        Unexpected error:
            <errors.aggregate | len:1, cap:1>:
            clusterConfig.value.controlPlane.nutanix.machineDetails: Invalid value: v1alpha1.NutanixClusterConfigSpec{Nutanix:(*v1alpha1.NutanixSpec)(nil), GenericClusterConfigSpec:v1alpha1.GenericClusterConfigSpec{KubernetesImageRepository:(*string)(nil), Etcd:(*v1alpha1.Etcd)(nil), Proxy:(*v1alpha1.HTTPProxy)(nil), ImageRegistries:[]v1alpha1.ImageRegistry(nil), GlobalImageRegistryMirror:(*v1alpha1.GlobalImageRegistryMirror)(nil), Users:[]v1alpha1.User(nil), EncryptionAtRest:(*v1alpha1.EncryptionAtRest)(nil), DNS:(*v1alpha1.DNS)(nil), KubeProxy:(*v1alpha1.KubeProxy)(nil), NTP:(*v1alpha1.NTP)(nil)}, Addons:(*v1alpha1.NutanixAddons)(nil), ControlPlane:(*v1alpha1.NutanixControlPlaneSpec)(0xc000f85080), ExtraAPIServerCertSANs:[]string(nil)}: invalid data, expected a map for the provided schema with type=object evaluating rule: Either 'image' or 'imageLookup' must be set, but not both
            [
                <*field.Error | 0xc000f0bf40>{
                    Type: "FieldValueInvalid",
                    Field: "clusterConfig.value.controlPlane.nutanix.machineDetails",
                    BadValue: <v1alpha1.NutanixClusterConfigSpec>{
                        Nutanix: nil,
                        GenericClusterConfigSpec: {
                            KubernetesImageRepository: nil,
                            Etcd: nil,
                            Proxy: nil,
                            ImageRegistries: nil,
                            GlobalImageRegistryMirror: nil,
                            Users: nil,
                            EncryptionAtRest: nil,
                            DNS: nil,
                            KubeProxy: nil,
                            NTP: nil,
                        },
                        Addons: nil,
                        ControlPlane: {
                            Nutanix: {
                                MachineDetails: {
                                    VCPUsPerSocket: 1,
                                    VCPUSockets: 2,
                                    MemorySize: {
                                        i: {value: 8589934592, scale: 0},
                                        d: {Dec: nil},
                                        s: "",
                                        Format: "BinarySI",
                                    },
                                    Image: {Type: "name", UUID: nil, Name: "fake-image"},
                                    ImageLookup: nil,
                                    Cluster: {
                                        Type: "name",
                                        UUID: nil,
                                        Name: "fake-pe-cluster",
                                    },
                                    Subnets: [
                                        {Type: "name", UUID: nil, Name: "fake-subnet"},
                                    ],
                                    AdditionalCategories: nil,
                                    BootType: "legacy",
                                    SystemDiskSize: {
                                        i: {value: 42949672960, scale: 0},
                                        d: {Dec: nil},
                                        s: "",
                                        Format: "BinarySI",
                                    },
                                    Project: nil,
                                    GPUs: nil,
                                },
                                FailureDomains: nil,
                            },
                            GenericControlPlaneSpec: {
                                AutoRenewCertificates: nil,
                            },
                            GenericNodeSpec: {
                                Taints: nil,
                                NodeRegistration: nil,
                            },
                        },
                        ExtraAPIServerCertSANs: nil,
                    },
                    Detail: "invalid data, expected a map for the provided schema with type=object evaluating rule: Either 'image' or 'imageLookup' must be set, but not both",
                },
            ]
        occurred
```

**Comment by jimmidyson** [comment]
```yaml
        controlPlane:
          nutanix: # <- This is marked optional! But if it is missing, no VMs can be created.
```

Technically, this isn't quite true. CAREN could be used in ClusterClass templates that already configure the control plane accordingly and do not require CAREN patches to apply. I know this is not a common use case, but it is possible.

jimmidyson:
Fixed the incorrect line reporting in e0c868f6c0f7a138f7418281928ea77a68a64b86 btw, will push a separate PR for that as well.

dlipovetsky:
> Technically, this isn't quite true. CAREN could be used in ClusterClass templates that already configure the control plane accordingly and do not require CAREN patches to apply. I know this is not a common use case, but it is possible.

Okay. That's something I was wondering myself.

This means there are two sources of truth: the ClusterClass templates, and the topology variables.

If these fields might be defined in the templates, the fields should remain optional in the topology variables.

On the other hand, if the fields are _not_ in the templates, and the fields are _not_ in the topology variables, then there is not enough information to create the Machines.

Nothing prevents this from scenario, so users could fall into it without realizing it.

PR #1286, opened 2025-09-03
feat: automatically enable Cilium kube-proxy replacement feature

dkoshkin:
This PR enables Cilium's kube-proxy replacement feature automatically when clusters disable kube-proxy installation for both new clusters and during upgrades.
During upgrades, the Cilium handler will apply the new configuration, wait for the DaemonSet to be rolled out and then delete the kube-proxy DaemonSet and its ConfigMap.

Fixes #


pkg/handlers/generic/lifecycle/addons/helmaddon.go
```diff
@@ -231,7 +231,7 @@ func waitToBeReady(
 				if obj.Generation != obj.Status.ObservedGeneration {
 					return false, nil
 				}
-				return conditions.IsTrue(obj, caaphv1.HelmReleaseProxiesReadyCondition), nil
+				return conditions.IsTrue(obj, clusterv1.ReadyCondition), nil
```
dkoshkin, line 234: `HelmReleaseProxiesReadyCondition` returns too early.

pkg/handlers/generic/lifecycle/cni/cilium/handler.go
```diff
@@ -231,11 +239,129 @@ func (c *CiliumCNI) apply(
...
+
+	return nil
+}
+
+const (
+	kubeProxyName      = "kube-proxy"
+	kubeProxyNamespace = "kube-system"
+)
+
+func waitForCiliumToBeReady(
```
dkoshkin, line 291: Initially I tried to just set HCP's helm `wait` option, but during upgrades it was getting `Ready` prematurely.
Keeping the wait outside also makes it possible to support the wait for other providers besides helm.

pkg/handlers/generic/lifecycle/addons/helmaddon.go
jimmidyson: Thanks! It's that `conditions.SetSummary` that sets the ready condition. This will possibly change when CAAPH supports v1beta2 conditions so we will need to look out for this.

yanhua121 (commented):
Since this PR is on stack of #1288, and I see in the PR #1288 (merged) the annotation "skip-kube-proxy" is not added to the file examples/capi-quick-start/nutanix-cluster-failuredomain-cilium-helm-addon.yaml, but added to the file examples/capi-quick-start/nutanix-cluster-cilium-helm-addon.yaml. Is this intentionally? Are these example files only used for e2e tests?

dlipovetsky (approved):
Our plan is to represent kube-proxy replacement in the API. But this makes sense in the meantime.

charts/cluster-api-runtime-extensions-nutanix/addons/cni/cilium/values-template.yaml
```diff
@@ -33,10 +33,10 @@ socketLB:
 envoy:
   image:
     useDigest: false
-k8sServiceHost: auto
 {{- with .ControlPlane }}
 {{- range $key, $val := .metadata.annotations }}
 {{- if eq $key "controlplane.cluster.x-k8s.io/skip-kube-proxy" }}
+k8sServiceHost: auto
```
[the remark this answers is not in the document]
dkoshkin: I've added it back because unfortunately just setting `kubeProxyReplacement` does not trigger a Pod rollout.
In a follow up I will remove it if the rest of the config for IP preservation does result in new Pods.

Although `kubeKroxyReplacement` is not really necessary with the current config without kube-proxy replacement.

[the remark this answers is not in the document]
jimmidyson: I do not like this - it's too opaque as to why the necessary pods are being rolled out. We should have a better defined trigger for this.

PR #1288, opened 2025-09-03
feat: enable Cilium kube-proxy replacement for new clusters

dkoshkin:
This PR enables Cilium's kube-proxy replacement feature automatically when clusters disable kube-proxy installation for new clusters.

Fixes #


hack/examples/overlays/clusters/aws/cilium/crs/kustomization.yaml.tmpl
```diff
@@ -20,3 +20,6 @@ patches:
   - target:
       kind: Cluster
     path: ../../../../../patches/aws/crs-strategy.yaml
+  - target:
```
dkoshkin, line 23: Intentionally leaving this off for Docker clusters to continue testing kube-proxy there.

jimmidyson:
@dkoshkin Pushed a follow up commit 9fe0052e64def17c43a77063d8fefd83cb3a4c51 that uses CAAPH templating rather than a separate templating function. Sorry I meant to send a PR to your PR branch but messed up and pushed directly, can revert if you disagree with the approach

jimmidyson (approved):

hack/examples/overlays/clusters/aws/cilium/helm-addon/kustomization.yaml.tmpl
```diff
@@ -14,3 +14,6 @@ patches:
   - target:
       kind: Cluster
     path: ../../../../../patches/cilium.yaml
+  - target:
+      kind: Cluster
+    path: ../../../../../patches/skip-kube-proxy.yaml
```
jimmidyson, line 19: @dkoshkin Only adding this to the HelmAddon strategy cluster examples.

CRS generation set the K8S_SERVICE_HOST literally to `auto`. When using
the Helm chart, `auto` looks up the host from the `cluster-info` ConfigMap
but this is not available when generating the CRS manifests. To use this,
we would need to update the CRS templates when applying them to the cluster
but we can defer that for now.

hack/addons/kustomize/cilium/kustomization.yaml.tmpl
```diff
@@ -29,5 +30,6 @@ helmCharts:
       enabled: false
       relay:
         enabled: false
+    k8sServiceHost: ""
```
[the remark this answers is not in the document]
dkoshkin: Verified on a running cluster, it does not

PR #1297, opened 2025-09-12
refactor: keep only common types in Generic config specs

dkoshkin:
After EKS types were introduced, the fields in `GenericClusterConfigSpec` and `GenericNodeSpec` are no longer all "generic" because some do not apply to EKS clusters, as it does not use kubeadm to create the cluster.

This PR attempts to address this by only keeping the fields that can be set for both EKS and all other infra types in `GenericClusterConfigSpec` and moves the rest of the fields to `KubeadmClusterConfigSpec`, and similarly for `GenericNodeSpec`.

Because the structs are included `inline` in the individual Cluster types this does not change the external yaml API, only how they are referenced in Go (notice how the CRDs for <aws|docker|nutanix>clusterconfigs did not change). 

Fixes #

Unit tests.

If we agree with this approach I will have separate refactor PRs to reorganize the handlers and the public docs.


api/v1alpha1/clusterconfig_types.go
```diff
@@ -177,29 +185,21 @@ type NutanixClusterConfigSpec struct {
...
 
-func (s GenericClusterConfig) VariableSchema() clusterv1.VariableSchema { //nolint:gocritic,lll // Passed by value for no potential side-effect.
-	return genericClusterConfigVariableSchema
+func (s KubeadmClusterConfig) VariableSchema() clusterv1.VariableSchema { //nolint:gocritic,lll // Passed by value for no potential side-effect.
+	return kubeadmClusterConfigVariableSchema
 }
 
-// GenericClusterConfigSpec defines the desired state of GenericClusterConfig.
-type GenericClusterConfigSpec struct {
+// KubeadmConfigSpec defines configuratiion that can be set when using kubeadm to bootstrap the cluster.
```
jimmidyson, line 201:
```suggestion
// KubeadmConfigSpec defines configuration that can be set when using kubeadm to bootstrap the cluster.
```

faiq:
nice work!

PR #1306, opened 2025-09-16
feat: Enable configurable parallel image pulls

jimmidyson:

Default behaviour is retained, that is serialized image pulls actually configured by disabling forced serialized image pulls and setting max parallel image pulls to 1.

A value of zero means unlimited parallel image pulls.

Fixes #

### DIFF
--- a/api/v1alpha1/clusterconfig_types.go
+++ b/api/v1alpha1/clusterconfig_types.go
@@ -252,7 +252,6 @@ type KubeadmClusterConfigSpec struct {
 	// If set to 0, the maximum number of parallel image pulls will be unlimited.
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:validation:Minimum=0
-	// +kubebuilder:validation:Default=1
 	MaxParallelImagePullsPerNode *int32 `json:"maxParallelImagePullsPerNode,omitempty"`
 }
 


thunderboltsid (approved):

pkg/handlers/generic/mutation/parallelimagepulls/inject.go
```diff
@@ -0,0 +1,174 @@
...
+		vars,
+		h.variableName,
+		h.variableFieldPath...,
+	)
+	switch {
+	case err != nil && !variables.IsNotFoundError(err):
+		return err
+	case variables.IsNotFoundError(err):
+		log.V(5).Info("max parallel image pulls is not set, using default value of 1")
+		maxParallelImagePullsPerNode = 1
```
dkoshkin, line 86: This will cause an unwanted Cluster rollout because this file will now always be written out. Should we remove the default in the API and leave the defaulting to Kubernetes?
Otherwise we need to update the handler versions and only include this in the new one.

jimmidyson: Updated the logic to prevent unwanted cluster rollout on CAREN upgrade.

api/v1alpha1/clusterconfig_types.go
```diff
@@ -246,6 +246,14 @@ type KubeadmClusterConfigSpec struct {
...
 	// +kubebuilder:validation:Optional
 	KubeProxy *KubeProxy `json:"kubeProxy,omitempty"`
+
+	// MaxParallelImagePullsPerNode defines the maximum number of parallel image pulls performed by each kubelet.
+	// If not set, the default value of 1 will be used.
+	// If set to 0, the maximum number of parallel image pulls will be unlimited.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Default=1
+	MaxParallelImagePullsPerNode *int32 `json:"maxParallelImagePullsPerNode,omitempty"`
```
dkoshkin, line 256: I may be wrong, but wouldn't having `+kubebuilder:validation:Default=1` in there means that its rendered in the variable?
Is that what we want? With the rest of the APIs the convention here is to keep it pretty minimal.
```suggestion
	// MaxParallelImagePullsPerNode defines the maximum number of parallel image pulls performed by each kubelet.
	// If not set, the default value of 1 will be used.
	// If set to 0, the maximum number of parallel image pulls will be unlimited.
	// +kubebuilder:validation:Optional
	// +kubebuilder:validation:Minimum=0
	MaxParallelImagePullsPerNode *int32 `json:"maxParallelImagePullsPerNode,omitempty"`
```

dkoshkin: Removing the default can also simplify the handler

jimmidyson: Removing the default won't change the handler because it will still need to handle being explicitly set to 1, which should still result in no rollout if previously unset - the effective value would be the same.

jimmidyson: I'll remove it.

dkoshkin (approved):

PR #1307, opened 2025-09-17
fix: Ciliums kube-proxy replacement rollout and wait

dkoshkin:
Make the migration process from kube-proxy to Cilium's kube-proxy replacement more resilient.
Just setting `kubeProxyReplacement: true` is not enough for the Cilium operator to restart the DaemonSet Pods and pickup the new configuration. Instead of relying on `k8sServiceHost` to cause a rollout, this change forces a rollout during the migration process.
This also fixes a potential race where the Cilium DaemonSet wait returned early and delete kube-proxy before all the Pods were restarted.

Another fix here is that this whole migration process is now safer and only done once when kube-proxy is installed.

Pulled out from https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1295

Fixes #

### DIFF
--- a/charts/cluster-api-runtime-extensions-nutanix/addons/cni/cilium/values-template.yaml
+++ b/charts/cluster-api-runtime-extensions-nutanix/addons/cni/cilium/values-template.yaml
@@ -33,11 +33,7 @@ socketLB:
 envoy:
   image:
     useDigest: false
-{{- with .ControlPlane }}
-{{- range $key, $val := .metadata.annotations }}
-{{- if eq $key "controlplane.cluster.x-k8s.io/skip-kube-proxy" }}
 k8sServiceHost: auto
-kubeProxyReplacement: true{{ break }}
-{{- end }}
-{{- end }}
+{{- if .EnableKubeProxyReplacement }}
+kubeProxyReplacement: true
 {{- end }}
--- a/pkg/handlers/generic/lifecycle/cni/cilium/handler.go
+++ b/pkg/handlers/generic/lifecycle/cni/cilium/handler.go
@@ -12,6 +12,7 @@ import (
 	"github.com/spf13/pflag"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	"sigs.k8s.io/cluster-api/controllers/remote"
@@ -228,6 +229,7 @@ func (c *CiliumCNI) apply(
 			c.client,
 			helmChart,
 		).
+			WithValueTemplater(templateValues).
 			WithDefaultWaiter()
 	case "":
 		resp.SetStatus(runtimehooksv1.ResponseStatusFailure)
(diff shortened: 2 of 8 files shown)


charts/cluster-api-runtime-extensions-nutanix/addons/cni/cilium/values-template.yaml
```diff
@@ -33,11 +33,7 @@ socketLB:
...
   image:
     useDigest: false
-{{- with .ControlPlane }}
-{{- range $key, $val := .metadata.annotations }}
-{{- if eq $key "controlplane.cluster.x-k8s.io/skip-kube-proxy" }}
 k8sServiceHost: auto
-kubeProxyReplacement: true{{ break }}
-{{- end }}
-{{- end }}
+{{- if .EnableKubeProxyReplacement }}
```
dkoshkin, line 37: Using builtin topolgy templating results in a deadlock because the .ControlPlane is not updated until after the lifecycle hook returns.

jimmidyson:
> @jimmidyson the link just brings me to the PR, mind sharing which file its pointing to?

Hmm the link works fine for me. Let's try again - https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1255/files#diff-366132233cc4268f6ebd520d434ccb6de1a5f7b204b5ffbe6988806345af9e4eR1-R50

Looking at file `hack/examples/bases/eks/cluster/cluster.yaml` and the content in lines 1-50

pkg/handlers/generic/lifecycle/cni/cilium/template.go
```diff
@@ -0,0 +1,42 @@
...
+	// Assume when kube-proxy is skipped, we should enable Cilium's kube-proxy replacement feature.
+	templateInput := input{
+		EnableKubeProxyReplacement: capiutils.ShouldSkipKubeProxy(cluster),
+	}
+
+	var b bytes.Buffer
+	err = ciliumTemplate.Execute(&b, templateInput)
+	if err != nil {
+		return "", fmt.Errorf(
+			"`failed setting target Cluster name and namespa`ce in template: %w",
```
jimmidyson, line 36: nit
```suggestion
			"failed setting target Cluster name and namespace in template: %w",
```

pkg/handlers/generic/lifecycle/cni/cilium/handler.go
```diff
@@ -259,65 +261,145 @@ func runApply(
...
 		},
 	}
+	if err := c.Get(ctx, ctrlclient.ObjectKeyFromObject(ds), ds); err != nil {
+		return fmt.Errorf("failed to get cilium daemon set: %w", err)
+	}
+
+	// Update the DaemonSet to force a rollout.
+	annotations := ds.Spec.Template.Annotations
+	if annotations == nil {
+		annotations = make(map[string]string)
```
jimmidyson, line 376: nit
```suggestion
		annotations = make(map[string]string, 1)
```

pkg/handlers/generic/lifecycle/cni/cilium/handler.go
```diff
@@ -259,65 +261,145 @@ func runApply(
...
+	}
+	if _, ok := annotations[restartedAtAnnotation]; !ok {
+		// Only set the annotation once to avoid a race conditition where rollouts are triggered repeatedly.
+		annotations[restartedAtAnnotation] = time.Now().UTC().Format(time.RFC3339Nano)
```
jimmidyson, line 380: Not sure we need nano precision. Let's do the same as [`kubectl rollout restart`](https://github.com/kubernetes/kubernetes/blob/v1.34.1/staging/src/k8s.io/kubectl/pkg/polymorphichelpers/objectrestarter.go#L41).
```suggestion
		annotations[restartedAtAnnotation] = time.Now().UTC().Format(time.RFC3339)
```

test-deployment-service.yaml
```diff
@@ -0,0 +1,39 @@
+apiVersion: apps/v1
+kind: Deployment
```
supershal, line 2: is this file checked in by accident?

supershal (commented):
LGTM. thanks

supershal (approved):

test-deployment-service.yaml
dkoshkin: Yep, thanks

pkg/handlers/generic/lifecycle/cni/cilium/handler.go
```diff
@@ -259,65 +261,145 @@ func runApply(
...
+		client,
+		ctrlclient.ObjectKeyFromObject(cluster),
+	)
+	if err != nil {
+		return fmt.Errorf("error creating remote cluster client: %w", err)
+	}
+
+	// If kube-proxy is not installed,
+	// assume that the one-time migration of kube-proxy is complete and return early.
+	isKubeProxyInstalled, err := isKubeProxyInstalled(ctx, remoteClient)
```
dlipovetsky, line 284: More of a note to self: When `cleanupKubeProxy` returns, all we know is that the API server has set the deletion timestamp on the resources. They may still be around the next time the handler is called. Then, `isKubeProxyInstalled` will return true, and we'll end up running the remainder of the code, including `cleanupKubeProxy`, again. But since `forceCiliumRollout` only sets the "restartedAt" annotation once,  this is safe.

pkg/handlers/generic/lifecycle/cni/cilium/handler.go
```diff
@@ -259,65 +261,145 @@ func runApply(
...
+
+	// Update the DaemonSet to force a rollout.
+	annotations := ds.Spec.Template.Annotations
+	if annotations == nil {
+		annotations = make(map[string]string, 1)
+	}
+	if _, ok := annotations[restartedAtAnnotation]; !ok {
+		// Only set the annotation once to avoid a race conditition where rollouts are triggered repeatedly.
+		annotations[restartedAtAnnotation] = time.Now().UTC().Format(time.RFC3339)
+	}
```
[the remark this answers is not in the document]
jimmidyson: Correct. Let's figure this out when we need to.

PR #1309, opened 2025-09-17
feat: add volume APIs to AWS and EKS Nodes

dkoshkin:
Exposes CAPA's [rootVolume](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/88cb4b92b1a76591623e9d5ef347bfdc22010622/api/v1beta1/types.go#L184) and [nonRootVolumes](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/88cb4b92b1a76591623e9d5ef347bfdc22010622/api/v1beta1/types.go#L188C33-L188C47) APIs in CAREN API.
I did not use the capav1.Volume type directly so that we can have proper kubebuilder annotations and used basic types instead of pointers that the CAPA uses (inconsistently).

Fixes #

Unit tests and also created an EKS cluster with 2 Nodepools
Defaults with default root volume
image

Second nodepool with a modified root and additional volume
image


api/v1alpha1/aws_node_types.go
```diff
@@ -50,6 +54,10 @@ type AWSGenericNodeSpec struct {
 	// PlacementGroup specifies the placement group in which to launch the instance.
 	// +kubebuilder:validation:Optional
 	PlacementGroup *PlacementGroup `json:"placementGroup,omitempty"`
+
+	// Configuration options for the root storage volume.
+	// +kubebuilder:validation:Optional
+	RootVolume *Volume `json:"rootVolume,omitempty"`
```
jimmidyson, line 60: Could we add non-root volumes too please? We've had conversations previously about adding separate disk for image storage, logs, etc to prevent filling up of root disk and killing nodes.

How about something like:

```suggestion
	Volumes *AWSVolumes `json:"volumes,omitempty"`
```

And `AWSVolumes` defined as:

```
type AWSVolumes struct {
	// Configuration options for the root storage volume.
	Root Volume `json:"root"`

	// Configuration options for non-root storage volumes.
	NonRoot []Volume `json:"nonroot,omitempty"`
}
```

Happy to leave the non-root configurations to a follow up PR, but would like to move root volume under `volumes.root` anyway for extensibility. WDYT?

dkoshkin: Great idea, I'll make this change.

dkoshkin: Agreed @faiq, I left it optional and let CAPA do its defaults.

api/v1alpha1/aws_node_types.go
```diff
@@ -105,3 +113,48 @@ type AMILookup struct {
 	// +kubebuilder:validation:MaxLength=32
 	BaseOS string `json:"baseOS,omitempty"`
 }
+
+// +kubebuilder:validation:XValidation:rule="has(self.root) || size(self.nonroot) > 0",message="either root or nonroot must be specified"
```
jimmidyson, line 117: Thinking about this it seems redundant. It's not that if one of these isn't specified then something will break so neither is strictly required. This just adds processing for no actual functional benefit.
```suggestion
```

dkoshkin: 👍 Im good with reverting the commit too

PR #1317, opened 2025-09-24
refactor: Use metallb APIs instead of unstructured

jimmidyson:
Also refactors the apply logic to return early on anything other than a conflict. Other errors are not necessarily retryable and hence returning early is better practice.

I decided to do this while I was thinking about introducing the Cilium load-balancer and looking at using vendored Cilium API types instead of unstructured - doing this for metallb was nice and simple.

Fixes #


make/apis.mk
```diff
@@ -30,11 +30,14 @@ PROVIDER_API_PATHS_caaph := api/v1alpha1
...
 
+PROVIDER_MODULE_metallb := go.universe.tf/metallb
+PROVIDER_API_PATHS_metallb := api/v1beta1 api/v1beta2
+
 # Add third-party CAPI provider types above
 
 .PHONY: apis.sync
 apis.sync: ## Syncs third-party CAPI providers' types
-apis.sync: $(addprefix api.sync.,capa caaph capx) mod-tidy.api go-fix.api
+apis.sync: $(addprefix api.sync.,capa caaph capx metallb) mod-tidy.api go-fix.api
```
dkoshkin, line 40: Im liking this external API mechanism!

dkoshkin (approved):
Thanks for the clear errors on conflicts!

make/apis.mk
jimmidyson: It's really great! Until you try to sync cilium APIs which I'm doing now and they have massive API surface and loads of package dependencies 😞

dkoshkin (approved):

pkg/handlers/generic/lifecycle/serviceloadbalancer/metallb/handler.go
```diff
@@ -162,37 +162,48 @@ func (n *MetalLB) Apply(
...
-								"%w. This resource has been modified in the workload cluster, it must only contain the %q IP Address Pool", //nolint:lll // Long error message,
-								err,
-								configInput.Name,
-							)
-						}
+					// Return early if the error is not a conflict or an internal error.
+					// Internal errors are generally seen when the necessary CRD webhooks are not yet registered so are
+					// retryable.
+					if !apierrors.IsConflict(err) && !apierrors.IsInternalError(err) {
+						return false, err
```
dlipovetsky, line 169: I think there's a slight issue with the handling, if I read it correctly: the handler reports a "conflict" when there is either a conflict, or an internal error.

So, how about we handle the internal errors, then non-conflict errors, and finally the conflict errors? 

```suggestion
					if apierrors.IsInternalError(err) {
						// Retry on internal errors as these are generally seen when the necessary
						// CRD webhooks are not yet registered.
						return false, nil
					}

					// Return early if the error is not a conflict.
					if !apierrors.IsConflict(err) {
						return false, err
					}

					// At this point, we have handled both internal and non-conflict errors,
					// so we must be dealing with a conflict.
```

jimmidyson: Following linter recommendation, refactored this to a switch statement.

dkoshkin (approved):

PR #1322, opened 2025-09-26
ci: Upgrade devbox install action and remove custom golangci-lint

jimmidyson:

The latest devbox-install-action removes the need for custom nix installation on self-hosted runners.

Also the nixhub.io indexing has been fixed upstream so can remove our custom installation of golangci-lint and use upstream discovered version.

Fixes #


dkoshkin (approved):
Thanks

PR #1323, opened 2025-09-27
refactor: Stricter EKS kube-proxy mode in Go API

jimmidyson:

Use Go types to restrict valid values rather than post-generation edits. This does not change the external schema for Kubernetes API clients, but provides stricter Go API semantics.

Fixes #


api/v1alpha1/clusterconfig_types.go
```diff
@@ -220,7 +236,7 @@ type EKSClusterConfigSpec struct {
 
 	// KubeProxy defines the configuration for kube-proxy.
 	// +kubebuilder:validation:Optional
-	KubeProxy *KubeProxy `json:"kubeProxy,omitempty"`
+	KubeProxy *EKSKubeProxy `json:"kubeProxy,omitempty"`
```
dkoshkin, line 239: Will this work? https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/82a4976bf68bd8aed1d00c0fa5bc5b174a5a5c10/api/variables/json.go#L16
with the wrong type https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/82a4976bf68bd8aed1d00c0fa5bc5b174a5a5c10/api/variables/aggregate_types.go#L28

jimmidyson: Right, and I'm pretty sure `EKSKubeProxy` is only ever going to be a strict subset of `KubeProxy`.

Go clients would understand the Go API which defines the custom type for `EKSKubeProxy` so I don't think this is a problem.

dkoshkin (approved):

supershal (approved):
Great way to separate out without affecting the API.

PR #1330, opened 2025-09-29
build: downgrade clusterctl version

dkoshkin:
Newer version fails when running an older CAPI version: 
```
$ clusterctl get kubeconfig $CLUSTER_NAME
Error: this version of clusterctl could be used only with "v1beta2" management clusters, "v1beta1" detected
```

Can switch to latest once https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1325 is merged

Fixes #

### DIFF
--- a/devbox.json
+++ b/devbox.json
@@ -2,7 +2,7 @@
   "packages": {
     "actionlint":                           "latest",
     "chart-testing":                        "latest",
-    "clusterctl":                           "latest",
+    "clusterctl":                           "1.10.4",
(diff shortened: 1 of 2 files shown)


dkoshkin:
> @dkoshkin Where do you hit this error? Why don't we see this in CI?

I ran `clusterctl get kubeconfig`. I don't think we run `clusterctl` directly in the e2e tests.

mhrabovcin (approved):

PR #1420, opened 2025-12-02
feat: Audit Cluster API resources at the request level

dlipovetsky:
These are critical resources, and changes to their data should be included in the audit log. Currently, only changes to their metadata is included.

Fixes #

I will also update the handler version, as we have previously done, e.g. in #1391.

Before I do that, I'd like to see if there are other improvements we can make the policy.

And perhaps it's time to make it configurable.


jimmidyson (commented):
Can we make the audit policy configurable too?

dlipovetsky:
I'm making the audit policy configurable in #1564. After that merges, I'll update this PR, so that the default policy audits Cluster API resources at the request level.

PR #1564, opened 2026-04-04
feat: Configurable audit log backend and policy

dlipovetsky:
Allows user to configure the audit log backend, and policy.

Adds types to the cluster configuration that mirror the relevant kube-apiserver flags.

Introduces the auditlog handler. This handler patches the control plane kubeadm configuration template with flags, and  the audit policy, based on the cluster configuration.

The audit log configuration is handled by a cluster variable.

The audit policy is read from a referenced ConfigMap. If no ConfigMap is referenced, the default policy, which is embedded, is used.

Fixes #


dlipovetsky:
Currently waiting on #1606.

PR #1575, opened 2026-04-09
fix: properly handle v1beta1 types in webhooks

dkoshkin:
By setting apiVersions: v1beta2 in the WebhookConfiguration, the API server will do the conversion before sending the request.

See https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-matchpolicy

Doing this while v1beta1 in a transitional period and is not yet removed.

Fixes #

Tested by creating a v1beta1 cluster and no longer seeing

```
Warning: cluster.x-k8s.io/v1beta1 Cluster is deprecated; use cluster.x-k8s.io/v1beta2 Cluster
Error from server: error when creating "/Users/dimitri.koshkin/Downloads/input-v1b1.json": admission webhook "addons-defaulter.caren.nutanix.com" denied the request: no kind "Cluster" is registered for version "cluster.x-k8s.io/v1beta1" in scheme "pkg/runtime/scheme.go:110"

Warning: cluster.x-k8s.io/v1beta1 Cluster is deprecated; use cluster.x-k8s.io/v1beta2 Cluster
Error from server: error when creating "/Users/dimitri.koshkin/Downloads/input-v1b1.json": admission webhook "addons-defaulter.caren.nutanix.com" denied the request: unable to decode cluster.x-k8s.io/v1beta1, Kind=Cluster into *v1beta2.Cluster
```

### DIFF
--- a/charts/cluster-api-runtime-extensions-nutanix/templates/webhooks.yaml
+++ b/charts/cluster-api-runtime-extensions-nutanix/templates/webhooks.yaml
@@ -14,14 +14,14 @@ webhooks:
       service:
         name: '{{ include "chart.name" . }}-admission'
         namespace: '{{ .Release.Namespace }}'
-        path: /mutate-v1beta1-addons
+        path: /mutate-v1beta2-addons
     failurePolicy: Fail
     name: addons-defaulter.caren.nutanix.com
     rules:
       - apiGroups:
           - cluster.x-k8s.io
         apiVersions:
-          - '*'
+          - v1beta2
         operations:
           - CREATE
         resources:
@@ -36,14 +36,14 @@ webhooks:
       service:
         name: '{{ include "chart.name" . }}-admission'
         namespace: '{{ .Release.Namespace }}'
-        path: /mutate-v1beta1-cluster
+        path: /mutate-v1beta2-cluster
     failurePolicy: Fail
     name: cluster-defaulter.caren.nutanix.com
     rules:
       - apiGroups:
           - cluster.x-k8s.io
         apiVersions:
-          - '*'
+          - v1beta2
         operations:
           - CREATE
           - UPDATE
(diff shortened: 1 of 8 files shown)


faiq (dismissed):
nice change! this more accurately represents the types we care about

guillermoz-nutanix (approved):

charts/cluster-api-runtime-extensions-nutanix/templates/webhooks.yaml
```diff
@@ -14,14 +14,14 @@ webhooks:
       service:
         name: '{{ include "chart.name" . }}-admission'
         namespace: '{{ .Release.Namespace }}'
-        path: /mutate-v1beta1-addons
+        path: /mutate-v1beta2-addons
```
jimmidyson: Let's drop the version from the path.

charts/cluster-api-runtime-extensions-nutanix/templates/webhooks.yaml
```diff
@@ -14,14 +14,14 @@ webhooks:
...
     failurePolicy: Fail
     name: addons-defaulter.caren.nutanix.com
     rules:
       - apiGroups:
           - cluster.x-k8s.io
         apiVersions:
-          - '*'
+          - v1beta2
```
jimmidyson, line 24: Let's explicitly add `matchPolicy: Equivalent` throughout -even though it's default, being explicit makes it a little clearer what's expected, and let's add a comment too.

charts/cluster-api-runtime-extensions-nutanix/templates/webhooks.yaml
dkoshkin: SG, done in c2aa5cdf918446597a10ae005a346ed9ac0a8e5b

charts/cluster-api-runtime-extensions-nutanix/templates/webhooks.yaml
dkoshkin: 👍 548aca09788e60df20eef78eeae978a81021d312

charts/cluster-api-runtime-extensions-nutanix/templates/webhooks.yaml
faiq: why? isn't it better to be more explicit?

dkoshkin: Changing the name causes churn and temporary downtime of the webhook during an upgrade where the API version changes, since the updated `MutatingWebhookConfiguration` is already deployed but the already running webhook Pod still expects the older name.

I think the explicitness comes from setting the version in `apiVersions`, and if we ever need different endpoints based on the API we can make that the exception.

jimmidyson (approved):

atulv7 (approved):

PR #1592, opened 2026-04-14
feat: [NCN-113518] Add enforceNodeAllocatable field

jimmidyson:

CAREN exposes `systemReserved` and `kubeReserved` in the `KubeletConfiguration` API, but
without `enforceNodeAllocatable` these reservations only affect scheduling — nothing
prevents system daemons or the kubelet from consuming more than their declared reservation
at runtime. This PR adds an opt-in `enforceNodeAllocatable` field that enables
cgroup-based enforcement.

When `system-reserved` or `kube-reserved` (or their `-compressible` variants) are
specified, CAREN automatically injects the well-known systemd cgroup paths (`/system.slice`
and `/system.slice/kubelet.service`) so users never need to understand cgroup internals.

Fixes https://jira.nutanix.com/browse/NCN-113518

- 8 unit tests for template rendering covering: pods-only, all values sorted,
  system-reserved cgroup injection, kube-reserved cgroup injection,
  system-reserved-compressible, kube-reserved-compressible, mixed compressible/full,
  and empty/nil no-op.
- Integration patch tests for both worker and control plane handlers.
- Variable validation tests pass for AWS, Docker, and Nutanix providers, including
  CEL mutual exclusivity rejection tests.
- CRD YAMLs verified to contain `enforceNodeAllocatable` with 5-value enum constraint,
  `maxItems: 3`, `uniqueItems: true`, and field-level CEL validation rules.

- The change is fully opt-in: when the field is not set, no new patches are emitted,
  making this a no-op for existing clusters. No handler version bump is required.
- Five accepted values: `pods`, `system-reserved`, `kube-reserved`,
  `system-reserved-compressible`, `kube-reserved-compressible`.
- The `-compressible` variants (K8s v1.32+) enforce only CPU and are the recommended
  starting point. `system-reserved` and `system-reserved-compressible` are mutually
  exclusive (same for `kube-reserved`), enforced via field-level CEL validation.
- MaxItems stays at 3 because mutual exclusivity prevents more than one system variant,
  one kube variant, and `pods`.
- Values are sorted alphabetically before rendering for idempotent output.
- Applies to both control plane and worker nodes. EKS is excluded (no kubeadm).
- Review the spec at `.specify/features/NCN-113518/spec.md` for full design rationale.


pkg/handlers/generic/mutation/kubeadm/kubeletconfiguration/kubelet_config.go
```diff
@@ -107,6 +111,21 @@ func toTemplateInput(cfg *v1alpha1.KubeletConfiguration) *kubeletConfigTemplateI
 		}
 	}
 
+	if len(cfg.EnforceNodeAllocatable) > 0 {
```
dkoshkin, line 116: I think the answer is yes, but not setting a value still leaves the [default pod enforcement](https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/#enforcing-node-allocatable)?

docs/content/customization/kubeadm/kubelet-configuration.md
```diff
@@ -76,3 +77,41 @@ spec:
...
+      - name: clusterConfig
+        value:
+          controlPlane:
+            kubeletConfiguration:
+              systemReserved:
+                cpu: "500m"
+                memory: "1Gi"
+              kubeReserved:
+                cpu: "200m"
+                memory: "512Mi"
```
dkoshkin, line 190: What does it mean to be "advisory " when `systemReserved` and `kubeReserved` are set without `enforceNodeAllocatable`? Should the API force to set both? Or maybe automatically set when `Reserved` are set?

TBH I would expect when I set this to be enforced without any additional API.
```
              systemReserved:
                cpu: "500m"
                memory: "1Gi"
              kubeReserved:
                cpu: "200m"
                memory: "512Mi"
```

pkg/handlers/generic/mutation/kubeadm/kubeletconfiguration/kubelet_config.go
jimmidyson: Correct.

docs/content/customization/kubeadm/kubelet-configuration.md
jimmidyson: I've updated the language to express what it means more clearly.

jimmidyson: > TBH I would expect when I set this to be enforced without any additional API.

Enforcing is risky and so by default these values are only used for scheduling, not for enforcement. Enforcement is opt-in.

dkoshkin: Ahh its used during scheduling, got it.

dkoshkin (approved):

supershal (approved):
LGTM

PR #1593, opened 2026-04-14
docs: Improve kubelet configuration documentation

jimmidyson:

The kubelet configuration documentation was a bare list of field names with minimal
examples. This PR adds a reference table with types, constraints, and descriptions for
every field (including the compressible enforcement variants), plus focused examples for
nested/map configurations: resource reservations, eviction thresholds, graceful node
shutdown, image GC, container log rotation, and NUMA topology management. Upstream docs
are linked rather than duplicated.

Requires #1592

Documentation-only change. Verified markdown renders correctly.

This PR is stacked on top of #1592 which adds the `enforceNodeAllocatable` field. The
base branch is `NCN-113518-enforce-node-allocatable`. After #1592 merges, rebase this
onto `main`.

### DIFF
--- a/docs/content/customization/kubeadm/kubelet-configuration.md
+++ b/docs/content/customization/kubeadm/kubelet-configuration.md
@@ -41,7 +41,7 @@ For full upstream documentation on each setting, see the
 | `shutdownGracePeriod` | duration (e.g. `"30s"`) | Total time the node delays shutdown for pod termination. |
 | `shutdownGracePeriodCriticalPods` | duration (e.g. `"10s"`) | Time reserved for terminating critical pods during shutdown. Must be ≤ `shutdownGracePeriod`. |
 | `seccompDefault` | boolean | Apply the runtime's default seccomp profile (`RuntimeDefault`) to pods that do not specify one. See [Default seccomp profile](#default-seccomp-profile). |
-| `enforceNodeAllocatable` | list of `pods`, `system-reserved`, `kube-reserved` | Which resource reservations are enforced via cgroups. See [Enforce node allocatable](#enforce-node-allocatable). |
+| `enforceNodeAllocatable` | list of `pods`, `system-reserved`, `kube-reserved`, `system-reserved-compressible`, `kube-reserved-compressible` | Which resource reservations are enforced via cgroups. See [Enforce node allocatable](#enforce-node-allocatable). |
 
 ## Default seccomp profile
 
--- a/docs/content/customization/kubeadm/kubelet-configuration.md
+++ b/docs/content/customization/kubeadm/kubelet-configuration.md
@@ -12,30 +12,36 @@ Kubelet configuration is supported for:
 
 There is no cluster-level default; control plane and worker settings are configured independently.
 
+All fields are optional. When a field is not set, the kubelet default applies and no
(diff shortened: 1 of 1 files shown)


docs/content/customization/kubeadm/kubelet-configuration.md
```diff
@@ -12,31 +12,37 @@ Kubelet configuration is supported for:
...
-- `containerLogMaxFiles`
-- `imageGCHighThresholdPercent`
-- `imageGCLowThresholdPercent`
-- `maxParallelImagePulls`
-- `shutdownGracePeriod`
-- `shutdownGracePeriodCriticalPods`
-- `enforceNodeAllocatable`
-
-## Example
+| Field | Type | Description |
```
faiq, line 23: We shouldn't put all of these in here and instead just link out to the relevant upstream configuration.

jimmidyson: We do link out [above](https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1593/changes#diff-239bf4b375f68c67fea5e67ee394e2a3746116b6546140491556c916375a47c1R19), but I felt like this very high level doc was worth it, especially expressing things like `Must be > imageGCLowThresholdPercent`.

dlipovetsky: Our documentation may go out of sync with the upstream in the future. However, our documentation is tied to our Go types, and if upstream changes, we'll presumably need to update the types, and then we'll update our documentation.

dlipovetsky: Slightly off-topic: Do we want to define CEL rules for our types, to reflect what's documented? Or are we relying on upstream rules rejecting our patch?

jimmidyson: > Do we want to define CEL rules for our types, to reflect what's documented?

We already have validation in place for kubelet config fields - see https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/main/api/v1alpha1/kubelet_types.go#L38-L179

docs/content/customization/kubeadm/kubelet-configuration.md
```diff
@@ -12,30 +12,36 @@ Kubelet configuration is supported for:
...
-- `containerLogMaxFiles`
-- `imageGCHighThresholdPercent`
-- `imageGCLowThresholdPercent`
-- `maxParallelImagePulls`
-- `shutdownGracePeriod`
-- `shutdownGracePeriodCriticalPods`
-- `seccompDefault`
-- `enforceNodeAllocatable`
+| Field | Type | Description |
+|-------|------|-------------|
```
dkoshkin, line 24: Assuming these are accurate and generated from an upstream doc.

dkoshkin (dismissed):
Thanks!

docs/content/customization/kubeadm/kubelet-configuration.md
jimmidyson: They are.

supershal (approved):
Thanks.

thunderboltsid (approved):

PR #1598, opened 2026-04-17
feat: [NCN-110043] Add Cilium as a ServiceLoadBalancer provider

jimmidyson:
## Summary

Adds [Cilium](https://cilium.io) as an alternative ServiceLoadBalancer provider alongside MetalLB.

- Extends `spec.topology.variables.clusterConfig.addons.serviceLoadBalancer.provider` enum from `{MetalLB}` to `{MetalLB, Cilium}`.
- New lifecycle sub-provider applies a `CiliumLoadBalancerIPPool` and `CiliumL2AnnouncementPolicy` (from `cilium.io/v2` and `cilium.io/v2alpha1`) to each workload cluster via server-side apply.
- Admission validator rejects `provider=Cilium` unless the cluster uses the Cilium CNI AND has kube-proxy disabled, and blocks provider changes on live clusters (delete-and-recreate is required).
- Cilium CNI values template conditionally enables `l2announcements` and raises `k8sClientRateLimit.qps/burst` when the cluster picks Cilium as its SLB provider — following [upstream guidance for L2 announcements](https://docs.cilium.io/en/stable/network/l2-announcements/).
- Nutanix PC-IP-in-LoadBalancer-range preflight check generalised to any SLB provider (no longer MetalLB-specific).

Jira: [NCN-110043](https://jira.nutanix.com/browse/NCN-110043)

## Vendoring strategy

Cilium's `pkg/k8s/apis/cilium.io/v2` is a kitchen-sink package that would drag ~30 internal packages into `api/external/` if synced whole. Instead, a bespoke `api.sync.cilium` make target syncs only the two type files we need and rewrites `slimv1.LabelSelector` (from Cilium's slim apimachinery fork) to `metav1.LabelSelector`. The types are wire-compatible; upstream's own `zz_generated.deepcopy.go` already uses `metav1`.

## Handler-version safety

Topology-mutation handler output is unchanged (Task 13 verified). All new behaviour is in lifecycle handlers, webhooks, and the Cilium CNI values template. No handler version bump required per `.cursor/rules/handler-version-safety.mdc`.

## Files of note

- `api/external/github.com/cilium/cilium/...` — vendored Cilium CRDs (two type files + register/doc + deepcopy)
- `make/apis.mk` — custom `api.sync.cilium` target
- `pkg/handlers/lifecycle/serviceloadbalancer/cilium/` — new sub-provider
- `pkg/webhook/cluster/cilium_loadbalancer_validator.go` — admission validator
- `pkg/webhook/cluster/nutanix_validator.go` — generalised PC-IP check
- `charts/cluster-api-runtime-extensions-nutanix/addons/cni/cilium/values-template.yaml` — conditional l2announcements
- `docs/content/addons/serviceloadbalancer.md` — user-facing docs with MetalLB vs Cilium comparison

## Test plan

- [x] Unit tests: `go test ./api/... ./pkg/...` (including envtest for admission webhooks and handler scaffolding)
- [x] Handler patch-output parity check (Task 13)
- [x] Pre-commit hooks (golangci-lint, gitlint, apis-sync, etc.) green
- [x] E2E waiter wired into `WaitForServiceLoadBalancerToBeReadyInWorkloadCluster`
- [ ] Manual e2e on Nutanix with `provider=Cilium` cluster template (requires follow-up fixture, see below)

## Deliberately deferred (follow-up PR)

- Cilium-SLB cluster fixture: `hack/examples/patches/nutanix/cilium-slb.yaml` + kustomize flavor + `caren.yaml` variables. Once this lands, the existing quick-start spec will exercise the Cilium SLB path automatically — no further test-code changes needed.

## Spec & plan

- `.specify/features/NCN-110043/spec.md`
- `.specify/features/NCN-110043/plan.md`


pkg/handlers/lifecycle/cni/cilium/template.go
```diff
@@ -21,6 +22,14 @@ func templateValues(cluster *clusterv1.Cluster, text string) (string, error) {
 		return "", fmt.Errorf("failed to check if kube-proxy is disabled: %w", err)
 	}
 
+	spec, err := apivariables.UnmarshalClusterConfigVariable(cluster.Spec.Topology.Variables)
+	if err != nil {
+		return "", fmt.Errorf("failed to unmarshal cluster config: %w", err)
+	}
+	slbIsCilium := spec != nil && spec.Addons != nil &&
```
dkoshkin: nit: I haven't seen "slb" and was a little confused what it meant. Im fine with it though it we want to start using it in the repo.
```suggestion
	serviceLBIsCilium := spec != nil && spec.Addons != nil &&
```

pkg/webhook/cluster/cilium_loadbalancer_validator.go
```diff
@@ -0,0 +1,109 @@
...
+		return admission.Errored(http.StatusBadRequest, err)
+	}
+	if !cluster.Spec.Topology.IsDefined() {
+		return admission.Allowed("")
+	}
+
+	if cluster.Annotations[v1alpha1.PreflightChecksSkipAnnotationKey] ==
+		v1alpha1.PreflightChecksSkipAllAnnotationValue {
+		return admission.Allowed("")
+	}
```
dkoshkin: Technically this isn't a preflight right? I'm also not sure if we would actually want to skip based on an annotation.
```suggestion
```

pkg/webhook/cluster/cilium_loadbalancer_validator.go
```diff
@@ -0,0 +1,109 @@
...
+		oldCfg, oldErr := variables.UnmarshalClusterConfigVariable(oldCluster.Spec.Topology.Variables)
+		if oldErr == nil && oldCfg != nil &&
+			oldCfg.Addons != nil && oldCfg.Addons.ServiceLoadBalancer != nil {
+			newProvider := ""
+			if cfg != nil && cfg.Addons != nil && cfg.Addons.ServiceLoadBalancer != nil {
+				newProvider = cfg.Addons.ServiceLoadBalancer.Provider
+			}
+			if oldCfg.Addons.ServiceLoadBalancer.Provider != newProvider {
+				return admission.Denied(fmt.Sprintf(
+					"ServiceLoadBalancer provider change from %q to %q is not supported; "+
```
dkoshkin, line 74: Could this be a CEL on the API instead?

test/e2e/serviceloadbalancer_helpers.go
```diff
@@ -27,6 +27,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/api/v1alpha1"
+	slbcilium "github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pkg/handlers/lifecycle/serviceloadbalancer/cilium"
```
dkoshkin: Should the changes here be part of https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1599 instead?

pkg/handlers/lifecycle/cni/cilium/template.go
jimmidyson: Renamed to `serviceLBIsCilium` in 4b8455ed.

pkg/webhook/cluster/cilium_loadbalancer_validator.go
jimmidyson: Agreed — this is a validating webhook, not a preflight check, so reusing the preflight-skip annotation conflated two unrelated subsystems, and the checks it guards are hard correctness invariants that shouldn't be bypassable. Removed the annotation bypass (and its unit test case) in 4b8455ed.

pkg/webhook/cluster/cilium_loadbalancer_validator.go
jimmidyson: Feasible, but more involved than a single field rule, so I've left it as a webhook for now and would suggest a follow-up if we want to retire the webhook.

There's good precedent for `self == oldSelf` immutability on these topology variables (`KubeProxy.Mode`, `EKSKubeProxy.Mode`), so a CEL rule on `ServiceLoadBalancer.Provider` would cover the common case. The caveat: a field-level `self == oldSelf` only fires when `serviceLoadBalancer` is present in both old and new, so it wouldn't catch removing the whole `serviceLoadBalancer` block on update (which this validator currently rejects via the empty-new-provider path). To fully replace the webhook you'd need a transition rule at the `addons`/struct level rather than on the field.

The other two checks here are also cross-field (Cilium SLB requires Cilium CNI, and requires kube-proxy disabled), so moving everything to CEL would mean rules at the `clusterConfig`/`addons` level too. Net: doable and arguably cleaner, but a bigger change than this PR — happy to do it as a follow-up if you'd prefer.

test/e2e/serviceloadbalancer_helpers.go
jimmidyson: Moved to the e2e fixture PR (#1599): removed here in 4b8455ed and re-added alongside the fixtures in 23fdfa7e.

PR #1605, opened 2026-04-22
fix: require Prism Central 7.5 in preflight

thunderboltsid:
## Summary
- raise the CAREN Prism Central preflight minimum from 7.3 to 7.5
- update the Prism Central version tests so 7.3.x is rejected and 7.5.x remains allowed
- align the public preflight gate with support expectations based on API call usage

## Test plan
- [x] `go test ./pkg/webhook/preflight/nutanix/...`

Made with [Cursor](https://cursor.com)


abhay-nutanix (approved):
LGTM

kshcore (approved):

PR #1606, opened 2026-04-24
fix: Update CoreDNS mapping file

dlipovetsky:
The release of Kubernetes v1.36.0 a couple of days ago made the current mapping out of date. 

Fixes #

### DIFF
--- a/api/versions/coredns.go
+++ b/api/versions/coredns.go
@@ -28,6 +28,7 @@ const (
 	Kubernetes_V1_33 = "v1.33"
 	Kubernetes_V1_34 = "v1.34"
 	Kubernetes_V1_35 = "v1.35"
+	Kubernetes_V1_36 = "v1.36"
 )
 
 // CoreDNS versions.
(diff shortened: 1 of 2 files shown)


dlipovetsky:
The pre-commit fails if the mapping is out of date. To update it, we run `make coredns.sync`. However, that does not update the hard-coded mapping in the unit test.

I think this test is overspecified, i.e. there's no point in verifying the mapping with a test, if we already verify it as part of the build steps. If we insist on verifying the mapping with a test, then let's have the make target update the mapping in the test, too.

@thunderboltsid WDYT?

thunderboltsid:
Thanks, Daniel. I checked the code and agree with your read.

`make coredns.sync` only regenerates `api/versions/coredns.go`, while `api/versions/coredns_test.go` duplicates the generated map by hand. That makes the full-map assertion brittle for every mapping update.

I would keep the behavior coverage around `GetCoreDNSVersion` parsing/lookup and maybe add/keep a test that `GetKubernetesToCoreDNSVersionMap()` returns a copy, but I do not think the unit test should assert the entire generated mapping. So I am good with removing that full-map assertion rather than teaching `coredns.sync` to update the test too.

thunderboltsid (approved):
Looks good. Thanks for updating the CoreDNS mapping and removing the duplicated generated mapping assertion.

atulv7 (approved):

PR #1615, opened 2026-05-01
feat: Preflight check for control plane endpoint

dlipovetsky:
This check makes sure the control plane endpoint is not already being used before a new cluster is created. When no control plane nodes exist yet, it tries to open a TCP connection to the endpoint host and port. If the connection succeeds, that likely means something else (possibly another cluster) is already using that endpoint, so the check fails to stop the new cluster from being created. If the connection fails, or if control plane nodes already exist, the check passes.

Fixes #

### DIFF
--- a/pkg/webhook/preflight/nutanix/controlplane.go
+++ b/pkg/webhook/preflight/nutanix/controlplane.go
@@ -1,4 +1,4 @@
-// Copyright 2024 Nutanix. All rights reserved.
+// Copyright 2026 Nutanix. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 package nutanix
@@ -36,16 +36,22 @@ func (c *controlPlaneEndpointCheck) Name() string {
 }
 
 func (c *controlPlaneEndpointCheck) Run(ctx context.Context) preflight.CheckResult {
-	result := preflight.CheckResult{
-		Allowed: true,
+	if !c.cluster.Spec.ControlPlaneRef.IsDefined() {
+		// The control plane reference is defined, so the control plane has been initialized,
+		// and the endpoint may establish a TCP connection.
+		return preflight.CheckResult{
+			Allowed: true,
+		}
 	}
 
-	if !c.cluster.Spec.ControlPlaneRef.IsDefined() {
-		result.Allowed = true
-		return result
-	}
+	// The control plane reference is not defined, so the control plane has not been initialized,
+	// and if the endpoint establishes a TCP connection, then the most likely conclusion is that the
+	// endpoint is in use by another cluster. It is possible that the endpoint is a load balancer or
+	// proxy that accepts connections even if the control plane is not initialized, but in that
+	// case, the user can skip this check.
 
 	dialer := &net.Dialer{
+		// The default TCP timeout is 60 seconds, which is too long for this check.
 		Timeout: 3 * time.Second,
 	}
 	addr := net.JoinHostPort(
@@ -54,20 +60,22 @@ func (c *controlPlaneEndpointCheck) Run(ctx context.Context) preflight.CheckResu
 	)
 	conn, err := dialer.DialContext(ctx, "tcp", addr)
 	if err != nil {
-		result.Allowed = true
-		return result
-	}
-	if conn != nil {
-		conn.Close()
-		result.Allowed = false
-		result.Causes = append(result.Causes, preflight.Cause{
-			//nolint:lll // Message is long.
-			Message: fmt.Sprintf(
-				"The control plane endpoint %s established a TCP connection, before any control plane nodes have been created. The endpoint is in use, possibly by another cluster.",
-				addr,
-			),
-		})
+		return preflight.CheckResult{
+			Allowed: true,
+		}
 	}
 
-	return result
+	conn.Close()
+	return preflight.CheckResult{
+		Allowed: false,
+		Causes: []preflight.Cause{
+			{
+				//nolint:lll // Message is long.
+				Message: fmt.Sprintf(
+					"The control plane endpoint %s established a TCP connection, before any control plane nodes have been created. The endpoint is in use, possibly by another cluster.",
+					addr,
+				),
+			},
+		},
+	}
 }
--- a/pkg/webhook/preflight/nutanix/controlplane.go
+++ b/pkg/webhook/preflight/nutanix/controlplane.go
@@ -27,6 +27,10 @@ func newControlPlaneEndpointChecks(
 	}
 }
 
+// controlPlaneEndpointCheck is a check that verifies that the control plane
+// endpoint is not in use, before any control plane nodes have been created. If
+// the endpoint is in use, the check should fail, to prevent the cluster from
+// being created.
 type controlPlaneEndpointCheck struct {
 	cluster *clusterv1.Cluster
 }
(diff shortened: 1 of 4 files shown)


faiq (approved):

pkg/webhook/preflight/nutanix/controlplane.go
```diff
@@ -0,0 +1,73 @@
...
+	return "NutanixControlPlaneEndpoint"
+}
+
+func (c *controlPlaneEndpointCheck) Run(ctx context.Context) preflight.CheckResult {
+	result := preflight.CheckResult{
+		Allowed: true,
+	}
+
+	if !c.cluster.Spec.ControlPlaneRef.IsDefined() {
+		result.Allowed = true
```
jimmidyson: This is redundant.

pkg/webhook/preflight/nutanix/controlplane.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+func (c *controlPlaneEndpointCheck) Name() string {
+	return "NutanixControlPlaneEndpoint"
+}
+
+func (c *controlPlaneEndpointCheck) Run(ctx context.Context) preflight.CheckResult {
+	result := preflight.CheckResult{
+		Allowed: true,
+	}
```
jimmidyson: It seems odd to initialize this here rather than just return what we want at the various returns in this function. There's no aggregation happening here so let's remove this and return appropriately at various return points.

pkg/webhook/preflight/nutanix/controlplane.go
```diff
@@ -0,0 +1,73 @@
...
+		result.Causes = append(result.Causes, preflight.Cause{
+			//nolint:lll // Message is long.
+			Message: fmt.Sprintf(
+				"The control plane endpoint %s established a TCP connection, before any control plane nodes have been created. The endpoint is in use, possibly by another cluster.",
+				addr,
+			),
+		})
+	}
+
+	return result
```
jimmidyson: AFAIK `conn` will be non-nil if `err` is nil. The code can be simpler:

```suggestion
	defer conn.Close()
	return preflight.CheckResult{
		Allowed: false,
		Causes: []preflight.Cause{{
			//nolint:lll // Message is long.
			Message: fmt.Sprintf(
				"The control plane endpoint %s established a TCP connection, before any control plane nodes have been created. The endpoint is in use, possibly by another cluster.",
				addr,
			),
		}},
	}
```

pkg/webhook/preflight/nutanix/controlplane.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	dialer := &net.Dialer{
+		Timeout: 3 * time.Second,
+	}
+	addr := net.JoinHostPort(
+		c.cluster.Spec.ControlPlaneEndpoint.Host,
+		strconv.Itoa(int(c.cluster.Spec.ControlPlaneEndpoint.Port)),
+	)
+	conn, err := dialer.DialContext(ctx, "tcp", addr)
```
jimmidyson, line 73: While connecting to the CP endpoint host:port would show it's in use, it doesn't mean the IP is in use somewhere else already. Trying to think if there is anything else we could do here.

dlipovetsky: I don't understand your comment.

The purpose of this check is not to avoid an IP conflict. It is to alert the user when the endpoint responds to a TCP connection _before_ the control plane has been initialized.

There are edge cases where the endpoint accepts a TCP connection before the control plane has been initialized. For example, the endpoint is a loadbalancer that, for some reason, accepts a TCP connection even before it has any healthy backend targets. In such an edge case, the user should skip the check.

pkg/webhook/preflight/nutanix/controlplane.go
dlipovetsky: Fixed in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1615/commits/93e542b2a2da4813d2f75dee4e759c21fdc44bc1

pkg/webhook/preflight/nutanix/controlplane.go
dlipovetsky: Fixed in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1615/commits/93e542b2a2da4813d2f75dee4e759c21fdc44bc1

pkg/webhook/preflight/nutanix/controlplane.go
dlipovetsky: Good point! Fixed in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1615/commits/93e542b2a2da4813d2f75dee4e759c21fdc44bc1

pkg/webhook/preflight/nutanix/controlplane.go
dlipovetsky: In https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1615/commits/93e542b2a2da4813d2f75dee4e759c21fdc44bc1, I added comments to better explain the purpose of this check.

dlipovetsky:
Rebased on main

pkg/webhook/preflight/nutanix/controlplane.go
dlipovetsky: @jimmidyson WDYT?

jimmidyson: OK i think i get it now, thanks! It doesn't prevent IP reuse, which is possible in a mixed environment where NKP is not the only usage in the infrastrusture, but does attempt to prevent a CP IP for NKP clusters at least.

pkg/webhook/preflight/nutanix/controlplane.go
```diff
@@ -0,0 +1,81 @@
...
+
+type controlPlaneEndpointCheck struct {
+	cluster *clusterv1.Cluster
+}
+
+func (c *controlPlaneEndpointCheck) Name() string {
+	return "NutanixControlPlaneEndpoint"
+}
+
+func (c *controlPlaneEndpointCheck) Run(ctx context.Context) preflight.CheckResult {
```
jimmidyson, line 42: Is this check run on `update` as well? I think it does since we introduced update preflight. If so, this will prevent updates on all clusters and we need to add tests that assert this doesn't prevent updates.

pkg/webhook/preflight/nutanix/controlplane.go
```diff
@@ -0,0 +1,81 @@
...
+	if !c.cluster.Spec.ControlPlaneRef.IsDefined() {
+		// The control plane reference is defined, so the control plane has been initialized,
+		// and the endpoint may establish a TCP connection.
```
jimmidyson: This comment inverts the condition above.

pkg/webhook/preflight/nutanix/controlplane.go
jimmidyson: This should either be a `create` only preflight or check if any CP nodes have been created so it does not prevent updates.

pkg/webhook/preflight/nutanix/controlplane.go
```diff
@@ -0,0 +1,81 @@
...
+
+	dialer := &net.Dialer{
+		// The default TCP timeout is 60 seconds, which is too long for this check.
+		Timeout: 3 * time.Second,
+	}
+	addr := net.JoinHostPort(
+		c.cluster.Spec.ControlPlaneEndpoint.Host,
+		strconv.Itoa(int(c.cluster.Spec.ControlPlaneEndpoint.Port)),
+	)
+	conn, err := dialer.DialContext(ctx, "tcp", addr)
```
jimmidyson, line 73: I wonder if a `GET` to `/healthz/version` would also be useful to assert whether this has been reused as an existing CP VIP?

pkg/webhook/preflight/nutanix/controlplane.go
```diff
@@ -0,0 +1,81 @@
...
+
+	// The control plane reference is not defined, so the control plane has not been initialized,
+	// and if the endpoint establishes a TCP connection, then the most likely conclusion is that the
+	// endpoint is in use by another cluster. It is possible that the endpoint is a load balancer or
+	// proxy that accepts connections even if the control plane is not initialized, but in that
+	// case, the user can skip this check.
+
+	dialer := &net.Dialer{
+		// The default TCP timeout is 60 seconds, which is too long for this check.
+		Timeout: 3 * time.Second,
```
jimmidyson, line 67: I assume this runs in parallel with other checks? It means checks will take a minimum of 3s in the successful (`Allowed`) flow but if in parallel that's not a big deal, if in serial then need to think more to ensure we come in under the 30s timeout.

dlipovetsky: Yes, all checks do run in parallel.

pkg/webhook/preflight/nutanix/controlplane.go
dlipovetsky: Interesting idea! I think either `/healthz` or `/version` would work. We would have to skip TLS verification to make the HTTPS request.

What if something other a Kubernetes cluster responds? I think we want to fail the check in that case, too.

In general, if no control plane machines exist, I think we should not be able to form a TCP connection to the endpoint VIP.

pkg/webhook/preflight/nutanix/controlplane.go
dlipovetsky: Yes, the condition is wrong! :see_no_evil:

pkg/webhook/preflight/nutanix/controlplane.go
dlipovetsky: I had the condition wrong here. If there is a control plane reference, then the cluster has already been reconciled by CAPX, which has started to create machines, so it is too late to run this check.

The idea is that we can catch an issue before the cluster is ever reconciled by CAPX.

dlipovetsky:
I had to rebase on main to resolve conflicts.

dlipovetsky:
The first implementation was incorrect. I fixed that, and added a test case to simulate a TCP timeout, so now all paths are covered. I also updated the PR description.

pkg/webhook/preflight/nutanix/controlplane.go
dlipovetsky: The check runs on update, but passes once the control plane reference is defined. The topology cluster controller [sets this field](https://github.com/kubernetes-sigs/cluster-api/blob/b64483027d15466f02215c8a6d3df8a3e3422c2f/exp/topology/desiredstate/desired_state.go#L751). Once the field is defined, the endpoint is set, and this check isn't relevant anymore.

dkoshkin (approved):
Thanks for the detailed comments.

pkg/webhook/preflight/nutanix/controlplane.go
dlipovetsky: This idea might be useful to give additional context to the error, e.g. to say something like "the control plane endpoint responds, and it appears to be a Kubernetes API." The check is still useful without it, but we could add it in the future.

PR #1620, opened 2026-05-04
ci: Upload e2e test artifacts

dlipovetsky:
To triage flaky e2e tests, we need to upload the artifacts, so that we can review them. The test logs alone do not have enough information. For example, we can't figure out why a VM is not created within a timeout.

Fixes #

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -26,10 +26,6 @@ on:
         description: The OS image to use for the machine template
         type: string
         required: false
-      job-name:
-        description: Display name of the calling job, used to name the uploaded artifact
-        type: string
-        required: true
 
 jobs:
   e2e-test:
@@ -122,6 +118,6 @@ jobs:
         name: Upload test artifacts
         uses: actions/upload-artifact@v4
         with:
-          name: "${{ github.workflow }} - ${{ inputs.job-name }}.zip"
+          name: "e2e-test-artifacts"
           path: ${{ env.ARTIFACTS }}
           retention-days: 7
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -29,6 +29,9 @@ on:
 
 jobs:
   e2e-test:
+    env:
+      # ARTIFACTS must be an absolute path, so we use the github.workspace variable.
+      ARTIFACTS: ${{ github.workspace }}/_artifacts
     runs-on: ${{ inputs.runs-on }}
     permissions:
       contents: read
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e.yml
```diff
@@ -110,3 +110,11 @@ jobs:
           check_name: 'e2e test report'
           detailed_summary: true
           require_passed_tests: true
+
+      - name: Upload test artifacts
+        if: success() || failure() # always run even if the previous step fails
```
supershal: there is an `always()` function that we can use.

supershal:
We can also at least dump the cluster manifest to recreate the issue. Its hard to find what machine details provided in the test (i.e. image name or image lookup) and its correct or not.

.github/workflows/e2e.yml
dlipovetsky: I used `!cancelled()` based on the discussion in https://github.com/orgs/community/discussions/25789, so that the steps always runs, unless the workflow is cancelled.

dlipovetsky:
One artifacts is published for each job in the matrix, but each job uses the same name for the artifact, making them impossible to distinguish. I'll fix that.

dlipovetsky:
image

supershal (commented):
can we compose the job name using inputs.focus, inputs.provider and inputs.kubernetesVersion ?

dlipovetsky:
> can we compose the job name using inputs.focus, inputs.provider and inputs.kubernetesVersion ?

Great idea! Done in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1620/commits/9cb848f5ec29525a2d93fefa50ec2fb48fba2f4d

supershal (approved):
Thanks. This is helpful to figure out failures offline instead of scrolling through huge logs lines in GHA.

I did not noticed any secret data being revealed in the logs or resources. Please double check if I missed anything.

dlipovetsky:
image

Reflects updated artifact names

dkoshkin:
How does this work for OSS repos? Any potential to leak creds?

jimmidyson:
> How does this work for OSS repos? Any potential to leak creds?

Yes! Please don't use GitHub artifacts for this. We used to do this until @thunderboltsid noticed it!

dlipovetsky:
> > How does this work for OSS repos? Any potential to leak creds?
> 
> Yes! Please don't use GitHub artifacts for this. We used to do this until @thunderboltsid noticed it!

As a precaution, we can use artifacts and require maintainer approval before actions can run against a PR.

If that's not enough, then alternative do you have in mind?

We currently have zero information to help triage failing e2e tests, so we need something.

PR #1625, opened 2026-05-16
feat: [NCN-114349] Label addon namespaces with privileged PSA

jimmidyson:
## Summary

- Three lifecycle handlers (Nutanix CSI driver, Konnector Agent, NFD Helm
  strategy) now ensure their workload-cluster namespace exists with
  `pod-security.kubernetes.io/enforce=privileged` before installing the Helm
  chart, mirroring the existing MetalLB pattern. Without this, the privileged
  workloads in `ntnx-system` and `node-feature-discovery` are rejected at
  admission once an operator enables baseline or stricter Pod Security
  Admission on the workload cluster.
- A new shared `PrivilegedPodSecurityEnforceLabels` constant in
  `pkg/handlers/utils` ensures the three handlers agree on the label set. The
  PSA enforce-version label is intentionally omitted because PSA defaults to
  ``latest`` when no enforce-version is set.
- The NFD `ClusterResourceSet` strategy is intentionally unchanged — the CRS
  ConfigMap already bakes the privileged label into the namespace manifest.
  The Nutanix CCM handler and Nutanix Flow CNI handler are also not in scope
  because both install into `kube-system` (the documented PSA exemption) and
  do not touch ``ntnx-system``.
- Lifecycle handlers only — no topology mutation handler patch output changes,
  so no handler version bump is required (constitution principle VII).

Jira: [NCN-114349](https://jira.nutanix.com/browse/NCN-114349)

## Test plan

- [x] ``go build ./...`` clean.
- [x] New ``It`` block in
      ``pkg/handlers/utils/utils_integration_test.go`` proves the label is
      applied via SSA, idempotently on repeat reconciles, and added in place
      to a pre-existing label-less namespace; ``enforce-version`` is never set.
- [x] ``go test ./pkg/handlers/utils ./pkg/handlers/lifecycle/csi/... ./pkg/handlers/lifecycle/konnectoragent/... ./pkg/handlers/lifecycle/nfd/...``
      green locally (envtest 1.29.5).
- [x] ``git diff --name-only origin/main | grep mutation/`` → empty
      (verifies no mutation-handler change).
- [ ] Manual smoke on a workload cluster: install Nutanix CSI + Konnector
      Agent + NFD, enable ``clusterConfig.podSecurityAdmission.enforce: baseline``,
      and confirm ``kubectl get ns ntnx-system node-feature-discovery -o jsonpath='{range .items[*]}{.metadata.name}{": "}{.metadata.labels.pod-security\\.kubernetes\\.io/enforce}{\"\\n\"}{end}'``
      prints ``privileged`` for each and that the addon pods stay healthy.

### DIFF
--- a/.specify/features/NCN-114349/plan.md
+++ b/.specify/features/NCN-114349/plan.md
@@ -1,0 +1,237 @@
+<!--
+ Copyright 2026 Nutanix. All rights reserved.
+ SPDX-License-Identifier: Apache-2.0
+-->
+
+# Implementation Plan: Privileged PSA label on ntnx-system and node-feature-discovery namespaces
+
+**Branch**: `NCN-114349-psa-label-ntnx-system-nfd-namespaces`
+**Date**: 2026-05-14
+**Spec**: [./spec.md](./spec.md)
+
+## Summary
+
+Three lifecycle handlers must ensure a workload-cluster namespace exists with
+`pod-security.kubernetes.io/enforce=privileged` *before* installing their
+Helm-based addons. The required behaviour and helper already exist in
+`pkg/handlers/utils.EnsureNamespaceWithMetadata`, and the MetalLB lifecycle
+handler is the canonical template
+(`pkg/handlers/lifecycle/serviceloadbalancer/metallb/handler.go`). We follow
+that template in each of the three affected handlers and add a tiny shared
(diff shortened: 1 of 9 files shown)


thunderboltsid (approved):

dkoshkin (approved):
Nice generic approach over messing with helm values for individual addons.

PR #1627, opened 2026-05-18
fix: Surface Cluster topology patch decode failures

Vedansh074:
Previously, when a patch selector matched a template but strict decoding into the typed view failed, MutateIfApplicable only logged the issue at V(5) and returned nil. As a result, the mutator was silently skipped and the original unstructured object was returned without any patches applied, with no visible indication at normal log levels.

I encountered this while creating a cluster where the capx-controller version was newer and introduced the NutanixMachineTemplateStatus field. Since CAREN’s vendored CAPX types had not yet been updated, strict decoding rejected the object due to the unknown field. Because of this, the NutanixMachineTemplate mutators silently failed to patch subnets and images, leading to broken cluster configuration without surfacing any clear error.

This change propagates the conversion error instead of suppressing it. As a result, the meta GeneratePatches handler now logs these failures at Error level, making future provider API drift immediately visible instead of silently producing partially patched or broken clusters.

I tested this on the mentioned  setup and was able to see the error clearly:

`0518 14:03:22.602010    1 meta.go:123] "Mutator failed" err="failed to convert unstructured object to typed object: strict decoding error: unknown field \"status\"" template="infrastructure.cluster.x-k8s.io/v1beta1/NutanixMachineTemplate" holder="controlplane.cluster.x-k8s.io/v1beta2/KubeadmControlPlane/default/nkp-vedansh-caren-test-2-rcm7v" holderRef={"apiVersion":"controlplane.cluster.x-k8s.io/v1beta2","kind":"KubeadmControlPlane","namespace":"default","name":"nkp-vedansh-caren-test-2-rcm7v","fieldPath":"spec.machineTemplate.spec.infrastructureRef"} objectKind="infrastructure.cluster.x-k8s.io/v1beta1, Kind=NutanixMachineTemplate" handlerName="nutanixClusterv6configpatch" index=3 handler="*machinedetails.nutanixMachineDetailsPatchHandler"

`

### DIFF
--- a/common/pkg/capi/clustertopology/patches/generator.go
+++ b/common/pkg/capi/clustertopology/patches/generator.go
@@ -55,7 +55,7 @@ func MutateIfApplicable[T runtime.Object](
 			"objKind", obj.GetObjectKind().GroupVersionKind(),
 			"expectedType", fmt.Sprintf("%T", &typed),
 		).Info("not matching type")
-		return nil
+		return fmt.Errorf("failed to convert unstructured object to typed object: %w", err)
 	}
 
 	// Create a deep clone of the typed object.
--- a/common/pkg/capi/clustertopology/patches/generator.go
+++ b/common/pkg/capi/clustertopology/patches/generator.go
@@ -55,7 +55,11 @@ func MutateIfApplicable[T runtime.Object](
 			"objKind", obj.GetObjectKind().GroupVersionKind(),
 			"expectedType", fmt.Sprintf("%T", &typed),
 		).Info("not matching type")
-		return fmt.Errorf("failed to convert unstructured object to typed object: %w", err)
+		return fmt.Errorf(
+			"failed to convert unstructured object (%s) to typed object: %w",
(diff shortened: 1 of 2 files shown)


common/pkg/capi/clustertopology/patches/generator.go
```diff
@@ -55,7 +55,7 @@ func MutateIfApplicable[T runtime.Object](
 			"objKind", obj.GetObjectKind().GroupVersionKind(),
 			"expectedType", fmt.Sprintf("%T", &typed),
 		).Info("not matching type")
-		return nil
+		return fmt.Errorf("failed to convert unstructured object to typed object: %w", err)
```
jimmidyson: Let's print the GVK of the unstructured object in this message too.

Vedansh074: Done

jimmidyson (approved):

thunderboltsid (approved):

PR #1631, opened 2026-05-28
build: add a extra SAN in Kind cluster

dkoshkin:
Useful when deploying this KIND cluster with a reverse tunnel. This extra SAN can then point back to a bastion host.

Fixes #

### DIFF
--- a/hack/kind/kind-base-config.yaml
+++ b/hack/kind/kind-base-config.yaml
@@ -16,6 +16,14 @@ kubeadmConfigPatches:
     apiVersion: kubelet.config.k8s.io/v1beta1
     kind: KubeletConfiguration
     nodeStatusMaxImages: -1
+  - |
+    kind: ClusterConfiguration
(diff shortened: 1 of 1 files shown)


hack/kind/kind-base-config.yaml
```diff
@@ -16,6 +16,14 @@ kubeadmConfigPatches:
     apiVersion: kubelet.config.k8s.io/v1beta1
     kind: KubeletConfiguration
     nodeStatusMaxImages: -1
+  - |
+    kind: ClusterConfiguration
+    apiVersion: kubeadm.k8s.io/v1beta3
+    apiServer:
+      certSANs:
+        - "localhost"
+        - "127.0.0.1"
```
dkoshkin, line 25: These are here by default in a KIND cluster, but need to be set explicitly when passing `apiServer.certSANs` configs.

jimmidyson (approved):

legacyrj (approved):

PR #1632, opened 2026-05-28
test: Verify preflight check when cluster is both paused and unpaused

dlipovetsky:
The existing test did not cover the case where the cluster was being unpaused, i.e. when it transitions from being paused, to not being paused.

The code was correct, and did not change. The coverage was incomplete, and is now 100%.

Fixes #


dkoshkin (approved):

dlipovetsky:
Rebased on main and force-pushed.

legacyrj (approved):

dlipovetsky:
Rebased on main.

PR #1633, opened 2026-05-28
fix: Summarize preflight failures in result message

dlipovetsky:
We want individual check failures to reach users. The message field is usually propagated by API clients, whereas the details field is not, so we now summarize the information in the details field in the message field.

Fixes #


dkoshkin (approved):
Thanks!

faiq (approved):

PR #1225, opened 2025-07-21
feat: Add more context to MetalLB config apply conflict errors

dlipovetsky:
**What problem does this PR solve?**:
When there is a conflict writing the MetalLB configuration, we want to give more information to the user.

**Which issue(s) this PR fixes**:
Fixes https://jira.nutanix.com/browse/NCN-107584

**How Has This Been Tested?**:

**Special notes for your reviewer**:

PR #1643, opened 2026-06-08
fix: return metalLB config apply conflict errors

Vedansh074:
For MetalLB configuration, we want the Cluster resource to remain the single source of truth, meaning any manual changes made via kubectl edit should be reconciled back to the state defined in the Cluster object. To make this behavior clearer, https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1225/ introduced more explicit user-facing messages.

However, we were also retrying on conflict errors. Since these conflicts cannot be resolved through retries alone, the retry loop continued until the Cluster API webhook request eventually timed out.

This PR makes the fix to directly return the conflicting error which will be shown instead of generic timeout that happened earlier.

Fixes # https://jira.nutanix.com/browse/NCN-114970

### DIFF
--- a/pkg/handlers/lifecycle/serviceloadbalancer/metallb/handler.go
+++ b/pkg/handlers/lifecycle/serviceloadbalancer/metallb/handler.go
@@ -193,8 +193,8 @@ func (n *MetalLB) Apply(
 						err,
 					)
 
-					// Return false with no error to retry the apply.
-					return false, nil
+					// Conflicts won't resolve by retrying; return the error immediately.
+					return false, applyErr
 				default:
 					// Otherwise return the error early and do not retry.
 					return false, err


thunderboltsid (approved):

thunderboltsid:
Please run this by @dlipovetsky for the second stamp.

dlipovetsky:
Thanks for fixing this!

It's important that all handlers return either success or an error within the timeout, otherwise the caller (Cluster API topology controller) reports only that "context deadline exceeded," which gives no useful information to use the user as to why the handler failed.

I'd like to find a generic way to enforce this across handlers. That's probably cancelling the context passed to the handler _before_ the timeout, so that there's time for handlers to handle the cancelled context.

 I _think_ that, if we cancel the context with time to spare, this particular handler will return `applyErr`.

dlipovetsky (approved):

PR #1644, opened 2026-06-09
fix: remediate CVEs by upgrading golang.org/x/crypto and x/net 

shubham2g:
fix: remediate CVEs by upgrading golang.org/x/crypto and x/net dependencies


jimmidyson:
Not saying we shouldn't do this, but the linked JIRA NCN-114540 (inferred from branch name), it doesn't show that CAREN is vulnerable. Can you link to a report that says it is? We run govulncheck and there are no failures for that.

thunderboltsid (approved):

thunderboltsid:
@shubham2g you might have to amend your commit to make pre-commit happy

Vedansh074 (approved):

PR #1649, opened 2026-06-10
ci: [NCN-115081] Sign and attest release images and Helm chart

jimmidyson:
## Summary

- Add keyless (Sigstore/Fulcio/Rekor via GitHub Actions OIDC) cosign signatures and SLSA build-provenance attestations to the release pipeline for both container images (`cluster-api-runtime-extensions-nutanix`, `cluster-api-runtime-extensions-helm-chart-bundle-initializer`) and the Helm chart. No long-lived keys or secrets.
- Publish the Helm chart to `ghcr.io` as an OCI artifact at the flat, image-distinct path `cluster-api-runtime-extensions-nutanix-chart` using `oras` with Helm OCI media types. `Chart.yaml`, the existing `.tgz` release asset, and the gh-pages index are unchanged; the chart remains `helm pull`-able.
- Signing/attestation run as workflow steps after the goreleaser release so the `ko`-built and buildx-built images are handled uniformly; digests are resolved from the registry via `crane`. Signing runs under `set -euo pipefail` so an unsigned release cannot be published silently.
- Add `cosign` and `oras` to devbox.

Jira: [NCN-115081](https://jira.nutanix.com/browse/NCN-115081)

## Test plan

- [x] `actionlint` passes on the workflow.
- [x] `cosign`/`oras` install via devbox; pre-commit (gitlint, actionlint) green.
- [x] Local `oras push` -> registry -> `helm pull` roundtrip verified; `crane digest` resolves the manifest; pulled chart `Chart.yaml` name/version intact.
- [ ] Post-release verification on a `workflow_dispatch`/tag run: `cosign verify` and `gh attestation verify` succeed for both images and the OCI chart, and `helm pull oci://...` works.

## Notes

- cosign installs as v3 (keyless is the default; no behavior change).
- Consumer-facing verification docs are not included in this PR.


.github/workflows/release-tag.yml
```diff
@@ -62,12 +64,80 @@ jobs:
       - name: Package Helm chart and attach to release
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          VERSION: ${{ github.ref_name }}
```
dkoshkin: nit: could this be specified once at the top level?

dkoshkin (approved):

.github/workflows/release-tag.yml
jimmidyson: Done.

PR #1650, opened 2026-06-11
feat: [NCN-115115] Add opt-in automatic kubelet resource reservations

jimmidyson:
Jira: [NCN-115115](https://jira.nutanix.com/browse/NCN-115115)

## Summary

- Adds an opt-in `kubeletConfiguration.automaticReservations.profile: CapacityTiered` directive (control plane and workers). When set, each node computes `kubeReserved` (CPU + memory) and a `memory.available: 100Mi` hard-eviction threshold **at boot** from its actual capacity, via an embedded POSIX script run through a `preKubeadmCommand`.
- The capacity-tiered formula mirrors GKE's node-allocatable model (EKS uses the same approach): a larger reserved fraction on small nodes, shrinking as the node grows.
- Mutually exclusive with explicit `systemReserved`, `kubeReserved`, and `evictionHard` — enforced declaratively (CEL on the CRD) and by the admission webhook with a clear message. Other kubelet fields (e.g. `maxPods`) can still be combined with it.
- Opt-in and a no-op when unset: default ClusterClasses are unchanged and existing handler patch output is byte-identical for inputs without the field, so no handler version bump and no rollouts on upgrade.

## Spec / Plan

- `.specify/features/NCN-115115/spec.md`
- `.specify/features/NCN-115115/plan.md`

## Test Plan

- [x] Script unit test covers the tiered formula across node sizes (1c/512Mi → 60m/255Mi, 2c/8Gi → 70m/1843Mi, 8c/16Gi → 90m/2662Mi, 16c/64Gi → 110m/5611Mi, 128c/512Gi → 390m/17407Mi)
- [x] Handler tests: worker + control plane inject the script file and `preKubeadmCommand` when enabled; no script injected when unset
- [x] Webhook tests: rejects auto + explicit reservations; accepts auto alone
- [x] `make lint` clean (root/api/common); affected packages pass; no `go-generate` drift; default ClusterClasses unchanged vs `main`
- [ ] e2e / manual smoke: confirm a workload-cluster node writes `/etc/kubernetes/patches/kubeletconfiguration50+strategic.json` with size-appropriate `kubeReserved` and reduced allocatable


test/framework/bootstrap/kindprovider.go
```diff
@@ -90,3 +91,151 @@ func CreateKindBootstrapClusterAndLoadImages(
...
+		options = append(options, bootstrap.WithDualStackFamily())
+	}
+	if input.LogFolder != "" {
+		options = append(options, bootstrap.LogFolder(input.LogFolder))
+	}
+	return bootstrap.NewKindClusterProvider(input.Name, options...)
+}
+
+// dockerSocketHostPath resolves the host path of the Docker socket to mount into the bootstrap
+// cluster. It honours CAREN_E2E_DOCKER_SOCKET first, then a unix:// DOCKER_HOST (the standard
```
dkoshkin, line 138: >It honours

Nice! 🏴󠁧󠁢󠁥󠁮󠁧󠁿

dkoshkin (approved):
Great improvement!

pkg/handlers/generic/mutation/kubeadm/kubeletconfiguration/embedded/compute-reservations.sh
```diff
@@ -0,0 +1,80 @@
...
+mkdir -p "${patch_dir}"
+cat >"${patch_file}" <<EOF
+---
+apiVersion: kubelet.config.k8s.io/v1beta1
+kind: KubeletConfiguration
+kubeReserved:
+  cpu: "${cpu_milli}m"
+  memory: "${mem_mib}Mi"
+evictionHard:
+  memory.available: "100Mi"
```
thunderboltsid, line 79: The eviction-hard threshold is hardcoded to 100Mi regardless of node size. is this intentional? rather than something that should scale with node memory tier, since a 100Mi hard threshold is proportionally tiny on very large-memory nodes.

thunderboltsid (approved):

pkg/handlers/generic/mutation/kubeadm/kubeletconfiguration/embedded/compute-reservations.sh
jimmidyson: This is intentional, it doesn't need to scale with node size. It's a safety valve to hard kill pods if this limit is hit and should leave enough overhead for kernel/kubelet to reclaim, literally just enough to handle pod eviction. This is consistent with [EKS default](https://github.com/awslabs/amazon-eks-ami/blob/main/nodeadm/internal/kubelet/config.go#L111) and [GKE](https://docs.cloud.google.com/kubernetes-engine/docs/concepts/plan-node-sizes) btw and this is also manually overridable.

PR #1665, opened 2026-07-01
fix: Adding mpio flag for metro clusters - csi

vijayaraghavanr31:
Nutanix CSI multipath I/O (applyMpioConfigs) must be enabled on metro/VHA clusters to handle the multi-path storage access across AHV sites. 

What does this PR change?

Two files changed:

charts/.../addons/csi/nutanix/values-template.yaml - adds applyMpioConfigs: {{ .ApplyMpioConfigs }} to the CSI Helm values template so the flag can be dynamically rendered per cluster.

pkg/handlers/lifecycle/csi/nutanix/handler.go - introduces a templateValuesFunc that is wired into the CSI addon via WithValueTemplater. It:

Inspects the cluster's ClusterConfigSpec variables and MachineDeployment failure domains
Detects metro topology by checking if any failure domain name starts with NutanixMetro/ or NutanixMetroSite/
Sets ApplyMpioConfigs: true in the rendered Helm values if a metro failure domain is found; false otherwise
Result: CSI is automatically deployed with MPIO enabled on metro clusters and disabled on standard single-site clusters - no manual override needed.

Fixes #

Tested locally with kind cluster

CSI HCP:
`valuesTemplate: |-
    # Disable creating the Prism Central credentials Secret, the Secret will be created by the handler.
    createPrismCentralSecret: false
    # Disable creating the Prism Element credentials Secret, it won't be used the CSI driver as configured here.
    createSecret: false
    pcSecretName: nutanix-csi-credentials
    applyMpioConfigs: true

    tolerations:
      - key: CriticalAddonsOnly
        operator: Exists
      - effect: NoExecute
        operator: Exists
        tolerationSeconds: 300
      - effect: NoSchedule
        operator: Exists`

### DIFF
--- a/charts/cluster-api-runtime-extensions-nutanix/addons/csi/nutanix/values-template.yaml
+++ b/charts/cluster-api-runtime-extensions-nutanix/addons/csi/nutanix/values-template.yaml
@@ -3,6 +3,7 @@ createPrismCentralSecret: false
 # Disable creating the Prism Element credentials Secret, it won't be used the CSI driver as configured here.
 createSecret: false
 pcSecretName: nutanix-csi-credentials
+applyMpioConfigs: {{ .ApplyMpioConfigs }}
 
 tolerations:
   - key: CriticalAddonsOnly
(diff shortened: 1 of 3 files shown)


abhay-nutanix (approved):

yanhua121 (approved):
The code looks good to me.

PR #1671, opened 2026-07-13
fix: restore CIS/STIG extraArgs lost in CAPX v1.10 upgrade

vijayaraghavanr31:
The CAPX v1.9 -> v1.10 bump changed kubeadm extraArgs from a map[string]string (v1beta3) to []Arg (v1beta4). The regenerated nutanix-cluster-class.yaml silently dropped 9 security-hardening args across apiServer, controllerManager, and scheduler.
The newly created nutanix-specific CIS/STIG patch files used strategic merge patch format, which replaces the entire extraArgs array instead of appending to it. This caused each subsequent patch to wipe the entries set by the previous one, leaving only the STIG tls-min-version entry in the final output.

Fix by:
- Restoring the dropped args in nutanix-cluster-class.yaml
- Converting CIS/STIG control-plane-flags-nutanix.yaml to JSON patch (op: add /-) so entries are appended, not replaced
- Adding extraargs-nutanix.yaml initialize patch to create the empty apiServer/extraArgs array before JSON patches append to it


examples/capi-quick-start/nutanix-cluster-calico-crs.yaml
```diff
@@ -35,15 +35,16 @@ metadata:
...
     [
       {
         "type": "basic_auth",
         "data": {
           "prismCentral":{
             "username": "${NUTANIX_USER}",
             "password": "${NUTANIX_PASSWORD}"
-          }
+          },
+          "prismElements": null
```
dkoshkin, line 47: Curious why this is changing?

hack/examples/overlays/clusterclasses/nutanix/kustomization.yaml.tmpl
```diff
@@ -36,6 +36,9 @@ patches:
   - target:
       kind: KubeadmControlPlaneTemplate
     path: ../../../patches/initialize/kubeadmcontrolplanetemplate/extravolumes.yaml
+  - target:
+      kind: KubeadmControlPlaneTemplate
+    path: ../../../patches/initialize/kubeadmcontrolplanetemplate/extraargs-nutanix.yaml
```
dkoshkin: Maybe?
```suggestion
    path: ../../../patches/initialize/kubeadmcontrolplanetemplate/extraargs.yaml
```

hack/examples/patches/initialize/kubeadmcontrolplanetemplate/extraargs-nutanix.yaml
```diff
@@ -0,0 +1,14 @@
+# Copyright 2026 Nutanix. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# Initialize apiServer/extraArgs as an empty array.
+# Required because kubeadm v1beta4 (CAPX v1.10+) uses []Arg (a list) rather than
+# map[string]string for extraArgs, so JSON patch op:add with path /- must find an
+# existing array. The extravolumes.yaml initialize patch replaces apiServer with {}
```
dkoshkin: It kinda seems that this is kinda a bug and should not be overriding if it already exists? Any way to have kustomize only add it if nil? https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/d4908ee164022fb665933b3cf7cfa8e43daeb205/hack/examples/patches/initialize/kubeadmcontrolplanetemplate/extravolumes.yaml#L4-L6 

Its a little strange because we now have an oder dependency between these 2 files.
Maybe we should just rename `extravolumes.yaml` to something like `api-server.yaml`?

examples/capi-quick-start/nutanix-cluster-calico-crs.yaml
vijayaraghavanr31: These example files are generated by running make generate-examples which pulls CAPX's cluster-template-clusterclass.yaml as a kustomize base resource. The prismElements: null changes originate from CAPX's own kustomize overlay sources - specifically templates/overlays/base/basic-auth/kustomization.yaml introduced in [CAPX PR #747](https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pull/747). The CAREN example files had drifted from this since they were last regenerated.

dkoshkin (approved):

examples/capi-quick-start/nutanix-cluster-calico-crs.yaml
thunderboltsid: We should address this in CAPX too, this change seems unnecessary.

thunderboltsid (approved):

PR #1673, opened 2026-07-14
feat: add Rtt and witness guardrails

abhay-nutanix:
## Summary

Adds two new Nutanix metro preflight guardrails that run during Cluster
admission, alongside the existing single-metro and Prism-Element-scale checks:

- **Prism Central hosting guardrail** (`prismCentralMetroHostingCheck`): rejects
  a metro Cluster when Prism Central resides on either of the two Prism Elements
  used for metro. If PC lived on a metro PE, a single PE failure would take down
  Prism Central together with half of the metro.
- **Replication latency guardrail** (`metroReplicationLatencyCheck`): rejects a
  metro Cluster when the network round-trip time (RTT) between the two metro
  Prism Elements exceeds `5ms`, the supported bound for safe synchronous
  replication.

Both checks are only added when the Cluster references a metro configuration.

## Details

- `metroMaxRTTMillis = 5.0` defines the supported RTT bound.
- Fail-open behaviour: when the hosting PE or the RTT cannot be determined
  (e.g. not synchronous-replication capable), the checks do **not** block —
  there is no evidence of a violation. Transient API failures surface as
  retryable internal errors.
- The RTT check only evaluates when the Cluster resolves to exactly two Prism
  Elements; any other count is reported by the PE-scale check instead.

## Testing

- Unit tests added/updated in `metro_test.go` covering allow/deny/undetermined/
  error paths for both checks, and updated expected check counts.
- Verified live on a metro cluster (`abhay-test`) by deploying a branch image:
  - Valid `metro0` config is **allowed** (PC on a non-metro PE; RTT `0.15ms`).
  - A crafted metro whose PEs include the PC-hosting PE is **denied** with the
    PC-hosting message.
- `go build`, `gofmt`, `golangci-lint` (0 issues), and package unit tests pass.

## Notes

- No topology mutation handler patch output changes, so no handler version bump
  is required (per handler-version-safety guidance).

## Testing Doc

- https://docs.google.com/document/d/1hVUcrWwzWtvTgc0te4JQ45HSjoZcCvX7oR0xSBTs-GU/edit?usp=sharing


thunderboltsid:
This will impact the set of privileges required by least privilege users.
cc @manoj-nutanix @shivendra-ntnx

thunderboltsid (approved):

vijayaraghavanr31 (approved):

PR #1675, opened 2026-07-14
feat: Enforce all node pools use metro failure domains

abhay-nutanix:
- When a Cluster uses metro (references at least one `NutanixMetro`/
  `NutanixMetroSite` failure domain), a new preflight check now requires
  **every** Control Plane and Worker node pool to be placed on a
  `NutanixMetro`/`NutanixMetroSite` failure domain.
- Rejects mixed configurations that would leave some node pools without
  synchronous-replication protection, so a "metro cluster" is fully HA at the
  infrastructure level rather than partially protected.
- Purely structural check (no Prism Central API calls); only runs when metro is
  already in use, so non-metro clusters are unaffected.

## Behavior

- Metro not used anywhere: no change; the check is not emitted.
- Metro used, all node pools on metro/metroSite FDs: allowed.
- Metro used, a node pool on a plain FD: rejected, cause names the node pool and
  the failure domain.
- Metro used, a node pool with no FD: rejected, cause names the node pool.

## Notes

- Policy is **strict**: both Control Plane and Workers must be metro.

### DIFF
--- a/pkg/webhook/preflight/nutanix/metro.go
+++ b/pkg/webhook/preflight/nutanix/metro.go
@@ -135,10 +135,113 @@ func newMetroChecks(cd *checkDependencies) []preflight.Check {
 				nclient:            cd.nclient,
 				errMessage:         errMessage,
 			},
+			// Every Control Plane and Worker node pool must be placed on a
+			// metro failure domain: otherwise only some node pools would be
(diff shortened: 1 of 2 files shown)


yanhua121 (approved):
Looks good to me.

thunderboltsid (approved):

PR #1677, opened 2026-07-20
feat: dockerhub nutanix - Don't trim org prefix for dockerhub nutanix, always wire credential provider

nilay-ntnx:

- Do not trim org prefix for Dockerhub Nutanix registry. This is so that it can co-exist with customer's existing registry (which could also be docker.io).
- Always wire credential provider (Dynamic-credential-provider). This is to remove an edge case where dynamic-credential-provider may not be configured and we can't configure dockerhub creds during upgrade. 


nilay-ntnx:
Splitting this PR into two - 
- https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1678
- https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1679

PR #1678, opened 2026-07-23
feat: Dockerhub Integration - Don't trim org prefix for Dockerhub Nutanix registry

nilay-ntnx:
Do not trim org prefix for Dockerhub Nutanix registry. This is so that it can co-exist with customer's existing registry (which could also be docker.io).

### DIFF
--- a/pkg/handlers/v6/generic/mutation/generic/imageregistries/credentials/credentials_secret.go
+++ b/pkg/handlers/v6/generic/mutation/generic/imageregistries/credentials/credentials_secret.go
@@ -106,16 +106,26 @@ func kubeletStaticCredentialProviderSecretContents(configs []providerConfig) (st
 			return "", fmt.Errorf("failed parsing registry URL: %w", err)
 		}
 
+		// Org-scoped keying (host + path, e.g. registry-1.docker.io/nutanix) is
+		// applied ONLY to NKP's own docker.io/nutanix org; every other registry
+		// keeps the original host-only key to limit blast radius.
+		trimmedPath := strings.TrimRight(registryURL.Path, "/")
+		pathPrefix := ""
+		if isDockerHubNutanixOrgURL(registryURL.Host, trimmedPath) {
+			pathPrefix = trimmedPath
+		}
+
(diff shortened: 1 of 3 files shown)


pkg/handlers/v6/generic/mutation/generic/imageregistries/credentials/credentials_secret.go
```diff
@@ -106,16 +106,26 @@ func kubeletStaticCredentialProviderSecretContents(configs []providerConfig) (st
 			return "", fmt.Errorf("failed parsing registry URL: %w", err)
 		}
 
+		// Org-scoped keying (host + path, e.g. registry-1.docker.io/nutanix) is
```
dkoshkin: We don't want to change this handler.
We need to this (probably all in a single PR)
1. Bump the handler versions, similar to https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1684 but `pkg/handlers/v5/` should only contain this `imageregistries` handler. 
2. Update this handler to reflect what is currently in main in `pkg/handlers/generic/mutation/generic/imageregistries/credentials/credentials_secret.go` 
3. Make the change in `pkg/handlers/generic/mutation/generic/imageregistries/credentials/credentials_secret.go` as you have in this PR.
4. Include the changes from https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1679/changes 

This is all needed to prevent workload clusters from getting reconciled after a mgmt cluster upgrade and then new patches getting generated with the new `imageregistries` handler logic.

jimmidyson:
I wonder if we should actually not trim all specified registries paths here 🤔

dkoshkin:
> I wonder if we should actually not trim all specified registries paths here 🤔

We talked that too @jimmidyson, the thinking was to minimize changes and potential regressions on existing customers.
We can always revisit that in the future once we can do more testing to validate it.

dkoshkin (approved):
Thanks

jimmidyson (approved):

PR #1679, opened 2026-07-23
feat: Dockerhub Integration - always wire kubelet credential provider by default

nilay-ntnx:
Always wire credential provider (Dynamic-credential-provider). This is to remove an edge case where dynamic-credential-provider may not be configured and we can't configure dockerhub creds during upgrade.

### DIFF
--- a/api/v1alpha1/constants.go
+++ b/api/v1alpha1/constants.go
@@ -39,6 +39,14 @@ const (
 	GlobalMirrorVariableName = "globalImageRegistryMirror"
 	// ImageRegistriesVariableName is the image registries patch variable name.
 	ImageRegistriesVariableName = "imageRegistries"
+	// WireImageCredentialProviderByDefaultVariableName is the patch variable name for the flag that
+	// wires the kubelet dynamic credential provider on every node even when no registry/mirror is configured.
+	//nolint:gosec // Does not contain hard coded credentials.
+	WireImageCredentialProviderByDefaultVariableName = "wireImageCredentialProviderByDefault"
+
+	// DefaultKubeletCredentialProviderRegistryURL is the registry the credential
+	// provider is wired for by default (Docker Hub) when no registry/mirror is set.
+	DefaultKubeletCredentialProviderRegistryURL = "https://registry-1.docker.io" //nolint:gosec // Not a credential.
 
 	// DNSVariableName is the DNS external patch variable name.
 	DNSVariableName = "dns"
--- a/api/v1alpha1/constants.go
+++ b/api/v1alpha1/constants.go
@@ -39,13 +39,10 @@ const (
 	GlobalMirrorVariableName = "globalImageRegistryMirror"
 	// ImageRegistriesVariableName is the image registries patch variable name.
 	ImageRegistriesVariableName = "imageRegistries"
-	// WireImageCredentialProviderByDefaultVariableName is the patch variable name for the flag that
-	// wires the kubelet dynamic credential provider on every node even when no registry/mirror is configured.
-	//nolint:gosec // Does not contain hard coded credentials.
-	WireImageCredentialProviderByDefaultVariableName = "wireImageCredentialProviderByDefault"
 
-	// DefaultKubeletCredentialProviderRegistryURL is the registry the credential
-	// provider is wired for by default (Docker Hub) when no registry/mirror is set.
+	// DefaultKubeletCredentialProviderRegistryURL is the registry the kubelet
+	// dynamic credential provider is always wired for (Docker Hub) when no
+	// registry/mirror is configured.
 	DefaultKubeletCredentialProviderRegistryURL = "https://registry-1.docker.io" //nolint:gosec // Not a credential.
 
 	// DNSVariableName is the DNS external patch variable name.
--- a/api/v1alpha1/constants.go
+++ b/api/v1alpha1/constants.go
@@ -44,6 +44,7 @@ const (
 	// dynamic credential provider is always wired for (Docker Hub) when no
 	// registry/mirror is configured.
 	DefaultKubeletCredentialProviderRegistryURL = "https://registry-1.docker.io" //nolint:gosec // Not a credential.
+	DockerHubRegistryURL                        = "https://docker.io"            //nolint:gosec // Not a credential.
 
 	// DNSVariableName is the DNS external patch variable name.
 	DNSVariableName = "dns"
--- a/pkg/handlers/generic/mutation/generic/imageregistries/credentials/inject.go
+++ b/pkg/handlers/generic/mutation/generic/imageregistries/credentials/inject.go
@@ -89,9 +89,6 @@ func (h *imageRegistriesPatchHandler) Mutate(
 	)
 
 	switch {
-	case variables.IsNotFoundError(imageRegistriesErr) && variables.IsNotFoundError(globalMirrorErr):
-		log.V(5).Info("Image Registry Credentials and Global Registry Mirror variable not defined")
-		return nil
 	case imageRegistriesErr != nil && !variables.IsNotFoundError(imageRegistriesErr):
 		return imageRegistriesErr
 	case globalMirrorErr != nil && !variables.IsNotFoundError(globalMirrorErr):
@@ -139,8 +136,23 @@ func (h *imageRegistriesPatchHandler) Mutate(
 		return err
 	}
 	if len(registriesThatNeedConfiguration) == 0 {
-		log.V(5).Info("Image registry credentials are not needed")
-		return nil
+		// Nothing the user configured needs a credential provider. Unless the
+		// cluster opts out, still wire the kubelet dynamic credential provider for
+		// Docker Hub with an empty on-node credential file, so that Day-2 credential
+		// delivery (e.g. rotating the on-node file) needs no node roll.
+		wireByDefault, wireErr := wireCredentialProviderByDefault(vars)
+		if wireErr != nil {
+			return wireErr
+		}
+		if !wireByDefault {
+			log.V(5).Info("Image registry credentials are not needed")
+			return nil
+		}
+		log.V(5).Info("Wiring kubelet dynamic credential provider for Docker Hub with empty credentials")
+		registriesThatNeedConfiguration = append(
+			registriesThatNeedConfiguration,
+			providerConfig{URL: v1alpha1.DefaultKubeletCredentialProviderRegistryURL},
+		)
 	}
 
 	files, commands, generateErr := generateFilesAndCommands(
@@ -244,6 +256,24 @@ func (h *imageRegistriesPatchHandler) Mutate(
 	}
 
 	return nil
+}
+
+// wireCredentialProviderByDefault reports whether the kubelet dynamic credential
+// provider should be wired even when no image registry or mirror is configured.
+// It defaults to true when the variable is unset so the behavior is on-by-default.
+func wireCredentialProviderByDefault(vars map[string]apiextensionsv1.JSON) (bool, error) {
+	wire, err := variables.Get[bool](
+		vars,
+		v1alpha1.ClusterConfigVariableName,
+		v1alpha1.WireImageCredentialProviderByDefaultVariableName,
+	)
+	if err != nil {
+		if variables.IsNotFoundError(err) {
+			return true, nil
+		}
+		return false, err
+	}
+	return wire, nil
 }
 
 func ensureOwnerReferenceOnCredentialsSecrets(
--- a/pkg/handlers/generic/mutation/generic/imageregistries/credentials/inject.go
+++ b/pkg/handlers/generic/mutation/generic/imageregistries/credentials/inject.go
@@ -136,18 +136,10 @@ func (h *imageRegistriesPatchHandler) Mutate(
 		return err
 	}
 	if len(registriesThatNeedConfiguration) == 0 {
-		// Nothing the user configured needs a credential provider. Unless the
-		// cluster opts out, still wire the kubelet dynamic credential provider for
-		// Docker Hub with an empty on-node credential file, so that Day-2 credential
-		// delivery (e.g. rotating the on-node file) needs no node roll.
-		wireByDefault, wireErr := wireCredentialProviderByDefault(vars)
-		if wireErr != nil {
-			return wireErr
-		}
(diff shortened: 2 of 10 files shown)


pkg/handlers/generic/mutation/generic/imageregistries/credentials/inject.go
```diff
@@ -139,8 +136,15 @@ func (h *imageRegistriesPatchHandler) Mutate(
 		return err
 	}
 	if len(registriesThatNeedConfiguration) == 0 {
-		log.V(5).Info("Image registry credentials are not needed")
-		return nil
+		// No user-configured registry needs a credential provider. Always wire the
```
dkoshkin: So the rotation process doesn't update the credential file? It only touches the creds file?
Can you remind me what the file format of the static creds is (name, location, etc). Also how does it work when there are creds for `docker.io` and `docker.io/nutanix`?

dkoshkin: This change in default behavior means that after CAREN is upgraded on next reconciliation every cluster would get rolled out (which we don't want until an explicit upgrade).
We prevent this by updating the handler version and moving previous behavior to an older handler.
The older versions are in 
https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/tree/main/pkg/handlers/v5

Here 's a similar PR, and the [relevant](https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1514/changes#diff-31ffb3d17b46cb1a8cf974e71b51fee059838561ff07e1a1630645ca3d02cf78) change where we move previous implementation to the older handler.

nilay-ntnx: Rotation only rewrites the on‑node static credentials file. It never touches the kubelet/credential‑provider config. Our `dynamic-credential-provider` does not support modifying this config. 

So we have these files - 
1. `/etc/kubernetes/image-credential-provider-config.yaml` -> kubelet CredentialProviderConfig (`matchImages` fields, and which provider handles which image). Content of this file is something like this - [Link](https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/main/pkg/handlers/generic/mutation/generic/imageregistries/credentials/credential_provider_config_files_test.go#L62).
2. `/etc/kubernetes/static-image-credentials.json` - This is the static creds file that our DaemonSet writes/updates on Day 2 for rotating the credentials. So we only modify this file for rotation. 

Updating the first file also requires restarting kubelet. So it is important it carries the right `matchImages` entry so later the Day 2 rotation works by just updating the static creds file (File 2). 

docker.io vs docker.io/nutanix: both resolve to registry-1.docker.io, which is what we always wire (`DefaultKubeletCredentialProviderRegistryURL`). So `matchImages` sends the call to dynamic-credential-provider where we can have individual auths for `docker.io` and `docker.io/nutanix`.

nilay-ntnx: I have raised this separate PR for the same - 
https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1684

dkoshkin: @nilay-ntnx I don't think the `/etc/kubernetes/static-image-credentials.json` will work the way you expect it to when a user also has `docker.io` configured (which they need to avoid rate limiting)

This code only uses `registryURL.Host` part of the registry 
https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/9bc64546ac268cf71896d7d552cceab5bd8af062/pkg/handlers/generic/mutation/generic/imageregistries/credentials/credentials_secret.go#L109-L122

So then this file will be written like

```
"auth": {
  "registry-1.docker.io": {"username": "general-user", "password": "..."},
  "docker.io":            {"username": "general-user", "password": "..."},
  "registry-1.docker.io": {"username": "nutanix-user", "password": "..."},
  "docker.io":            {"username": "nutanix-user", "password": "..."}
}
```

Then when its gets serialized the duplicate fields are ignored and its whatever was defined last in `imageRegistries` that gets picked up by the kubelet-credential-provider.

Were you able to actually verify the expected behavior by creating a cluster with both `docker.io` and `docker.io/nutanix` and then trying to pull private images from both of these?

dkoshkin: 🤦 sorry forgot about this https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1678/changes

dkoshkin: https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1687

dkoshkin: I think I still dont understand. If a user did provide a registry, but nothing did set that doker.io/nutanix registry this won't run so the credentials file won't be wired correctly.
Are we assuming that the webhook will always add docker.io to the imageRegstires?
Wouldn't it be safter to just always add this entry regardless if its empty or not (and just check that its not already in the list)?

nilay-ntnx: Yes, I missed this scenario. Made the changes as you mentioned. Always add this entry unless it's already in the list.

jimmidyson: Can we get some tests added for this please?

nilay-ntnx: Added the unit test. Thanks!

pkg/handlers/generic/mutation/generic/imageregistries/credentials/inject_test.go
```diff
@@ -231,6 +231,39 @@ var _ = Describe("Generate Image registry patches", func() {
...
+					{
+						Operation: "add",
+						Path:      "/spec/template/spec/kubeadmConfigSpec/files",
+						ValueMatcher: gomega.ContainElement(
+							gomega.SatisfyAll(
+								gomega.HaveKeyWithValue(
+									"path", "/etc/kubernetes/dynamic-credential-provider-config.yaml",
+								),
+								gomega.HaveKeyWithValue(
+									"content", gomega.ContainSubstring("registry-1.docker.io"),
```
jimmidyson: Shouldn't this just be `docker.io`?

api/v1alpha1/constants.go
```diff
@@ -40,6 +40,11 @@ const (
 	// ImageRegistriesVariableName is the image registries patch variable name.
 	ImageRegistriesVariableName = "imageRegistries"
 
+	// DefaultKubeletCredentialProviderRegistryURL is the registry the kubelet
+	// dynamic credential provider is always wired for (Docker Hub) when no
+	// registry/mirror is configured.
+	DefaultKubeletCredentialProviderRegistryURL = "https://registry-1.docker.io" //nolint:gosec // Not a credential.
```
jimmidyson, line 46: This should be `https://docker.io`.

pkg/handlers/generic/mutation/generic/imageregistries/credentials/inject.go
```diff
@@ -138,9 +135,15 @@ func (h *imageRegistriesPatchHandler) Mutate(
...
 		return err
 	}
-	if len(registriesThatNeedConfiguration) == 0 {
-		log.V(5).Info("Image registry credentials are not needed")
-		return nil
+	// Always configure kubelet credential-provider for DockerHub so Day-2
+	// credential delivery works (without node rollout) even when no registry
+	// or another registry was declared. If imageRegistries already declares the
+	// default entry, its existing configuration is preserved.
+	if !hasDefaultDockerHubRegistry(imageRegistries) {
```
jimmidyson: Shouldn't this be for `docker.io/nutanix`?

api/v1alpha1/constants.go
nilay-ntnx: @jimmidyson It will not make a difference. The template will generate both entries `docker.io` and `registry-1.docker.io` in `matchImages` if we pass registry-1.docker.io.

https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/main/pkg/handlers/generic/mutation/generic/imageregistries/credentials/templates/kubelet-image-credential-provider-config.yaml.gotmpl#L15

https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/main/pkg/handlers/generic/mutation/generic/imageregistries/credentials/templates/dynamic-credential-provider-config.yaml.gotmpl#L24

nilay-ntnx: We have tests which also use this value and validate this behaviour - 
https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/main/pkg/handlers/generic/mutation/generic/imageregistries/credentials/credential_provider_config_files_test.go#L86

pkg/handlers/generic/mutation/generic/imageregistries/credentials/inject_test.go
nilay-ntnx: @jimmidyson It will not make a difference. The template will generate both entries `docker.io` and `registry-1.docker.io` in `matchImages` if we pass registry-1.docker.io.

https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/main/pkg/handlers/generic/mutation/generic/imageregistries/credentials/templates/kubelet-image-credential-provider-config.yaml.gotmpl#L15

https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/main/pkg/handlers/generic/mutation/generic/imageregistries/credentials/templates/dynamic-credential-provider-config.yaml.gotmpl#L24

Would you recommend we make it `docker.io`?

nilay-ntnx: So as per this template, passing `registry-1.docker.io` will append both `docker.io` and `registry-1.docker.io`, "but" passing only `docker.io` appends just `docker.io`.
So in theory passing registry-1 URL gives us a better `matchImages` coverage I think.

jimmidyson: `registry-1.docker.io` handling is historic and is not actually required, especially at the credential provider level where images from docker hub will always be `docker.io` domain.

However it doesn't cause any harm to continue using this mechanism here for now so I'm ok with it and we can remove the deprecated handling of `registry-1.docker.io` in a future version.

jimmidyson: Let's add a test that `docker.io` is also present here as this is actually the important configuration.

nilay-ntnx: Added this test, Thanks!

pkg/handlers/generic/mutation/generic/imageregistries/credentials/inject_test.go
```diff
@@ -192,8 +211,94 @@ var _ = Describe("Generate Image registry patches", func() {
...
+						ValueMatcher: gomega.ContainElements(
+							gomega.HaveKeyWithValue(
+								"path", "/etc/caren/install-kubelet-credential-providers.sh",
+							),
+							gomega.HaveKeyWithValue(
+								"path", "/etc/kubernetes/image-credential-provider-config.yaml",
+							),
+							gomega.HaveKeyWithValue(
+								"path", "/etc/kubernetes/dynamic-credential-provider-config.yaml",
+							),
```
jimmidyson, line 286: Test for content here too would be good.

pkg/handlers/generic/mutation/generic/imageregistries/credentials/inject_test.go
```diff
@@ -192,8 +211,94 @@ var _ = Describe("Generate Image registry patches", func() {
...
+								gomega.HaveKeyWithValue(
+									"path", "/etc/kubernetes/dynamic-credential-provider-config.yaml",
+								),
+								// The template adds docker.io as an explicit alias
+								// whenever registry-1.docker.io is configured.
+								gomega.HaveKeyWithValue(
+									"content", gomega.ContainSubstring("registry-1.docker.io"),
+								),
+								gomega.HaveKeyWithValue(
+									"content", gomega.ContainSubstring(`- "docker.io"`),
```
jimmidyson: Why different format for the content here? They should be the same right? Can we actually specify the full contents of the file as expected rather than just these loose checks?

nilay-ntnx: If you're referring to the extra hyphen here - That extra hyphen proves that `docker.io` appears as a separate `matchImages` entry. Without the hyphen, checking only "docker.io" would also match the substring inside "registry-1.docker.io" and would not prove that it was added independently.

Let me specify the complete contents of the file expected

jimmidyson (approved):

pkg/handlers/generic/mutation/generic/imageregistries/credentials/inject.go
dkoshkin: I think there was an offline discussion about this, can you please add a reply here (or an even better an explanation comment). I don't understand why its adding `docker.io` and not `docker.io/nutanix`

dkoshkin (approved):

pkg/handlers/generic/mutation/generic/imageregistries/credentials/inject.go
nilay-ntnx: We basically need `matchImages` in our kubelet-credential-provider config to forward the requests to dynamic-credential-provider for authenticating these pulls.

[docker.io](http://docker.io/) also should be fine since if our static-credential-provider.json file only has auth for "[docker.io/nutanix](http://docker.io/nutanix)", and not for "[docker.io](http://docker.io/)", it's a no-op and it will try to pull unauthenticated (same behaviour)

https://nutanix.slack.com/archives/C0AC737DQCR/p1786007581589069?thread_ts=1785970172.728599&cid=C0AC737DQCR

PR #1680, opened 2026-07-24
refactor: [NCN-116148] Share HTTPS server for hooks and admission

jimmidyson:
## Summary
- Serve runtime hooks and admission on one webhook listener/TLS cert (port 9443, `runtimehooks-tls`)
- Keep separate Services; drop `admission-tls` / 9444 / `--admission-webhook-cert-dir`

## Test plan
- [ ] `go test ./common/pkg/server/`
- [ ] Helm render: one Certificate, both Services → `runtimehooks`, no 9444
- [ ] Runtime Discovery + admission path against shared cert

### DIFF
--- a/.specify/features/NCN-116148/plan.md
+++ b/.specify/features/NCN-116148/plan.md
@@ -1,0 +1,651 @@
+<!--
+ Copyright 2026 Nutanix. All rights reserved.
+ SPDX-License-Identifier: Apache-2.0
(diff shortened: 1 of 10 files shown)


dkoshkin (approved):
My only concern was about upgrading, but the spec has accounted for that.

supershal (approved):

PR #1684, opened 2026-07-28
feat: dockerhub nutanix - update handler version, move prev to v6

nilay-ntnx:
Upgrading CAREN must be a no-op for existing Clusters. Any change that alters the
patch output of a registered topology mutation handler will cause Machine rollouts
across all managed clusters.

Since as part of Nutanix Dockerhub Integration feature, we are updating files under `pkg/handlers/*/mutation/`, we need to - 

- Copy the current handler implementation to `pkg/handlers/v{current}/` (v6)
- Bump the version in the handler name
- Register old handler versions in `pkg/handlers/{provider}/handlers.go`
- Update default ClusterClass definitions in
   `charts/cluster-api-runtime-extensions-nutanix/defaultclusterclasses/` to reference
   the new version.

Fixes #


dkoshkin:
We don't want to copy every handler to the old version, only where there's a defaulting change, `pkg/handlers/generic/mutation/generic/imageregistries/credentials/inject.go` in this case.

With the much smaller scope, lets stack this PR on https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1679 and merge the approved changes into that PR, that way we don't merge something that may break clusters into main

dkoshkin:
Done in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/1687

PR #1688, opened 2026-07-31
feat: add guardrail to avoid vpc for metro

abhay-nutanix:
- Add a preflight guardrail that blocks k8s-ha metro cluster deployments when any referenced metro subnet is VPC-backed.
- Return a clear validation error that explains VPC-backed subnets are intentionally unsupported in v1 and asks users to use non-VPC subnets.
- Extend metro preflight unit tests to cover VPC-backed subnet rejection while preserving existing metro validation behavior.

## Why

The first version of k8s-ha metro does not support VPC. Failing fast during preflight prevents unsupported
cluster topologies from being deployed and avoids later runtime failures.

## Risk and Rollout

- Scope is limited to Nutanix metro preflight validation.
- No behavior change for non-metro clusters.
- Metro clusters using non-VPC subnets continue to be allowed.

### DIFF
--- a/pkg/webhook/preflight/nutanix/metro.go
+++ b/pkg/webhook/preflight/nutanix/metro.go
@@ -827,6 +827,8 @@ func (mc *metroCheck) resolvePrismElement(
 type metroSubnetAttributes struct {
 	// layers holds the distinct network layer names (VLAN, OVERLAY, ...).
 	layers map[string]struct{}
+	// vpcReferences holds distinct VPC references observed on subnets.
+	vpcReferences map[string]struct{}
 	// profilesByFailureDomain holds subnet profile sets keyed by failure domain.
 	profilesByFailureDomain map[string]map[string]struct{}
 }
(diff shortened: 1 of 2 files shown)


vijayaraghavanr31 (approved):

thunderboltsid:
> Add a preflight guardrail that blocks k8s-ha metro cluster deployments when any referenced metro subnet is VPC-backed.

Why is this?

pkg/webhook/preflight/nutanix/metro.go
```diff
@@ -888,6 +894,15 @@ func (mc *metroCheck) checkSubnetConsistency(attrs *metroSubnetAttributes, resul
 		return
 	}
 
+	if len(attrs.vpcReferences) > 0 {
+		failCheck(result, mc.field, fmt.Sprintf(
+			"VPC-backed subnets are not supported for k8s-ha metro. NutanixMetro %q uses VPC reference(s): %s. Use non-VPC subnets and retry.", //nolint:lll // Message is long.
```
thunderboltsid, line 899: k8s-ha metro is not a standardized terminology. Please user terminology that matches w/ UI and CLI

thunderboltsid (approved):

PR #1698, opened 2026-08-17
feat: create 2 storageClass for metro cluster

abhay-nutanix:
The Nutanix CSI lifecycle handler installs the Nutanix CSI driver and reconciles the provider storage classes on the
target cluster.

The metro storage-class adjustment is applied when the target cluster is a metro cluster (it uses `NutanixMetro/` or
`NutanixMetroSite/` failure domains). This applies to both management and workload clusters.

In that case, the default Nutanix storage class name stays unchanged (for example `nutanix-volume`) and remains the
Kubernetes default. The handler sets
`computeAffinity: DISABLED` on that default class.

The handler also creates one additional non-default storage class with the `-site-affinity` suffix (for example
`nutanix-volume-site-locality`) that omits `computeAffinity` so the CSI default (`ENABLED`) applies for workloads that
need site-local placement.

For metro clusters, the current flow in pkg/handlers/lifecycle/csi/nutanix/handler.go is:

- Start with provider.StorageClassConfigs from the addon config.
- Check isMetroCluster(cluster) and verify default storage provider is Nutanix.
- If true, transform configs via metroStorageClassConfigs(...):
           - Deep-copy all incoming configs (no input mutation).
           - For the configured default class key (for example volume), set computeAffinity=DISABLED.
           - Create one extra non-default class by appending the suffix -site-locality to the default key (for example           volume-site-locality).
           - On that extra class, remove computeAffinity so CSI default behavior applies.
- Pass the resulting map to CreateStorageClassesOnRemote(...) in pkg/handlers/lifecycle/csi/utils/scs.go.

Then in CreateStorageClassesOnRemote(...):

- For each config entry, build a Kubernetes StorageClass.
- Name format is <provider>-<configName> (for Nutanix: nutanix-<name>).
- Merge default parameters + user parameters.
- Mark only one class as Kubernetes default:
          - when provider matches defaultStorage.Provider, and
          - config key equals defaultStorage.StorageClassConfig.
- Apply each class to the target cluster using server-side apply.

So for a metro cluster with default key volume, you effectively get:

nutanix-volume (default class, forced computeAffinity: DISABLED)
nutanix-volume-site-locality (non-default class, computeAffinity omitted)

### DIFF
--- a/pkg/handlers/lifecycle/csi/nutanix/handler.go
+++ b/pkg/handlers/lifecycle/csi/nutanix/handler.go
@@ -35,6 +35,9 @@ const (
 const (
 	defaultHelmReleaseName      = "nutanix-csi"
 	defaultHelmReleaseNamespace = "ntnx-system"
+	computeAffinityParameterKey = "computeAffinity"
+	computeAffinityDisabled     = "DISABLED"
+	metroStorageClassSuffix     = "-site-affinity"
 
 	//nolint:gosec // Does not contain hard coded credentials.
 	defaultCredentialsSecretName = "nutanix-csi-credentials"
@@ -184,10 +187,18 @@ func (n *NutanixCSI) Apply(
 		return fmt.Errorf("failed to apply nutanix CSI addon: %w", err)
 	}
 
+	storageClassConfigs := provider.StorageClassConfigs
+	if isMetroCluster(cluster) && defaultStorage.Provider == v1alpha1.CSIProviderNutanix {
+		storageClassConfigs = metroStorageClassConfigs(
+			provider.StorageClassConfigs,
+			defaultStorage.StorageClassConfig,
+		)
+	}
+
 	err = csiutils.CreateStorageClassesOnRemote(
 		ctx,
 		n.client,
-		provider.StorageClassConfigs,
+		storageClassConfigs,
 		cluster,
 		defaultStorage,
 		v1alpha1.CSIProviderNutanix,
@@ -198,6 +209,38 @@ func (n *NutanixCSI) Apply(
 		return fmt.Errorf("error creating StorageClasses for the Nutanix CSI driver: %w", err)
 	}
 	return nil
+}
+
+func metroStorageClassName(baseName string) string {
+	return baseName + metroStorageClassSuffix
+}
+
+func metroStorageClassConfigs(
+	configs map[string]v1alpha1.StorageClassConfig,
+	defaultStorageClassName string,
+) map[string]v1alpha1.StorageClassConfig {
+	metroConfigs := make(map[string]v1alpha1.StorageClassConfig, len(configs)+1)
+	for name, config := range configs {
+		metroConfigs[name] = *config.DeepCopy()
+	}
+
+	defaultConfig, found := metroConfigs[defaultStorageClassName]
+	if !found {
+		return metroConfigs
+	}
+
+	availabilityOptimized := *defaultConfig.DeepCopy()
+	if availabilityOptimized.Parameters == nil {
+		availabilityOptimized.Parameters = map[string]string{}
+	}
+	availabilityOptimized.Parameters[computeAffinityParameterKey] = computeAffinityDisabled
+	metroConfigs[defaultStorageClassName] = availabilityOptimized
+
+	siteAffinityConfig := *defaultConfig.DeepCopy()
+	delete(siteAffinityConfig.Parameters, computeAffinityParameterKey)
+	metroConfigs[metroStorageClassName(defaultStorageClassName)] = siteAffinityConfig
+
+	return metroConfigs
 }
 
 func templateValuesFunc(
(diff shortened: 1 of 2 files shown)


pkg/handlers/lifecycle/csi/nutanix/handler.go
```diff
@@ -35,6 +37,9 @@ const (
 const (
 	defaultHelmReleaseName      = "nutanix-csi"
 	defaultHelmReleaseNamespace = "ntnx-system"
+	computeAffinityParameterKey = "computeAffinity"
+	computeAffinityDisabled     = "DISABLED"
+	metroStorageClassSuffix     = "-metro"
```
vijayaraghavanr31: Let's not rename the default StorageClass - keep it as nutanix-volume (still the default), and for metro clusters just set computeAffinity: DISABLED on it. Then add an additional non-default class that keeps compute affinity enabled, named something like **<name>-data-locality or <name>-site-affinity**. This avoids breaking every consumer that references the default class by name.

pkg/handlers/lifecycle/csi/nutanix/handler.go
```diff
@@ -200,6 +214,57 @@ func (n *NutanixCSI) Apply(
...
+	}
+
+	managementCluster, err := capiutils.ManagementCluster(ctx, c)
+	if err != nil || managementCluster == nil {
+		return false
+	}
+
+	return managementCluster.Name == cluster.Name &&
+		managementCluster.Namespace == cluster.Namespace
+}
```
vijayaraghavanr31: we have to fix this. this is applicable for both management cluster and workload clusters

pkg/handlers/lifecycle/csi/nutanix/handler.go
```diff
@@ -200,6 +214,57 @@ func (n *NutanixCSI) Apply(
...
+		metroConfigs[name] = config
+	}
+
+	defaultConfig, found := metroConfigs[defaultStorageClassName]
+	if found {
+		siteAffinityConfig := defaultConfig
+		siteAffinityConfig.Parameters = maps.Clone(defaultConfig.Parameters)
+		if siteAffinityConfig.Parameters == nil {
+			siteAffinityConfig.Parameters = map[string]string{}
+		}
```
vijayaraghavanr31: Before inserting <default>-site-affinity, delete computeAffinity from siteAffinityConfig.Parameters; otherwise this class also gets DISABLED due to cloning from defaultConfig.

vijayaraghavanr31 (changes requested):

pkg/handlers/lifecycle/csi/nutanix/handler.go
```diff
@@ -35,6 +37,9 @@ const (
 const (
 	defaultHelmReleaseName      = "nutanix-csi"
 	defaultHelmReleaseNamespace = "ntnx-system"
+	computeAffinityParameterKey = "computeAffinity"
+	computeAffinityDisabled     = "DISABLED"
```
dkoshkin, line 39: Instead of explicitly doing this, have you considered extending the API https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/5b901075dd8d323ad50f1e2ff3711144bf59b415/examples/capi-quick-start/nutanix-cluster-cilium-helm-addon.yaml#L96-L99 and having clients define this SC?

It might be unexpected that a single set of configs from `storageClassConfigs` is now being used across 2 different SCs.

pkg/handlers/lifecycle/csi/nutanix/handler.go
```diff
@@ -200,6 +214,57 @@ func (n *NutanixCSI) Apply(
...
+	for name, config := range configs {
+		parameters := maps.Clone(config.Parameters)
+		if parameters == nil {
+			parameters = map[string]string{}
+		}
+		config.Parameters = parameters
+		if name == defaultStorageClassName {
+			config.Parameters[computeAffinityParameterKey] = computeAffinityDisabled
+		}
+		metroConfigs[name] = config
```
dkoshkin: Is `computeAffinity: ENABLED` the default now, why does `DISABLED` need to be set here explicitly? Also SC params are immutable, if we're changing the params it will break on upgrades.

pkg/handlers/lifecycle/csi/nutanix/handler.go
```diff
@@ -200,6 +214,57 @@ func (n *NutanixCSI) Apply(
...
 }
 
+func metroStorageClassName(baseName string) string {
+	return baseName + metroStorageClassSuffix
+}
+
+func metroStorageClassConfigs(
+	configs map[string]v1alpha1.StorageClassConfig,
+	defaultStorageClassName string,
+) map[string]v1alpha1.StorageClassConfig {
```
dkoshkin, line 221: Am I reading this function correctly, it takes in a map of `StorageClassConfig`, does some mutation to the default SC configs and then copies its value to a new `StorageClassConfig`?
Lets avoid mutating a SC in this function please, its very opaque.

dkoshkin (changes requested):
Is there a design doc for the changes here? Im not sure about the approach proposed here.

pkg/handlers/lifecycle/csi/nutanix/handler.go
abhay-nutanix: For metro we are not supporting Brownfield scenarios.

PR #1701, opened 2026-08-20
fix: ignore service load balancer ranges in Nutanix CCM

abhay-nutanix:
## Summary

- Update Nutanix CCM values templating to append ServiceLoadBalancer address ranges to `ignoredNodeIPs`.
- Keep scope limited to LB range handling only (no new API fields).
- Add/adjust unit tests in the Nutanix CCM handler to validate LB range rendering.

## Why

CCM node address discovery should not consider load balancer service IP pools/ranges as valid node IPs.
This change ensures configured LB ranges are explicitly ignored.

## What changed

- `pkg/handlers/lifecycle/ccm/nutanix/handler.go`
  - `ipsToIgnore(...)` now includes ServiceLoadBalancer `addressRanges` as `start-end` entries.
- `pkg/handlers/lifecycle/ccm/nutanix/handler_test.go`
  - adds/keeps test coverage verifying LB ranges are present in `ignoredNodeIPs`.
- Removed unrelated `IgnoredNodeIPs` API-field work to keep this patch strict.

## Expected behavior

When ServiceLoadBalancer ranges are configured, Nutanix CCM receives those ranges in `ignoredNodeIPs`,
preventing them from being used during node IP discovery.

### DIFF
--- a/pkg/handlers/lifecycle/ccm/nutanix/handler.go
+++ b/pkg/handlers/lifecycle/ccm/nutanix/handler.go
@@ -132,7 +132,7 @@ func (p *provider) Apply(
 		),
 		p.client,
 		helmChart,
-	).WithValueTemplater(templateValuesFunc(clusterConfig.Nutanix))
+	).WithValueTemplater(templateValuesFunc(clusterConfig))
 
 	if err = applier.Apply(ctx, cluster, p.config.DefaultsNamespace(), log); err != nil {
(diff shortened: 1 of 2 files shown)


vijayaraghavanr31 (approved):

dkoshkin (changes requested):
This type of handler change to the default behavior will cause a unwanted cluster rollout during CAREN upgrade, and we would need to include the previous behavior in https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/tree/main/pkg/handlers/v6

PR #1703, opened 2026-08-25
feat: add webhook for CSI flag for metro

abhay-nutanix:
Add a admission webhook to add flag computeAffinity:DISABLED for CSI for metro.

### DIFF
--- a/pkg/helpers/metro.go
+++ b/pkg/helpers/metro.go
@@ -1,0 +1,170 @@
+// Copyright 2026 Nutanix. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package helpers
+
+import (
+	"errors"
+	"fmt"
+	"slices"
+	"strings"
+
+	clusterv1 "sigs.k8s.io/cluster-api/api/core/v1beta2"
+
+	"github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/api/v1alpha1"
+	"github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/api/variables"
+)
+
+const (
+	// MetroFailureDomainPrefix is the CAPI failure-domain name prefix for a NutanixMetro.
+	MetroFailureDomainPrefix = "NutanixMetro/"
+	// MetroSiteFailureDomainPrefix is the CAPI failure-domain name prefix for a NutanixMetroSite.
+	MetroSiteFailureDomainPrefix = "NutanixMetroSite/"
+
+	// ComputeAffinityParameter is the Nutanix CSI StorageClass parameter that pins
+	// a volume to the Prism Element hosting the consuming VM.
+	ComputeAffinityParameter = "computeAffinity"
+	// ComputeAffinityDisabled disables PE-local volume affinity. Metro clusters
+	// require this value because they span two Prism Elements.
+	ComputeAffinityDisabled = "DISABLED"
+)
+
+// IsMetroFailureDomain reports whether fd names a NutanixMetro or NutanixMetroSite.
+func IsMetroFailureDomain(fd string) bool {
+	return strings.HasPrefix(fd, MetroFailureDomainPrefix) ||
+		strings.HasPrefix(fd, MetroSiteFailureDomainPrefix)
+}
+
+// IsMetroCluster returns true when the cluster uses metro-aware failure domains,
+// i.e. any control-plane or worker failure domain references a NutanixMetro or
+// NutanixMetroSite object (identified by the respective name prefix).
+func IsMetroCluster(cluster *clusterv1.Cluster) bool {
+	if cluster == nil || !cluster.Spec.Topology.IsDefined() {
+		return false
+	}
+
+	clusterConfig, err := variables.UnmarshalClusterConfigVariable(cluster.Spec.Topology.Variables)
+	if err == nil &&
+		clusterConfig != nil &&
+		clusterConfig.ControlPlane != nil &&
+		clusterConfig.ControlPlane.Nutanix != nil {
+		if slices.ContainsFunc(clusterConfig.ControlPlane.Nutanix.FailureDomains, IsMetroFailureDomain) {
+			return true
+		}
+	}
+
+	for i := range cluster.Spec.Topology.Workers.MachineDeployments {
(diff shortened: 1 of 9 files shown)


pkg/helpers/metro.go
```diff
@@ -0,0 +1,170 @@
+// Copyright 2026 Nutanix. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package helpers
```
dkoshkin: nit is there a more descriptive package name than "helpers" how about just "metro"?

pkg/helpers/metro.go
```diff
@@ -0,0 +1,170 @@
...
+	return false
+}
+
+// DefaultMetroCSIComputeAffinity sets computeAffinity=DISABLED on Nutanix CSI
+// StorageClassConfigs of a metro Cluster when the parameter is unset or empty.
+// It does not overwrite a non-empty user-provided value.
+func DefaultMetroCSIComputeAffinity(cluster *clusterv1.Cluster) (bool, error) {
+	if !IsMetroCluster(cluster) {
+		return false, nil
+	}
```
dkoshkin: Lets be more explicit and have this check in the webhook.
This function should assume that the caller has already checked that needs to call this function.

pkg/helpers/metro.go
```diff
@@ -0,0 +1,170 @@
...
+		if sc.Parameters[ComputeAffinityParameter] != "" {
+			continue
+		}
+		sc.Parameters[ComputeAffinityParameter] = ComputeAffinityDisabled
+		provider.StorageClassConfigs[name] = sc
+		mutated = true
+	}
+	if !mutated {
+		return false, nil
+	}
```
dkoshkin: Is this just an extra safety? Can't we always just marshal it which would be a no-op with no change?

pkg/webhook/cluster/metro_storageclass.go
```diff
@@ -0,0 +1,94 @@
...
+		return admission.Allowed("")
+	}
+
+	mutated, err := helpers.DefaultMetroCSIComputeAffinity(cluster)
+	if err != nil {
+		return admission.Errored(http.StatusBadRequest, err)
+	}
+	if !mutated {
+		return admission.Allowed("")
+	}
```
dkoshkin: I don't think we use this mutated pattern in any of the other webhooks, lets just always marshal it?

pkg/webhook/cluster/metro_storageclass.go
```diff
@@ -0,0 +1,94 @@
...
+
+	"github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pkg/helpers"
+)
+
+type metroCSIComputeAffinity struct {
+	client  ctrlclient.Client
+	decoder admission.Decoder
+}
+
+func NewMetroCSIComputeAffinity(
```
dkoshkin: How about we name this more generic like `NewMetroClustersDefaulter`, computeAffinity is very specific to the implementation of the metro helpers function, this webhook should just orchestrate all the changes.

pkg/metro/metro.go
```diff
@@ -0,0 +1,170 @@
...
+
+// IsMetroFailureDomain reports whether fd names a NutanixMetro or NutanixMetroSite.
+func IsMetroFailureDomain(fd string) bool {
+	return strings.HasPrefix(fd, MetroFailureDomainPrefix) ||
+		strings.HasPrefix(fd, MetroSiteFailureDomainPrefix)
+}
+
+// IsMetroCluster returns true when the cluster uses metro-aware failure domains,
+// i.e. any control-plane or worker failure domain references a NutanixMetro or
+// NutanixMetroSite object (identified by the respective name prefix).
```
dkoshkin, line 40: >or NutanixMetroSite object

Is this correct? We want default computeAffinity to Disabled even with `NutanixMetroSite`?
From the offline conversation I though that for a single site this doesn't apply and there we do want to keep `computeAffinity: Enabled` so that workloads don't get rescheduled on a different Metro site.

thunderboltsid: As long as they're on a NutanixMetro or NutanixMetroSite, they're covered by protection policies that ensure syncrep of data. ComputeAffinity was only needed for the "performance-optimized" case.

vijayaraghavanr31: MetroSite pins the node pools to a preferred PE, not the pods. With computeAffinity enabled, the PV gets pinned to the PE of the node where it was first bound. If the pod is later scheduled on a node from another site, this can cause the same issue.

Also, StorageClass is cluster-wide. So even if the initial node pool is single-site, a user can add another node pool later with a Metro failure domain, and we can run into the same issue.

So I think computeAffinity: Disabled is safer as the default for both NutanixMetro/ and NutanixMetroSite/.

cc: @abhay-nutanix

vijayaraghavanr31 (approved):
LGTM

pkg/webhook/cluster/metro_clusters.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	cluster := &clusterv1.Cluster{}
+	if err := a.decoder.Decode(req, cluster); err != nil {
+		return admission.Errored(http.StatusBadRequest, err)
+	}
+
+	if !cluster.Spec.Topology.IsDefined() || !metro.IsMetroCluster(cluster) {
+		return admission.Allowed("")
+	}
```
dkoshkin, line 56: nit: helps with readability. 
```suggestion
	if !cluster.Spec.Topology.IsDefined() {
		return admission.Allowed("")
	}
	
	if !metro.IsMetroCluster(cluster) {
		return admission.Allowed("")
	}
```

pkg/webhook/cluster/metro_clusters.go
```diff
@@ -0,0 +1,90 @@
...
+
+func (a *metroClusters) Defaulter() admission.HandlerFunc {
+	return a.defaulter
+}
+
+func (a *metroClusters) Validator() admission.HandlerFunc {
+	return a.validate
+}
+
+func (a *metroClusters) defaulter(
```
dkoshkin, line 41: I think the other webhooks have some unit tests, it would be great to add some here.

dkoshkin (approved):
Thanks, looks good.

pkg/webhook/cluster/metro_clusters.go
abhay-nutanix: Will add some

PR #1712, opened 2026-09-11
fix: metro preflights

abhay-nutanix:
- Reject metro Clusters whose NutanixMetroSite objects share a preferredFailureDomain or groupNameLabel, so both sites cannot pin to the same PE or topology segment (for example both sites preferring fd0).

- Reject names that would make CAPX’s default vHA category value longer than 64 characters: k8s-vha-capx-{cluster}-{metro}-default-{idx}. CAPX now fails fast on that limit ([cluster-api-provider-nutanix#782](https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pull/782)); this catches it at admission and asks to shorten the Cluster or metro name.

### DIFF
--- a/pkg/webhook/preflight/nutanix/metro.go
+++ b/pkg/webhook/preflight/nutanix/metro.go
@@ -394,23 +394,22 @@ func (c *metroVHACategoryNameCheck) Run(_ context.Context) preflight.CheckResult
 	result := preflight.CheckResult{Allowed: true}
 
 	// CAPX synthesizes a single "default" movement group and one category per
-	// metro failure domain (index 0 and 1).
+	// metro failure domain. Indexes 0 and 1 are single-digit, so length is the
+	// same; check idx 0 only to avoid duplicate causes.
 	for _, metroName := range c.metroNames {
 		domainName := vhaDomainName(c.clusterName, metroName)
-		for idx := range metroPrismElementCount {
-			value := vhaCategoryValue(domainName, vhaDefaultMovementGroup, idx)
-			if len(value) <= prismCategoryValueMaxLen {
-				continue
-			}
-			failCheck(&result, c.field, fmt.Sprintf(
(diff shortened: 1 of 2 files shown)


Vedansh074 (approved):

pkg/webhook/preflight/nutanix/metro.go
```diff
@@ -244,6 +268,154 @@ func (c *allNodePoolsMetroCheck) Run(_ context.Context) preflight.CheckResult {
...
+	// CAPX synthesizes a single "default" movement group and one category per
+	// metro failure domain (index 0 and 1).
+	for _, metroName := range c.metroNames {
+		domainName := vhaDomainName(c.clusterName, metroName)
+		for idx := range metroPrismElementCount {
+			value := vhaCategoryValue(domainName, vhaDefaultMovementGroup, idx)
+			if len(value) <= prismCategoryValueMaxLen {
+				continue
+			}
+			failCheck(&result, c.field, fmt.Sprintf(
```
vijayaraghavanr31: idx 0 and 1 produce the same string length (single-digit index), so a too-long name emits two nearly identical causes. Prefer checking once (e.g. only idx == 0) or break after the first failure.

pkg/webhook/preflight/nutanix/metro.go
```diff
@@ -244,6 +268,154 @@ func (c *allNodePoolsMetroCheck) Run(_ context.Context) preflight.CheckResult {
...
+}
+
+func vhaCategoryValue(vHADomainName, group string, idx int) string {
+	return fmt.Sprintf("k8s-vha-capx-%s-%s-%d", vHADomainName, group, idx)
+}
+
+func (c *metroVHACategoryNameCheck) Run(_ context.Context) preflight.CheckResult {
+	result := preflight.CheckResult{Allowed: true}
+
+	// CAPX synthesizes a single "default" movement group and one category per
```
vijayaraghavanr31, line 396: Nit / future: this only validates the synthesized default movement group, which is correct for the current CAPX metro path.
No change needed in this PR. When we support granular movement groups, extend this check to cover custom group names in the generated category values.

vijayaraghavanr31 (approved):

